PR #15, opened 2020-01-31
console entry point

luiztauffer:
creates a console shortcut to open an empty GUI.
```shell
nwbn-gui
```

it might be extended to include optional arguments, such as metadata file path


bendichter:
I like the idea of console scripts, but I don't see how this abstract function would ever be used. Wouldn't it be better to implement this in individual lab repos?

luiztauffer:
it’s a quick call to open a gui instance without the need to find any particular ‘nwb_gui.py’ script in the computer manually. I’ve been using it a lot, mostly to visualize nwb files with the explorer tab. Just a nice feature.
You’re right we could have a similar thing for each lab-specific repo! I can start it with Jaeger’s lab, how about `jaeger-nwb` for the shortcut name?

PR #16, opened 2020-02-01
Console entry point

luiztauffer:


bendichter:
How would we use that abstract function?

PR #17, opened 2020-02-04
Generalizible form class

luiztauffer:
Introduces `BasicFormFixed` and `BasicFormCollapsible` classes that aim to generalize most of the reading/writing/updating operations between matadata<-->forms.

- [x] fill forms from metadata
- [x] auto-update comboboxes
- [x] read metadata from forms
- [x] usable by extensions (e.g. [fret](https://github.com/ben-dichter-consulting/ndx-fret) already uses it)
- [ ] make them the base classes for all groups forms
- [ ] the linking or sub-grouping of other groups has to be defined for each GUI group's class. Perhaps that could be automated with the metafile json schema


luiztauffer:
@bendichter 
I think we should merge this, since it has the updated and functioning version for the main projects using the GUI so far. Furthermore, this has fixes necessary for Jaeger, so it would be good to have it as pip installable asap.

> - [ ] make them the base classes for all groups forms

This is partially done. Form classes used for axel, giocomo, buffalo and jaeger projects were updated. Since there are just too many groups (one per pynwb class), I propose we go updating according to the demand.

> - [ ] the linking or sub-grouping of other groups has to be defined for each GUI group's class. Perhaps that could be automated with the metafile json schema.

I guess this depends on how the metafiles will be organized finally. For now, all it takes is to declare inside the method `fields_info_update()` the specific types of links or subgroups (e.g. Device, OpticalChannel, TimeSeries, ...)

PR #20, opened 2020-03-20
Intan

luiztauffer:
Creates class `Intan2NWB` to convert electrophysiology data form `rhd` files to nwb.

This PR is based on the `improve_generalization_converter` branch and should only be merged after #19


nwbn_conversion_tools/ephys/intan/intan.py
```diff
@@ -0,0 +1,157 @@
...
+        ephys_ts = ElectricalSeries(
+            name=metadata_ecephys['ElectricalSeries'][0]['name'],
+            description=metadata_ecephys['ElectricalSeries'][0]['description'],
+            data=data_iter,
+            electrodes=electrode_table_region,
+            rate=sampling_rate,
+            starting_time=0.0,
+            conversion=es_conversion_factor
+        )
+        self.nwbfile.add_acquisition(ephys_ts)
```
bendichter: It seems like most of this is not specific to intan but general extracellular electrophysiology code. Remember, we want to minimize repeating code. If you think some of these lines might be used by another acquisition system, is there a way to make a general ecephys conversion class that holds some of this? It seems like this is more general and could be in `NWBEcephysConverter` and then have `NWBConverter`>`NWBEcephysConverter`>`Intan2NWB`.

Issue #24, opened 2020-04-06
testing


bendichter:
Add test for reading from recording extractor and writing

Use https://github.com/SpikeInterface/spikeextractors/tree/master/spikeextractors/example_datasets for generating fake data

Use actions to set up continuous integration [similar to NWBWidgets](https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/tree/master/.github/workflows)

bendichter:
@Armin12 can you help with this?

Armin12:
Sure!

yarikoptic:
and let me and/or @mvdm know if you would like some real data for neuralynx systems, we could provide! ;)

bendichter:
What's the best way to integrate real data with CI?

yarikoptic:
I guess it all depends on the sizes etc.  We could setup some datalad datasets (could even store data within github lfs while sizes allow for free storage, see [datalad handbook for an example](http://handbook.datalad.org/en/latest/basics/101-138-sharethirdparty.html#use-github-for-sharing-content)), and use datalad to get desired files from the cloudy cloud.

yarikoptic:
here is e.g. one example in heudiconv where we do some regression testing using data we fetch using datalad (from my server I believe): https://github.com/nipy/heudiconv/blob/ca402a9fa701551bf2c74f72d77223aaad7e5d68/heudiconv/tests/test_regression.py

PR #26, opened 2020-04-07
Testing ci

luiztauffer:
a kickstart for #24


luiztauffer:
@bendichter 
the tests are included in the package, this would allow for other projects to import any tests methods implemented here. 
Can you take a look and say if that's what you had in mind, before I make a new release on pypi?

bendichter:
Actually I mean testing _infrastructure_ should be in the package. For example, in SpikeExtractors, the code to generate a fake RecordingExtractor [here](https://github.com/SpikeInterface/spikeextractors/tree/master/spikeextractors/example_datasets) should be importable. The tests themselves should not be.

PR #30, opened 2020-05-11
Seg ext integration

bendichter:


bendichter:
@Saksham20 Can you make a general SegmentationExtractor class that Sima2NWB inherits from? Similar to https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/ecephys/ecephys.py#L9

bendichter:
Thanks for the refactor, @luiztauffer, this looks a lot cleaner. @Saksham20, please don't forget to refactor this as described above.

Saksham20:
@luiztauffer  thanks for making the change to forms to fill in default values. However, there are still some that are `"None"` strings instead of `NoneType`. Forexample, the starting_time, rate within DfOverF
I cannot get yaml to force load them as NoneType. Is there a work around?  

Maybe a better strategy could be to exclude those which are unfilled, while only enforcing those that are mandatory?

the updated yaml file that is output of gui: 
```
NWBFile:
  data_collection: null
  experiment_description: brief experiment description
  experimenter: testname
  identifier: nwbfile id
  institution: inistitution name
  keywords:
  - keyword1
  - keyword2
  lab: lab name
  notes: nwbfile notes
  pharmacology: pharma
  protocol: protocol used
  related_publications: pub1, pub2
  session_description: ses desc
  session_id: session id goes here
  session_start_time: 2020-02-02 00:00:00
  slices: null
  stimulus_notes: null
  surgery: surgery description
  virus: virus info
Ophys:
  Device:
  - name: ophys_Device1
  - name: ophys_Device2
  DfOverF:
    name: df/f_name
    roi_response_series:
    - comments: no comments
      description: description of roi resp series
      name: testroi_resp_series
      rate: None
      starting_time: None
      unit: ''
  Fluorescence:
    name: Fluorescence
    roi_response_series: []
  ImageSegmentation:
    name: ABCD
    plane_segmentations:
    - description: Segmented ROIs
      imaging_planes: ImagingPlane_1
      name: MyplaneSegmentation
  ImagingPlane:
  - conversion: 1.0
    description: ADDME
    device: ophys_Device1
    excitation_lambda: 488.0
    grid_spacing_unit: meters
    imaging_rate: 1.0
    indicator: NLS-GCaMP6s
    location: whole central brain
    name: ImagingPlane_1
    optical_channel: []
    origin_coords_unit: meters
    reference_frame: ''
    unit: meters
Subject:
  age: ''
  date_of_birth: 2010-02-02 00:00:00
  description: subject descriptoin
  genotype: genotype
  sex: F
  species: species
  subject_id: subjectID
  weight: weight number
```

bendichter:
I think leaving it blank should read as a proper `None` object (of `NoneType`). e.g.

```yaml
data_none_type:
data_str: None
```

Does that work?

Saksham20:
It could work that way, but that is the loaded yaml file sent as argument by the `nwb_conversion_gui()` into the conversion_module. 
One option is that I recursively fill in the None strings with NoneTypes wherever found OR 
The `nwb_conversion_gui()` function itself omits those fields and let pynwb APIs handle defaults.
I think the second would be a better option

bendichter:
@Saksham20 I don't understand

Saksham20:
@luiztauffer  Were you able to figure out a way to correctly output unfilled fields in the form? 

> @Saksham20 @luiztauffer The GUI should output a proper yaml file. If the field is meant to be None, then the yaml should be. I'd like to avoid creating work-arounds where we parse "None" or "null" strings if we can.
> 
> ```yaml
> data_none_type:
> ```

nwb_conversion_tools/gui/classes/forms_basic.py
```diff
@@ -50,9 +50,11 @@ def fill_fields_info(self):
             # Required fields get a red star in their label
             if 'default' not in field:
                 required = True
+                default = 'required: '+ str(field['type'])
```
bendichter, line 53: is this auto-filled?

nwb_conversion_tools/gui/classes/forms_basic.py
```diff
@@ -228,9 +247,11 @@ def fill_fields_info(self):
...
```
bendichter, line 250: ?

nwb_conversion_tools/gui/classes/forms_basic.py
```diff
@@ -279,9 +302,20 @@ def make_forms(self):
             # String types
             if field['type'] == 'str':
                 form = QLineEdit('')
+                if field['required']:
+                    form.setPlaceholderText(field['default'])
```
bendichter, line 306: I don't understand these lines. First of all, they seem to be copy/pasted all over the place, which is bad. Second of all, am I reading this right that we are not inputing the data for required fields? Why not?

nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -649,15 +657,24 @@ def __init__(self, parent):
...
+            spec = importlib.util.spec_from_file_location(os.path.basename(mod_file).strip('.py'), mod_file)
+            conv_module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(conv_module)
+            metadata = self.parent.read_metadata_from_form()
+            conv_module.conversion_function(source_paths=self.parent.source_paths,
+                                            f_nwb=self.parent.lin_nwb_file.text(),
+                                            metadata=metadata,
+                                            **self.parent.kwargs_fields)
+        else:
+            metadata = self.parent.read_metadata_from_form()
```
bendichter, line 673: If the user inputs both a module and a class, that should throw an error.

nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -28,9 +28,9 @@
 
 
 class Application(QMainWindow):
-    def __init__(self, metafile=None, conversion_module='', source_paths={},
-                 kwargs_fields={}, extension_modules={}, extension_forms={},
-                 show_add_del=False):
+    def __init__(self, metafile=None, conversion_module=None, source_paths=None,
+                 kwargs_fields=None, extension_modules=None, extension_forms=None,
+                 show_add_del=False, nwbfile_loc=None, conversion_class=None):
         super().__init__()
```
bendichter, line 34: Can you add a Numpy-style docstring here? If a `conversion_module` and a `conversion_class` are both entered, you should throw an error.

nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -679,11 +696,16 @@ def wheelEvent(self, event):
...
 # If it is imported as a module
-def nwb_conversion_gui(metafile=None, conversion_module='', source_paths={},
-                       kwargs_fields={}, extension_modules={}, extension_forms={},
-                       show_add_del=False):
+def nwb_conversion_gui(metafile=None, conversion_module=None, source_paths=None,
+                       kwargs_fields=None, extension_modules=None, extension_forms=None,
+                       show_add_del=False, nwbfile_loc=None, conversion_class=None):
     """Sets up QT application."""
+    if conversion_module:
+        raise DeprecationWarning('use of conversion_module will be replaced by conversion_class'
```
bendichter, line 704: 👍

nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -679,11 +696,16 @@ def wheelEvent(self, event):
...
+                                 'with its \'run_conversion\' method called directly')
     app = QtCore.QCoreApplication.instance()
+    if conversion_class is None and conversion_module is None:
+        raise Exception('provide one of conversion_module:str or conversion_class:class')
     if app is None:
```
bendichter, line 709: If both are conversion_module and conversion_class are provided, it should result in an error

nwb_conversion_tools/gui/classes/forms_basic.py
Saksham20: it is copy pasted twice, one for the  class `BasicFormCollaspsible`, and other for `BasicFormFixed`. It does not fill in any value for the required fields, as we had agreed on, it is for the user to input. When using `QLineEdit.text()` (within `read_fields()` method) to read the value of field to copy to dictionary it reads an empty string (if the user did not fill it)

nwb_conversion_tools/gui/classes/forms_basic.py
Saksham20: Fills in the form like so:
image

nwb_conversion_tools/gui/classes/forms_basic.py
Saksham20: only for the placeholder to display the grey text not the actual field, which remains empty for the user to fill

nwb_conversion_tools/gui/classes/forms_basic.py
bendichter: if there is a default value specified in the yaml file, that should be used.

nwb_conversion_tools/gui/classes/forms_basic.py
bendichter: ok that's fine

nwb_conversion_tools/gui/classes/forms_general.py
```diff
@@ -276,8 +276,11 @@ def write_fields(self, data={}):
             str_datetime = data['session_start_time'].strftime('%d/%m/%Y, %H:%M')
             self.form_session_start_time1.setText(str_datetime.split(',')[0])
             self.form_session_start_time2.setText(str_datetime.split(',')[1].strip())
-        if 'experimenter' in data:
-            self.form_experimenter.setText(data['experimenter'])
+        if 'experimenter' in data and data['experimenter'] is not None:
+            if isinstance(data['experimenter'],list):
+                self.form_experimenter.setText(','.join(str(x) for x in data['experimenter']))
+            else:
+                self.form_experimenter.setText(data['experimenter'])
```
Saksham20, line 283: we need to decide whether we want experimenter data to be an array, string or both. Currently it accepts both.
In the [schema ](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/metafile.schema.json#L23)its an array but the code expected it as a string initially.

Saksham20:
Issue with nwb_widgets #39

Issue #31, opened 2020-06-02
auto-populate source paths


bendichter:
Source paths should auto-populate in the GUI if specified during initialization

luiztauffer:
@bendichter is this still relevant?

PR #33, opened 2020-06-03
Gui refactor

Saksham20:
Planned changes to gui: 

- [x] fill in default paths in **'source paths'** box and **'output nwb file'** #31 
- [x] have .yaml also supported for loading and saving
- in forms: 
- [x] For required arguments: if unfilled, fill with placeholder with 'required class type'. 
- [x] For non required, if None then leave empty instead of 'None'. 
- [x] When loading metafile dict from form: omit keys for fields which are empty in form
- [x] looks directly for 'run_conversion' function and calls it with the set nwb path, source file path, and metadata file if passed in. 
- [ ] Expand form to incorporate IBL metadata file. Adding data fields.

### DIFF
--- a/nwb_conversion_tools/gui/nwb_conversion_gui.py
+++ b/nwb_conversion_tools/gui/nwb_conversion_gui.py
@@ -25,12 +25,14 @@ import importlib
 import yaml
 import sys
 import os
-
+import inspect
+# TODO: implement changes specified downstread
+#   -check kwargs
 
 class Application(QMainWindow):
     def __init__(self, metafile=None, conversion_module='', source_paths={},
                  kwargs_fields={}, extension_modules={}, extension_forms={},
-                 show_add_del=False):
+                 show_add_del=False, nwb_file=None, conversion_class=None):
         super().__init__()
         # Dictionary storing source files paths
         self.source_paths = source_paths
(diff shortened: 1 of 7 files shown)


nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -649,15 +655,20 @@ def __init__(self, parent):
 
     def run(self):
         #try:
-        mod_file = self.parent.conversion_module_path
-        spec = importlib.util.spec_from_file_location(os.path.basename(mod_file).strip('.py'), mod_file)
-        conv_module = importlib.util.module_from_spec(spec)
-        spec.loader.exec_module(conv_module)
+        if len(self.parent.conversion_module_path)>0:# if not an empty string (if value was selected from gui)
+            mod_file = self.parent.conversion_module_path
```
bendichter, line 659: do not input conversion_module_path. Input the class itself!

bendichter:
Please separate these out into individual PRs

nwb_conversion_tools/gui/nwb_conversion_gui.py
Saksham20: the class is being input, conversion_module path only exists if the user selects a path from the file drop down menu. If not, then it takes only the class into consideration

PR #34, opened 2020-06-04
directly call run_conversion

Saksham20:
use the class to call the run_conversion function within. 
However, i've still kept the older conversion_path, which is redundant, but will only be useful if the user selects another conversion script path from file drop down menu in the gui.

### DIFF
--- a/nwb_conversion_tools/converter.py
+++ b/nwb_conversion_tools/converter.py
@@ -23,13 +23,9 @@ class NWBConverter:
         """
         self.metadata = metadata
         self.source_paths = source_paths
-        self.nwbfile_saveloc = None
         # create self.nwbfile object
         if nwbfile is None:
             self.create_nwbfile(metadata['NWBFile'])
-        elif isinstance(nwbfile,str):
-            self.create_nwbfile(metadata['NWBFile'])
-            self.nwbfile_saveloc = nwbfile
         else:
             self.nwbfile = nwbfile
 
@@ -261,7 +257,7 @@ class NWBConverter:
                 index=index
             )
 
-    def save(self, to_path=None, read_check=True):
+    def save(self, to_path, read_check=True):
         """
         This method should not be overridden.
         Saves object self.nwbfile.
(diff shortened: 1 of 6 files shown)


nwb_conversion_tools/converter.py
```diff
@@ -23,10 +23,13 @@ def __init__(self, metadata, nwbfile=None, source_paths=None):
         """
         self.metadata = metadata
         self.source_paths = source_paths
-
+        self.nwbfile_saveloc = None
         # create self.nwbfile object
         if nwbfile is None:
             self.create_nwbfile(metadata['NWBFile'])
+        elif isinstance(nwbfile,str):
+            self.create_nwbfile(metadata['NWBFile'])
```
bendichter, line 31: nwbfile should not be a string.

nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -681,9 +692,13 @@ def wheelEvent(self, event):
...
 def nwb_conversion_gui(metafile=None, conversion_module='', source_paths={},
                        kwargs_fields={}, extension_modules={}, extension_forms={},
-                       show_add_del=False):
+                       show_add_del=False, nwbfile_loc=None, conversion_class=None):
     """Sets up QT application."""
     app = QtCore.QCoreApplication.instance()
+    if conversion_class is None and conversion_module=='':
+        raise Exception('provide one of conversion_module:str or conversion_class:class')
+    if nwbfile_loc is None:
+        nwbfile_loc = os.path.join(os.getcwd(),conversion_class.__name__ + '_nwbfile_gui.nwb')
```
bendichter, line 701: What is this? A default name for the NWB file? Can you make this blank? I want this to be filled in by the user

nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -25,12 +25,14 @@
...
 import sys
 import os
-
+import inspect
+# TODO: implement changes specified downstread
+#   -check kwargs
 
 class Application(QMainWindow):
     def __init__(self, metafile=None, conversion_module='', source_paths={},
                  kwargs_fields={}, extension_modules={}, extension_forms={},
```
bendichter, line 34: Null default values should be `None`. This applies to `conversion_module` and all of the subsequent `dict`s. I'd also like to mark conversion_module as depricated.

PR #35, opened 2020-06-04
auto populate source paths and nwbfile location, have yaml also supported

Saksham20:
source paths autopopulated. 
Also autopopulated the nwb file save location based on the current working directory. Can be changed by the user within gui.

### DIFF
--- a/nwb_conversion_tools/gui/nwb_conversion_gui.py
+++ b/nwb_conversion_tools/gui/nwb_conversion_gui.py
@@ -96,7 +96,7 @@ class Application(QMainWindow):
 
         self.lbl_nwb_file = QLabel('Output nwb file:')
         self.lbl_nwb_file.setToolTip("Path to the NWB file that will be created.")
-        self.lin_nwb_file = QLineEdit('')
+        self.lin_nwb_file = QLineEdit(self.nwb_file)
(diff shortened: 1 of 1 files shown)


Saksham20:
> Where is `self.nwb_file` assigned? Also, it should be called something like `self.output_nwb_filepath`. Calling it `self.nwb_file` suggests that it is a `File` object, like `h5py.File`.

so the `self.nwbfile_loc` gets assigned in the other branch, realized that later while splitting that main branch into three.  This branch should ideally have been on top of gui_fix1

nwb_conversion_tools/gui/nwb_conversion_gui.py
```diff
@@ -96,7 +96,7 @@ def init_meta_tab(self):
 
         self.lbl_nwb_file = QLabel('Output nwb file:')
         self.lbl_nwb_file.setToolTip("Path to the NWB file that will be created.")
-        self.lin_nwb_file = QLineEdit('')
+        self.lin_nwb_file = QLineEdit(self.nwbfile_loc)
```
bendichter, line 99:
```suggestion
        if self.nwbfile_loc is not None:
            self.lin_nwb_file = QLineEdit(self.nwbfile_loc)
```

PR #36, opened 2020-06-04
Gui fix3

Saksham20:
There are two fixes in this: (could not further break them into two seperate PRs without undoing some changes):
1. Required fields: if unfilled, will have a placeholder. This is in addition to the red asterisk. Just thought this might grab user attention more while filling up form. 
2. Not required fields: these have default values (defined in pynwb); but those that have a NoneType default, the form field now remains empty instead of filling with a text 'None'. This was necessary, since prior to this, while reading the form to write to the dict, it read it as a string instead of NoneType.

### DIFF
--- a/nwb_conversion_tools/gui/classes/forms_basic.py
+++ b/nwb_conversion_tools/gui/classes/forms_basic.py
@@ -50,7 +50,7 @@ class BasicFormCollapsible(CollapsibleBox):
             # Required fields get a red star in their label
             if 'default' not in field:
                 required = True
-                default = 'required: '+ str(field['type'])
+                default = ''
             # Optional fields
             else:
                 required = False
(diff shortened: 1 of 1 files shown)


Saksham20:
> Could you remove the autofill of required fields?

It does not autifill anything, its just a placeholder `QLineEdit().setPlaceholderText('text')` : greyed out text suggesting the type to input.
You mean you dont want the greyed out text?

nwb_conversion_tools/gui/classes/forms_basic.py
```diff
@@ -105,11 +105,20 @@ def make_forms(self):
...
+                else:
+                    form.setText(field['default'])
             # Float types
             elif field['type'] == 'float':
-                form = QLineEdit('')
-                form.setText(str(field['default']))
+                if field['default'] is not None:
+                    form = QLineEdit('')
+                    if field['required']:
+                        form.setPlaceholderText(str(field['default']))
```
bendichter, line 117: A default value should fill whether or not it is required. This applies throughout.

PR #38, opened 2020-06-14
send error type to nwb_gui logger

Saksham20:
- Is there an added functionality to having converter_module.py as a source of `run_conversion()` ?
- The functionality to display the error in the gui logger was commented out, is there  a reason for this?


bendichter:
closing because qt gui is being phased out and has moved to another repo

PR #42, opened 2020-06-22
change to forms_basic

Saksham20:
have cherry picked the changes corresponding to the forms_basic.py code, now a PR to master instead of SegExt-Integration

### DIFF
--- a/nwb_conversion_tools/gui/classes/forms_basic.py
+++ b/nwb_conversion_tools/gui/classes/forms_basic.py
@@ -50,9 +50,11 @@ class BasicFormCollapsible(CollapsibleBox):
             # Required fields get a red star in their label
             if 'default' not in field:
                 required = True
+                default = 'required: '+ str(field['type'])
             # Optional fields
             else:
                 required = False
+                default = field['default']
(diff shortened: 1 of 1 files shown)


nwb_conversion_tools/gui/classes/forms_basic.py
```diff
@@ -101,9 +105,20 @@ def make_forms(self):
             # String types
             if field['type'] == 'str':
                 form = QLineEdit('')
+                if field['required']:
+                    form.setPlaceholderText(field['default'])
+                else:
+                    form.setText(field['default'])
```
bendichter, line 111: I don't know the whole codebase, but these lines still seem wrong to me. If there is a default, that default should be entered into the field, not as a placeholder text, whether or not the field is required.

Saksham20:
@luiztauffer check [here ](https://github.com/catalystneuro/nwb-conversion-tools/pull/42/files#diff-9ef209691edfbf73b515bca105f5b4c6L52)

bendichter:
closing because the qt GUI is being phased out

PR #49, opened 2020-06-29
Change version

luiztauffer:
version for the package with new name


bendichter:
@luiztauffer I don't see why we need to change the version number

PR #53, opened 2020-07-07
Minor modifications to SpikeExtractor2NwbConverter

CodyCBakerPhD:
All changes enable custom column data from within the SortingExtractor objects to be written into the nwb file and format.

### DIFF
--- a/nwb_conversion_tools/ecephys/ecephys.py
+++ b/nwb_conversion_tools/ecephys/ecephys.py
@@ -103,47 +103,31 @@ class SpikeExtractor2NWBConverter(NWBConverter):
                         )
                 else:
                     if 'electrode_group' in self.SX.get_unit_property_names(unit_id=id):
-                        self.nwbfile.add_unit(id=id, spike_times=spkt, electrode_group=self.SX.get_unit_property(id,'electrode_group'))
+                        self.nwbfile.add_unit(id=id, spike_times=spkt,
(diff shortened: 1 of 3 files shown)


nwb_conversion_tools/ecephys/ecephys.py
```diff
@@ -70,22 +70,86 @@ def add_units(self):
...
+                        )
                 else:
-                    self.nwbfile.add_unit(id=id, spike_times=spkt)
+                    if 'electrode_group' in self.SX.get_unit_property_names(unit_id=id):
+                        self.nwbfile.add_unit(id=id, spike_times=spkt, electrode_group=self.SX.get_unit_property(id,'electrode_group'))
+                    else: 
+                        self.nwbfile.add_unit(id=id, spike_times=spkt)
+        # Extract and add custom column data from unit properties
+        descriptions = [
+            {
```
bendichter, line 111: This class should be generic to all SpikeExtractors, not specific to Neuroscope, so I'd rather not have this

CodyCBakerPhD: Fixed! Now assumes the unit property is a dictionary with all the relevant information specific to that dataset and unit.

CodyCBakerPhD:
Disregard; per our conversation, all of this will be used through SpikeInterface in the future.

Issue #54, opened 2020-07-15
Comments on NWBConverter


alejoe91:
Hi guys,

I have a couple of comments about the NWBConverter class and the organization of the repo.

The NWBConverter class contains a wide range of functions to populate the Nwb file and it serves as base class for all other converters.

However, it seems to me that there are 2 very distinct kind of converters:
- Domain-specific ones (e.g. `Spikeglx2NWB` or `Miniscope2NWB` ), that populate specific subfields of the NWB file (ecephys/microscopy)
- Lab-specific ones (e.g. `ToliasNWBConverter`), that should use the different other converters to make create the entire Nwb file

For the domain-specific converters, do we really need to inherit from the `NWBConverter` class? In my opinion, the domain-specific converters could be available as functions or standalone classes of the package (building blocks) that can be then used by the lab converters to construct and populate the Nwb file. 
Put it in another way, what is the purpose for say the `Miniscope2NWB` to inherit the `create_electrode_groups` method?

Another possibility, would be to have the basw `NWBConverter` class to handle file creation all the basic metadata (subject, user, etc.), and an inherited class would make use of these basic methods (that will be similar for all labs/cases) and add some lab-specific functionalities to convert domains. 

For example: 

```python
from nwb_conversion_tools import NWBConverter
from nwb_conversion_tools.ecephys import convert_intan
from nwb_conversion_tools.ophys import convert_miniscope
from nwb_conversion_tools.behavior import convert_bpod

class BuccinoLabNWBConverter(NWBConverter):
      def __init__(self, metadata, nwbfile):
              super().__init__(nwbfile=nwbfile, metadata=metadata, source_paths=source_paths)

      def run_conversion(intan_file, miniscope_file, bpod_file, metadata):
             # each function will take care of populating the right fields of the Nwb file only
             convert_intan(self.nwbfile, intan_file, metadata)
             convert_miniscope(self.nwbfile, miniscope_file, metadata)
             convert_bpod(self.nwbfile, bpod_file, metadata)

# usage
blab_converter = BuccinoLabNWBConverter(my_metadata, 'my_file.nwb')
blab_converter.run_conversion(intan_file, miniscope_file, bpod_file, my_metadata)
```

What do you guys think?

bendichter:
[...]

PR #59, opened 2020-08-24
source data schema and example

luiztauffer:
v 0.1.0


luiztauffer:
Here are tentative examples of source input data and schema. Those will be generated by the specific converters in the backend and returned to the GUI for creating the Forms in the first step: https://github.com/catalystneuro/nwb-conversion-tools/issues/54#issuecomment-663249122

@CodyCBakerPhD @bendichter can you check it out and let me know if these are covering all uses cases (files, directories and boolean options)? Or any other fields / types you foresee we might need?

bendichter:
@luiztauffer what do you think of this schema? (I accidentally merged, meant to create a PR)

bendichter:
I'm tempted to get rid of the source_files vs. conversion_options and just put them all in the same object. @luiztauffer, what would you think about that?

luiztauffer:
I included two modifications now on input data and schema handling:
1. when the input_data comes to the converter, it needs to know where to route each key:value pair
For example, this data will be received by a converter and should have specific parts correctly sent to an Ophys and an Ecephys data interfaces:
```json
{
	"source_data": {
		"path_calibration": "path_to/calibration.h5",
		"path_raw": "path_to/raw.h5",
		"path_tiff": "path_to/image.tif",
		"path_processed": "path_to/processed_data.h5"
	},
	"conversion_options": {
		"ophys_raw": true,
		"ophys_processed": true,
		"ecephys_spiking": true,
		"ecephys_processed": true
	}
}
```
this routing is important because of this:
https://github.com/catalystneuro/nwb-conversion-tools/blob/2a29c85ebd449a13356c167697387cf780324a60/nwb_conversion_tools/nwbconverter.py#L24-L25

luiztauffer:
2. when updating `input_schema['properties']` in series with several DataInterfaces, it needs to be a deep update. This must be working well now with these changes

bendichter:
@luiztauffer OK, can you fix merge conflicts?

luiztauffer:
> @luiztauffer OK, can you fix merge conflicts?

done

luiztauffer:
I believe this fixes #61 
@bendichter @CodyCBakerPhD ?

PR #66, opened 2020-10-07
updates on base converter

luiztauffer:
- avoid duplicate entries in lists (e.g. two datainterfaces have the same required item)
- get_metadata default behavior: loop through converter datainterfaces and call their own get_metadata
- add option to run_conversion, return nwbfile object instead of saving to file


nwb_conversion_tools/nwbconverter.py
```diff
@@ -13,6 +15,9 @@ def dict_deep_update(d, u):
             d[k] = dict_deep_update(d.get(k, {}), v)
         elif isinstance(v, list):
             d[k] = d.get(k, []) + v
+            # Remove repeated items if they exist
+            if len(v) > 0 and not isinstance(v[0], dict):
+                d[k] = list(np.unique(d[k]))
```
bendichter, line 20: maybe this would be a good place to use sets

nwb_conversion_tools/nwbconverter.py
```diff
@@ -56,7 +61,7 @@ def get_metadata_schema(self):
         """Compile metadata schemas from each of the data interface objects."""
         metadata_schema = get_root_schema()
         metadata_schema['properties'] = dict(
-            NWBFile=get_schema_from_hdmf_class(NWBFile),
+            NWBFile=get_schema_for_NWBFile(),
```
bendichter, line 64: why define this explicitly instead of generating it?

nwb_conversion_tools/nwbconverter.py
```diff
@@ -66,23 +71,46 @@ def get_metadata_schema(self):
...
-        with NWBHDF5IO(nwbfile_path, mode='w') as io:
-            io.write(nwbfile)
+        if save_to_file:
+            if nwbfile_path is not None:
+                raise TypeError('A path to the output file must be provided, but nwbfile_path got value None')
+            # run_conversion will always overwrite the existing nwbfile_path
+            with NWBHDF5IO(nwbfile_path, mode='w') as io:
+                io.write(nwbfile)
+            print(f'NWB file saved at {nwbfile_path}')
+            return None
```
bendichter, line 114:
```suggestion
```

nwb_conversion_tools/utils.py
```diff
@@ -1,12 +1,16 @@
...
+import numpy as np
 
 
-def get_base_schema():
+def get_base_schema(tag=None):
     base_schema = dict(
         required=[],
         properties={},
         type='object',
+        tag=['' if tag is None else tag][0],
```
bendichter, line 13: I'd rather have tag be optional

nwb_conversion_tools/utils.py
```diff
@@ -1,12 +1,16 @@
...
         additionalProperties=False
     )
```
bendichter, line 15:
```suggestion
    base_schema = dict(
        required=[],
        properties={},
        type='object',
        additionalProperties=False
    )
    if tag is not None:
        base_schema.update(tag=tag)
```

nwb_conversion_tools/utils.py
```diff
@@ -48,24 +50,155 @@ def get_schema_from_method_signature(class_method):
...
+            schema_arg[docval_arg['name']] = dict(type='number', description=docval_arg['doc'])
+
+        # type string
+        elif docval_arg['type'] is str or (isinstance(docval_arg['type'], tuple) and str in docval_arg['type']):
+            schema_arg[docval_arg['name']] = dict(type='string', description=docval_arg['doc'])
+
+        # type datetime
+        elif docval_arg['type'] is datetime or (isinstance(docval_arg['type'], tuple) and datetime in docval_arg['type']):
+            schema_arg[docval_arg['name']] = dict(type='string', description=docval_arg['doc'], format='date-time')
+
```
bendichter, line 76: This is good, but is there a way to refactor these lines so that you are repeating less code?

nwb_conversion_tools/utils.py
```diff
@@ -48,24 +50,155 @@ def get_schema_from_method_signature(class_method):
...
+        # if link to another nwb object
+        elif docval_arg['type'] in pynwb_link_types:
```
bendichter, line 78: I think the official way to determine this would be to parse cls.__nwbfields__ for `is_child`

nwb_conversion_tools/utils.py
```diff
@@ -48,24 +50,155 @@ def get_schema_from_method_signature(class_method):
...
+        else:
+            if not isinstance(docval_arg['type'], tuple):
+                docval_arg_type = [docval_arg['type']]
             else:
-                schema['required'].append(docval_arg['name'])
-            schema['properties'].update(schema_arg)
+                docval_arg_type = docval_arg['type']
+
+            # if another nwb object (or list of nwb objects)
+            if any([t.__module__.split('.')[0] == 'pynwb' for t in docval_arg_type if hasattr(t, '__module__')]):
```
bendichter, line 97: should use `isinstance` here I think

bendichter:
@luiztauffer really good work here! Mostly stylistic comments

@CodyCBakerPhD take a look, I think these changes will be relevant for you as well

nwb_conversion_tools/nwbconverter.py
luiztauffer: The general function `get_schema_from_docval()` will include many fields for the NWBFile object schema that will not be used by the converter nor the GUI. It would end up cluttering the NWBFile form with many fields that will not be used.
I haven't found yet a programmatic way of skipping these, while preserving the possibility that they might be used as fields for other objects (such as TimeIntervals or DynamicTable). This is a dirty temporary solution. 

image

nwb_conversion_tools/utils.py
luiztauffer: I couldn't find reference to this `is_child` property anywhere in pynwb nor hdmf docs. How would you go about using that? Take for example `pynwb.ophys.ImagingPlane`, we need to set the schema differently for:
- `optical_channel` (the pynwb object received will be indeed a child)
- `device` (creates a link to the pynwb object received)

bendichter: Oh, my fault, it's just `'child'`, e.g.

https://github.com/NeurodataWithoutBorders/pynwb/blob/c4631e9e8e09a46e427442fd173818fb40414283/src/pynwb/ecephys.py#L51

When a class's `__nwbfields__` has an attribute with `child==True`, it's not a link. When `child==False`, it's a link. If we want to use this, I think we'd have to change this function to point to the class instead of the docval.

nwb_conversion_tools/nwbconverter.py
bendichter: What about adding a `skip` arg to `get_schema_from_hdmf_class` that's a list of names of the arguments to skip

CodyCBakerPhD:
Looks like I'm one PR behind on checking this, but the lines you had added in the `NwbConverter` `__init__`

```
self.data_interface_objects = dict()
        input_data_routed = dict()
        for interface_name, interface in self.data_interface_classes.items():
            input_data_routed[interface_name] = dict()
            interface_schema = interface.get_input_schema()
            blocks = ['source_data', 'conversion_options']
            for b in blocks:
                if b in interface_schema:
                    input_data_routed[interface_name][b] = {
                        k: input_data[b][k]
                        for k in interface_schema[b]['properties'].keys()
                    }

        self.data_interface_objects = {name: data_interface(**input_data_routed[name])
```

breaks all current conversion scripts using the new format, even after adjusting the input format to match the example.

This is primarily because `interface.get_input_schema()` really just calls `get_schema_from_method_signature(...)` or the like on the internal data class (a RecordingExtractor, for instance), but this does not follow the nesting structure that the new `blocks` expects.

I'll work on fixing these issues on this current PR and will let you know when I get it working; for now, I've made a separate branch that temporarily reverts the above code block so I can run conversions in the meantime.

luiztauffer:
@CodyCBakerPhD what's the input schema you're using?

This piece of code should work with a schema like this: https://github.com/catalystneuro/nwb-web-gui/blob/master/nwb_web_gui/dashapps/pages/converter/example_schemas/schema_source.json

CodyCBakerPhD:
@luiztauffer For example,

`interface_schema = interface.get_input_schema() `

where `interface=NeuroscopeRecordingInterface`, which simply specifies that `cls.RX=NeuroscopeRecordingExtractor`, whereby `interface.get_input_schema()` simply calls `get_schema_from_method_signature(cls.RX)`, which certainly returns the correct information about what types of properties fields would be necessary, but does not nest the blocks by `'source_data'` or `'conversion_options'`.

I understand the `schema_source.json` is what the form/GUI side requires, but the actual data interfaces used need to adhere to it in that case, which is what I'll change on this branch.

luiztauffer:
`source_data` and `conversion_options` should be standard keywords used by any converter.
The first contain paths to source directories/files. The second boolean options.
We can change names or how these things should be organize in general, but we should agree on a standard

CodyCBakerPhD:
I am adhering to whatever standard is needed by the forms; it's just the data interfaces themselves that therefore need to be changed in tandem to match it.

luiztauffer:
got it! let me know if you face any limitation with this schema

CodyCBakerPhD:
@luiztauffer Was there a reason behind the removal of

`self.input_args = input_args`

from that previous PR?

luiztauffer:
> @luiztauffer Was there a reason behind the removal of
> 
> `self.input_args = input_args`
> 
> from that previous PR?

where was this exactly?
this line still exists on `BaseDataInterface`

CodyCBakerPhD:
https://github.com/catalystneuro/nwb-conversion-tools/pull/59/files#diff-91a2f17a5200d62cf54bb3dc0051c2df6c96650f247faf2bc514327211840919L24-L25

In the NWBConverter class. I do actually make use of them the stored values on occasion to reduce the re-use of input arguments like a common `folder_path`; but if the ones in the data interfaces still exist I can try to adjust to use those.

CodyCBakerPhD:
OK, I've debugged everything that was causing issues with the conversion scripts. Looks like there are a still a couple stylistic things yet to change in the PR, though.

bendichter:
@luiztauffer we should implement this in `get_schema_from_hdmf_class` and remove `get_schema_from_docval`

nwb_conversion_tools/nwbconverter.py
```diff
@@ -66,23 +71,47 @@ def get_metadata_schema(self):
...
         nwbfile = NWBFile(subject=subject, **metadata_dict['NWBFile'])
 
-        [data_interface.convert_data(nwbfile, metadata_dict[name], stub_test)
-         for name, data_interface in self.data_interface_objects.items()]
+        # add devices
+        for domain in ('Icephys', 'Ecephys', 'Ophys'):
+            if domain in metadata_dict:
+                dev_keys = [k for k in metadata_dict[domain].keys() if 'Device' in k]
+                for dev in dev_keys:
+                    nwbfile.create_device(**metadata_dict[domain][dev])
```
bendichter, line 103:
```suggestion
                [nwbfile.create_device(**metadata_dict[domain][obj]) for obj in metadata_dict[domain] if 'Device' in obj]
```

nwb_conversion_tools/utils.py
```diff
@@ -32,40 +37,182 @@ def get_input_schema():
...
+        if docval_arg['type'] == 'float' or (isinstance(docval_arg['type'], tuple) and 'float' in docval_arg['type']):
+            schema_arg[docval_arg['name']] = dict(type='number', description=docval_arg['doc'])
+
+        # type string
+        elif docval_arg['type'] is str or (isinstance(docval_arg['type'], tuple) and str in docval_arg['type']):
+            schema_arg[docval_arg['name']] = dict(type='string', description=docval_arg['doc'])
+
+        # type datetime
+        elif docval_arg['type'] is datetime or (isinstance(docval_arg['type'], tuple) and datetime in docval_arg['type']):
+            schema_arg[docval_arg['name']] = dict(type='string', description=docval_arg['doc'], format='date-time')
```
bendichter, line 79:
```suggestion
        schema_arg = {docval_arg['name']: dict(description=docval_arg['doc'])}

        # type float
        if docval_arg['type'] == 'float' or (isinstance(docval_arg['type'], tuple) and 'float' in docval_arg['type']):
            schema_arg[docval_arg['name']].update(type='number')

        # type string
        elif docval_arg['type'] is str or (isinstance(docval_arg['type'], tuple) and str in docval_arg['type']):
            schema_arg[docval_arg['name']].update(type='string')

        # type datetime
        elif docval_arg['type'] is datetime or (isinstance(docval_arg['type'], tuple) and datetime in docval_arg['type']):
            schema_arg[docval_arg['name']].update(type='string', format='date-time')
```

luiztauffer:
@CodyCBakerPhD could you please check if efa96386eca3ff1de985312774fdaf46ffa49e3d breaks anything in your DataInterfaces?
DataInterfaces method `get_metadata()` should receive a metadata dictionary with partial metadata retrieved by other DataInterfaces (it can be an empty dictionary). This is relevant if we have multiple DataInterfaces that can potentially get the same metadata info

CodyCBakerPhD:
@luiztauffer The principle of `efa9638` was discussed between Ben and I a while back, and I think we landed on the current version where it is the burden of the conversion specific NWBConverter class to auto-fill all metadata that gets passed off to the data interfaces; while adding the argument in `efa9638` wouldn't fundamentally break anything, the change in `2b6c8f8` will, so I'd prefer to keep that unchanged until we can further discuss the strategy.

Also, the principle of that change seems beyond the scope of this PR, and would likely be better served being introduced on a different branch.

luiztauffer:
@CodyCBakerPhD @bendichter 
I don’t follow, it seems to me that each DataInterface should be responsible to fetch the metadata available in their respective data files. Otherwise, why would we have DataInterfaces?

A problem arises when specific metadata info (such as NWBFile ou Subject fields) could potentially be retrieved by any two DataInterfaces. This is easily solved with passing on the partially filled metadata and letting the DataInterface check whether it should fill certain parts of it or not.

About 2b6c8f8739afdb9bf8b70c6aafaa432e66318cc8, which change exactly is problematic?
The metadata that gets passed to `converter.run_conversion()` might receive session_start_time as a ISO 8601 string (e.g. from Forms), but we can put a test if it comes as a datetime already.
The other change is because the metadata receive will comply to the schema, and the schema does not have keys for the converters names.

CodyCBakerPhD:
@luiztauffer The problem lies in generality of the data interfaces. I understand you are currently using a high specified data interface for your conversion, but I tend to rely on ones that ought to apply to general structures such as those in `SpikeInterface`.

Here's an example: we have a `BaseRecordingExtractorInterface`, and a specific `NeuroscopeRecordingExtractorInterface` (that merely specifies that the type of the recording extractor is a Neuroscope, nothing really beyond that). The `RecordingExtractor` uses a lot of metadata passed in the structure required by the `NWBRecordingExtractor.write_recording` format (which has _not_ changed over the previous refactors), so it has to get this metadata from somewhere. The `BaseRecordingExtractorInterface` would be impossible to generalize a `get_metadata()` function over all of its subinterfaces, since each interface is associated with a different format. At the next level, if you're proposing that the NeuroscopeRecordingExtractorInterface be responsible for autopopulation via  `get_metadata()`, this is also problematic because not every experiment conversion using Neuroscope stores this metadata in the same fashion (it generally uses .xml, but can also use .mat on occasion, all depending on the sessions to be converted).

To further illustrate then why the converter class ought to carry this burden, two specific instances of conversion would be the Yuta and Watson datasets, which have their own custom `YutaNWBConverter` and `WatsonNWBConverter` subclass of the `NWBConverter` class, and both use the common `NeuroscopeRecordingExtractorInterface` with metadata received from their `NWBConverter` classes. However, parts of this metadata from Yuta comes from both `.xml` and `.xlsx` files, whereas Watson has some in `.xml` and `.mat` files. Hence the Neuroscope specific Data Interface could not reliably determine where and how to access the relevant metadata, but the session/experiment specific Converter class can.

CodyCBakerPhD:
@luiztauffer On the 'issue' of multiple data interfaces using the same common metadata, it's true that is a slight inefficiency to duplicate information but generally these metadata are not very expensive in their size and scope. I've run into that case myself, but just repeating certain fields a couple times isn't going to cause any serious problems.

Again though, I really think we need to focus on just getting the more basic changes through on this PR so we can make sure the master branches of various conversions are compatible with the master branch of `nwb-conversion-tools` (which is not currently the case). This discussion of metadata efficiency would be better off on a separate branch and PR where it does not immediately effect whether or not certain active conversions can run.

luiztauffer:
@CodyCBakerPhD 
I agree we should merge this asap and make a release, I think we can do it today. I also agree with moving the metadata fetching discussion to the next PR. Lets settle these two immediate points before finalizing this PR:

1. About commit cb42eca4eec8d810a4ae7da38a5f5d7adaf5f953 :
To me it seems unnecessarily confusing to have two `input_schema`, one with and one without the `source_data` and `conversion_options` keywords. Although we flattened input_args for the `convert_data()` method to keep older code from breaking, I believe anytime we get a `schema`, for example by calling `DataInterface.get_input_schema()`, the returned object should comply to the one defined `schema`, be that input or metadata. Since this wasn’t breaking any current converters, I think we should reverse those changes.

2. About commit 2b6c8f8739afdb9bf8b70c6aafaa432e66318cc8 :
what exactly breaks previous code?

bendichter:
Hey guys, great convo. Sorry I'm late here. I'd like to weigh in, but at this point I think both of you are in this deeper than I am so feel free to push back if what I am suggesting is infeasible. 

**Handling metadata duplication**
The job of the NwbConverter is to coordinate across the DataInterfaces. If there are conflicts or duplicates of `get_metadata` or `get_metadata_schema` between the DataInterfaces, it is the job of the Converter to handle those conflicts by de-duplicating them, and potentially overwriting. 

**Separating the options from the source**
It's fine to separate out the source and option fields in the Form, but I'd rather not require that the DataInterfaces comply with that because it would be difficult to automatically generate a schema based on the function signature and I would like to have the ability to do that. If you want to render the form separating out source and option fields, that should also be the job of the Converter object to separate those out when serving to the Form, and recombine the attributes when feeding back to the DataInterface. Does that create any problems?


**Implementation of `get_metadata`**

`get_metadata` should be implemented by the DataInterface.

> this is also problematic because not every experiment conversion using Neuroscope stores this metadata in the same fashion (it generally uses .xml, but can also use .mat on occasion, all depending on the sessions to be converted).

Neuroscope uses .xml for metadata and binary for data. Any .mat or .xls files present are additional non-compliant metadata files added by the experimenter, These should not be considered to be part of the Neuroscope format and should be loaded in separately. The `NeuroscopeRecordingExtractorInterface` _should_ be responsible for `get_metadata`, and should pull it from the xml file.

luiztauffer:
**Separating the options from the source**
> It's fine to separate out the source and option fields in the Form, but I'd rather not require that the DataInterfaces comply with that because it would be difficult to automatically generate a schema based on the function signature and I would like to have the ability to do that. If you want to render the form separating out source and option fields, that should also be the job of the Converter object to separate those out when serving to the Form, and recombine the attributes when feeding back to the DataInterface. Does that create any problems?

Don’t worry about the Forms, it would be easy to reassemble these by type boolean or string for the nwb-web-gui, if it is the case that we remove the `source_data` and `conversion_options` keys from the `input_schema`. My point is we should agree on a `input_schema`, and methods named `get_input_schema()` should comply to that, to avoid confusion. 
Those keywords can be removed and used only on Forms without any problems, but I feel they might be useful in separating two distinct types of information, so I think they would be useful in the `input_schema`.
Generating a schema based on the function signature only fills in `source_data`, isn’t it so? The `conversion_options` need to be created by the DataInterface (data specific options) or the Converter (lab-specific options).
This is more of a feeling than a technical issue. That said, if there's no further comments on that, I'll assume we're removing `data_source` and `conversion_options` from `input_schema`.

**Implementation of get_metadata**
> Neuroscope uses .xml for metadata and binary for data. It's true, there are .mat and/or .xls files present, but they are additional non-compliant metadata files added by the experimenter. These file should not be considered to be part of the Neuroscope format and should be loaded in separately. The NeuroscopeRecordingExtractorInterface should be responsible for get_metadata, and should pull it from the xml file.

We could potentially break `Converter.get_metadata()` into `get_interfaces_metadata()`, which wouldn't be overwritten, and `get_lab_metadata()`, which would be Lab specific and resposible to fetch metadata that exists outside of the scope of specific DataInterfaces, for example lab specific details in csv or mat files

bendichter:
Why aren't `conversion_options` inputs to the DataInterface.__init__? Alternatively, one way we could sort this out is to have `source_data` be parameters of the `DataInterface` and `conversion_options` be optional args of the conversion method within that DataInterface. Would that be easier?

> We could potentially break Converter.get_metadata() into get_interfaces_metadata(), which wouldn't be overwritten, and get_lab_metadata(), which would be Lab specific and resposible to fetch metadata that exists outside of the scope of specific DataInterfaces, for example lab specific details in csv or mat files

We could do that, but wouldn't using `super()` like we generally do for `__init__` have the same effect?

CodyCBakerPhD:
@bendichter @luiztauffer Great discussion all, let's continue it on the next PR devoted exclusively to the metadata changes since that will require quite a refactor of the current data interfaces. Unless you have some final issues with the last 3 commits, then this ought to be good for merge as far as I'm concerned.

I finished removing the block structure from the input data routing, but I did keep the input schema related changes as I think those are a good direction, and appropriate to the spirit of this PR. Removed some of the excess metadata handling stuff as well, but feel free to add that back into the metadata PR.

I do think that the conversion_options debate would tend to be something much more suited as additional arguments passed into either the init of the interface or the run_conversion option. I use something exactly like that for deciding both whether to write out a stubbed (truncated) NWBFile, or whether or not to include spike waveforms.

luiztauffer:
> We could do that, but wouldn't using super() like we generally do for __init__ have the same effect?

You're right, that should do it!

I already opened another branch with the changes I think are important. I can't merge this since I'm the author, one of you need to give the approving review and merge it to master.

CodyCBakerPhD (approved):

PR #67, opened 2020-10-21
added interface for spikeglx recording extractor

CodyCBakerPhD:
Relevant to the tank-lab-to-nwb project

### DIFF
--- a/nwb_conversion_tools/datainterfaces/__init__.py
+++ b/nwb_conversion_tools/datainterfaces/__init__.py
@@ -1,0 +1,2 @@
+from .neuroscopedatainterface import NeuroscopeRecordingInterface, NeuroscopeSortingInterface
+from .spikeglxdatainterface import SpikeGLXRecordingInterface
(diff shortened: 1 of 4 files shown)


bendichter:
I definitely want to add this, but I'm just thinking about the best placement. We'll probably have dozens of interfaces like this, so let's make sure this is organized. I like the way spikeextractors is organized, having and `extractors` folder and having all the of extractors live in there, with an `__init__.py` that forwards all of the imports. So for here, that would mean:

1. This code would live in `nwb_conversion_tools/datainterfaces/spikeglxinterface.py`
2. There would be a file `nwb_conversion_tools/datainterfaces/__init__.py` that reads:

```python
from .spikeglxinterface import SpikeGLXRecordingInterface`
```

@CodyCBakerPhD what do you think of that?

CodyCBakerPhD:
@bendichter Yeah, that sounds great. I'll go ahead and include that organization structure in this PR

CodyCBakerPhD:
@bendichter OK, should be good now. All future specific instances of data interfaces will now go into that sub-folder for better organization. Only the base classes will remain in their current position.

PR #70, opened 2020-10-28
add meta programming for RecordingExtractor

bendichter:
Here's a way of doing metaprogramming that should take care of all of the recordingextractors

### DIFF
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -55,3 +55,30 @@ class BaseRecordingExtractorInterface(BaseDataInterface):
         se.NwbRecordingExtractor.write_recording(recording_extractor,
                                                  nwbfile=nwbfile,
                                                  metadata=metadata_dict)
+
+
+def recording_extractor_instance_to_data_interface_class(rx):
+    RX = rx.__class__
+    RX_name = rx.__class__.__name__
+
+    def __init__(self, **input_args):
+        BaseDataInterface.__init__(**input_args)
+        self.recording_extractor = rx
+
+    return type(  # create new class ( a lesser-known functionality of the `type` function)
+        RX_name + 'DataInterface',  # class name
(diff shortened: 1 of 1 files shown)


bendichter:
With this code we could use

`IntanRecordingExtractorDataInterface = recording_extractor_class_to_data_interface_class(IntanRecordingExtractor)`

then we wouldn't have to worry about declaring classes for each `RecordingExtractor`

bendichter:
This uses the `type` command in a lesser-known way to dynamically define a new class

CodyCBakerPhD:
This is a really interesting use of syntax; my question, now that stuff like that `NeuroscopeRecordingExtractor` will contain, importantly, its own internal functions for `get_metadata()` retrieval, how then would we append a function to a class made in this way? Would it be something like

`
def get_metadata()
    metadata = ...  # something retrieving metadata in class specific way
    return metadata

def recording_extractor_class_to_data_interface_class(RX, get_metadata):
    return type(
        RX.__name__ + 'DataInterface',  # class name
        (BaseRecordingExtractorInterface,),  # base classes
        dict(RX=RX,
               get_metadata=get_metadata)
    )
`

bendichter:
> This is a really interesting use of syntax; my question, now that stuff like that `NeuroscopeRecordingExtractor` will contain, importantly, its own internal functions for `get_metadata()` retrieval (from Issues [#2](https://github.com/catalystneuro/to-nwb-template/issues/2) and [#5](https://github.com/catalystneuro/to-nwb-template/issues/5) of nwb-conversion-template), how then would we append a function to a class made in this way? Would it be something like
> 
> ```
> def get_metadata()
>     metadata = ...  # something retrieving metadata in class specific way
>     return metadata
> 
> def recording_extractor_class_to_data_interface_class(RX, get_metadata):
>     return type(
>         RX.__name__ + 'DataInterface',  # class name
>         (BaseRecordingExtractorInterface,),  # base classes
>         dict(RX=RX, get_metadata=get_metadata)
>     )
> ```

yup!

PR #72, opened 2020-11-04
Implemented SIPickleDataInterfaces

alejoe91:
This PR adds the `SIPickleRecordingExtractorInterface` and `SIPickleSortingExtractorInterface`. 

This allows to process the data in the SI pipeline (e.g. synchronizing the recording and curating the sorting). Using the `se.load_extractor_from_pickle()` function, the converter is now capable of retrieving the state of the processed extractors and save it to NWB.

@bendichter @CodyCBakerPhD @luiztauffer what do you guys think?


bendichter:
Would it make more sense to inherit from https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/baserecordingextractorinterface.py?

alejoe91:
I looked into it, but not really because the pickle loader doesn't really have an `RX`, and the input is only the `pkl_file`

alejoe91:
It makes sense though to have a `BaseSIPickleDataInterface`, since the `get_metadata_schema` is the same.

make_env.yml
```diff
@@ -1,4 +1,4 @@
-name: nwb_conversion_env
+name: catalyst
```
alejoe91, line 1:
```suggestion
name: nwb_conversion_env
```

setup.py
```diff
@@ -25,9 +25,9 @@
     include_package_data=True,
     install_requires=[
         'pynwb', 'tqdm', 'natsort', 'numpy', 'scipy', 'bs4',
-        'pandas', 'jupyter', 'matplotlib', 'h5py', 'pyyaml', 'jupyter-client',
+        'pandas', 'jupyter', 'matplotlib', 'h5py', 'pyyaml',
```
alejoe91, line 28: @bendichter I changed the order here because I was getting some conflicts otherwise

CodyCBakerPhD:
I'll check through this, add that magic function we talked about, and refactor stuff within the next 24 hours so we can get this through and cleared off the ToDo stack.

setup.py
```diff
@@ -25,9 +25,9 @@
...
         'spikeextractors', 'spikesorters', 'spiketoolkit',
-        'PySide2', 'nwbwidgets', 'psutil', 'voila'
+        'PySide2', 'psutil', 'voila', 'jupyter-client'
```
bendichter, line 30:
```suggestion
```

PR #68, opened 2020-10-27
Schema compliance

luiztauffer:
Necessary changes for compliance with input and metadata schemas and manipulation.

PR #73, opened 2020-11-04
Refactor interface metadata

CodyCBakerPhD:
@bendichter The first of the major metadata handling changes is included in this PR. Subsequent ones will include the `NeuroscopeSorting` and `SpikeGLXRecording`.

Overall changes are built to reflect the structural shift of metadata auto-population away from specific `NWBConverter`s into the individual `DataInterfaces`.

This will render several past conversion scripts in other repositories incompatible with the current design; I have locally saved a branch of the current pre-metadata patch on my end for quick side-reversion, if that is ever needed for those conversions.

@luiztauffer I will also be uploading an example `RecordingExtractor` template tomorrow that uses this structure and showcases how, per our discussion, it ought to flow to keep an eye out for that. Also let me know if this PR breaks anything else on your end currently.

I also moved the `deep_dict_update` function to the `utils` as it is very useful for interacting with the new metadata structure (that is no longer nested by data interface name).

### DIFF
--- a/nwb_conversion_tools/basedatainterface.py
+++ b/nwb_conversion_tools/basedatainterface.py
@@ -17,7 +17,7 @@ class BaseDataInterface:
         pass
 
     @abstractmethod
-    def get_metadata(self, metadata):
+    def get_metadata(self):
         pass
 
     @abstractmethod
--- a/nwb_conversion_tools/nwbconverter.py
+++ b/nwb_conversion_tools/nwbconverter.py
@@ -5,22 +5,6 @@ from pynwb import NWBHDF5IO, NWBFile
 from pynwb.file import Subject
 from datetime import datetime
 import uuid
-import collections.abc
-import numpy as np
-
-
-def dict_deep_update(d, u):
-    for k, v in u.items():
-        if isinstance(v, collections.abc.Mapping):
-            d[k] = dict_deep_update(d.get(k, {}), v)
-        elif isinstance(v, list):
-            d[k] = d.get(k, []) + v
-            # Remove repeated items if they exist
-            if len(v) > 0 and not isinstance(v[0], dict):
-                d[k] = list(np.unique(d[k]))
-        else:
-            d[k] = v
-    return d
 
 
 class NWBConverter:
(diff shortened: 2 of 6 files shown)


nwb_conversion_tools/basedatainterface.py
```diff
@@ -17,7 +17,7 @@ def get_metadata_schema(self):
         pass
 
     @abstractmethod
-    def get_metadata(self, metadata):
+    def get_metadata(self):
```
CodyCBakerPhD, line 20: The usage of the downstream retrieval functions work to return a new dictionary object that the general converter then merges via deep update, so passing as a pointer to each data interface is no longer required.

nwb_conversion_tools/nwbconverter.py
```diff
@@ -97,7 +78,7 @@ def run_conversion(self, metadata_dict, nwbfile_path=None, save_to_file=True, st
 
         # Run data interfaces data conversion
         for name, data_interface in self.data_interface_objects.items():
-            data_interface.convert_data(nwbfile, metadata_dict[name], stub_test)
+            data_interface.convert_data(nwbfile, metadata_dict, stub_test)
```
CodyCBakerPhD, line 81: This is the main line that breaks previous converters; they could be easily refactored by adjusting to the new standard, though the majority of them are already uploaded to DANDI so I assume that isn't a priority.

luiztauffer:
should we merge this with #68 ?

bendichter:
Can you bump the version?

CodyCBakerPhD:
@luiztauffer If it's OK with you, I think that PR has a much broader scope and purpose (dealing with the input argument changes, as well as both input and metadata schema) so I'd rather not clutter it up with extra stuff. This branch is pretty self-contained and is still consistent with the [markdown for the metadata section](https://github.com/catalystneuro/nwb-conversion-tools/blob/schema_compliance/documentation/conversion_tools_structure.md) there, so it's also in line in terms of compatibility.

CodyCBakerPhD:
@bendichter Updated from 0.3.4 to 0.4.0 due to break in backwards compatibility. Is that correct?

bendichter:
> @bendichter Updated from 0.3.4 to 0.4.0 due to break in backwards compatibility. Is that correct?

The short answer is yes, 0.4.0 is a good choice. The long answer is that semantic versioning generally requires a major version bump (the first number) for breaking changes, however this does not apply in alpha mode, when the major version is 0. See [rule 4](https://semver.org/#spec-item-4) and [rule 8](https://semver.org/#spec-item-8)

PR #74, opened 2020-11-05
conversion_options routing

CodyCBakerPhD:
@bendichter @alejoe91 Simply adds an extra input argument to the initializer for the SpikeGLXRecordingInterface, synchronizing via a SubRecordingExtractor applied to the start frame identified by the first rising pulse.


nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -1,8 +1,29 @@
...
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
 
 
 class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
     RX = SpikeGLXRecordingExtractor
+
+    @classmethod
+    def get_input_schema(cls):
+        input_schema = super().get_input_schema()
+        input_schema['properties'].update(sync_with_ttl=dict(type='string', default=True))
```
bendichter, line 13: Seems like the type should be boolean ([ref](https://json-schema.org/understanding-json-schema/reference/boolean.html))

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -1,8 +1,29 @@
...
+            input_args.pop('sync_with_ttl')
+        else:
+            sync_with_ttl = False
+        super().__init__(**input_args)
+        if sync_with_ttl:
+            recording_ap = self.recording_extractor
+            ttl, states = recording_ap.get_ttl_events()
+            rising_times = ttl[states == 1]
+            start_time = recording_ap.frame_to_time(rising_times[0])
+            start_frame_ap = int(recording_ap.time_to_frame(start_time))
```
bendichter, line 28: Why are you doing frame_to_time and then time_to_frame?

bendichter:
@CodyCBakerPhD Overall looks good. Some questions about the implementation though

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
CodyCBakerPhD: Correct, fixed. Although, check comment below.

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
CodyCBakerPhD: Removed; all we need is rising_times[0]

CodyCBakerPhD:
@bendichter I did have another idea for implementation. This can either be as it is now in the `__init__`, OR the `synch_with_ttl` option can be a `conversion_option` (with an eye towards the upcoming input argument structure change) and this code can then switch to being called on `my_converter.run_conversion()` right before handing it off to the `super().convert_data`.

Which version do you prefer?

CodyCBakerPhD:
@bendichter Especially after the meeting today I'm leaning towards `conversion_options`... Also, how would you feel about the main `NWBConverter.run_conversion()` having dynamic `**conversion_options` as an argument to pass these sort of selective/specific options down to specific interfaces? Otherwise, every time we want to add that kind of option, we'd have to add it manually kind of like the way `write_waveforms()` works at the moment (that could be absorbed dynamically too).

bendichter:
@CodyCBakerPhD I agree, that all sounds good

CodyCBakerPhD:
@bendichter OK, moved it all to the conversion step.

@luiztauffer You should check this out if you have time; allowing the dynamic inputs to `run_conversion` would be perfect for your `conversion_options` schema, as it allows all data interfaces to see and use any interface-specific options via the dictionary pointer. There is one pathalogical use-case that could break this if two different interfaces have the same name for an option, but different intended uses for that option, but since we have full control over these interfaces it should be easy to maintain that standard (examples include `sync_with_ttl` and `write_waveforms`, with `stub_test` being universally used).

bendichter:
> There is one pathalogical use-case that could break this if two different interfaces have the same name for an option
We could handle that on the NWBConverter level by giving them prefixes going up and taking the prefixes away going down

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -1,8 +1,17 @@
...
+from spikeextractors import SpikeGLXRecordingExtractor, SubRecordingExtractor
 
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
 
 
 class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
     RX = SpikeGLXRecordingExtractor
+
+    def convert_data(self, nwbfile, metadata_dict: None, stub_test: bool = False, sync_with_ttl: bool = True,
+                     **conversion_options):
```
bendichter, line 11: Do we need to have the **conversion_options here too? Makes sense in the abstract class but I don't know if it's needed here, unless it is getting passed to another function

bendichter:
@CodyCBakerPhD what if we pulled the conversion args specific to the schema?

CodyCBakerPhD:
@bendichter Ahh, so if we handle the input arguments, metadata, and conversion options essentially the same way where they fetch the relevant values from a schema, then route them all appropriately to the corresponding downstream classes? Sure I can design it that way.

I assume it would be nice if all the relevant schemas were full `json`s as well, like the input schemas?

bendichter:
@CodyCBakerPhD can you resolve these conflicts?

CodyCBakerPhD:
@bendichter Did you want me to table the schema based conversion option stuff for later?

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
bendichter, line 11:
```suggestion
    def convert_data(self, nwbfile, metadata_dict: None, stub_test: bool = False, sync_with_ttl: bool = True):
```

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -9,6 +9,15 @@ class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
 
     RX = SpikeGLXRecordingExtractor
 
+    def convert_data(self, nwbfile, metadata_dict: None, stub_test: bool = False, sync_with_ttl: bool = True,
+                     **conversion_options):
+        """Primary function for converting recording extractor data to nwb."""
+        if sync_with_ttl:
+            ttl, states = self.recording_extractor.get_ttl_events()
+            rising_times = ttl[states == 1]
+            self.recording_extractor = SubRecordingExtractor(self.recording_extractor, start_frame=rising_times[0])
```
bendichter, line 18: Could we move this to the `BaseRecordingExtractorInterface`?

CodyCBakerPhD: Unfortunately, no. The `.get_ttl_events()` is a function specific to the IntanRecordingExtractor.

bendichter: ah, ok. @alejoe91 would it be possible to implement that method more generally?

bendichter: Looks like get_ttl_events has been implemented for several RecordingExtractor types but not all of them. It looks like there was some method to this but it's not obvious to me. I'll resolve this comment, since this is a SpikeExtractors issue.

CodyCBakerPhD: OK; I'll keep it in mind though moving forward, if we get some dev refactor downtime away from conversion, this ought to at least in principle be generalizable.

bendichter: Actually it looks like `get_ttl_events` is implemented in the `RecordingExtractor` here: https://github.com/SpikeInterface/spikeextractors/blob/53b145e4b2d4f58ba34bcdcf6599352c2e7212cf/spikeextractors/recordingextractor.py#L768 so we should be able to call it more generally. It'll just raise a `NotImplementedError` if it isn't implemented for that subtype

CodyCBakerPhD: Ohh ok, I'll move this over then

nwb_conversion_tools/nwbconverter.py
```diff
@@ -60,25 +60,29 @@ def get_metadata(self):
...
         """Build nwbfile object, auto-populate with minimal values if missing."""
         nwbfile_kwargs = dict(
-                session_description="no description",
-                identifier=str(uuid.uuid4()),
-                session_start_time=datetime.now()
-            )
-
+            session_description="no description",
+            identifier=str(uuid.uuid4()),
+            session_start_time=datetime.now()
```
bendichter, line 69: Can we not write incorrect data please?

bendichter, line 69:
```suggestion
```

CodyCBakerPhD: session_description should not be filled if not informative, and session_start_time auto-fills the same way through pynwb. We still need the identifier here though, correct?

bendichter: session_description and session_start_time are both required arguments

bendichter: session_start_time does not autofill

CodyCBakerPhD: So the session_description should be filled in only by a specific sub-class of the NWBConverter, or otherwise modified by a user via the metadata fields? and session_start_time, identifier always being populated within this main class?

CodyCBakerPhD: Oh, sorry... I was confusing sess_start_time with that file_create_time field

CodyCBakerPhD: So it looks like this block of code is there to ensure the creation of an NWBFile, even if the metadata['NWBFile'] doesn't have the fields session_description or session_start_time, with the values listed here as the defaults. Would you instead just want it to throw an error if those fields are not present in the metadata?

CodyCBakerPhD:
@bendichter As soon as we resolve the question of whether the converter should make the nwbfile with defaults session info, this PR should be good for final review

CodyCBakerPhD:
@bendichter Actually, per our conversation today, this PR will become focused on the conversion_option routing and will not include automated synchronization (which may not always dependably agree with user-desired alignment)

CodyCBakerPhD:
@bendichter Pending the decision on how best to include default NWBFile arguments, this PR should be good for final review. All changed focus on the conversion option routing, which is very similar to the input argument routing (indeed, they now share a common util function, too!).

nwb_conversion_tools/nwbconverter.py
bendichter: `session_description` can be safely filled with "no description". `session_start_time` is tricky though. It needs to be a `datetime`, and I don't think we can convert `pd.NaT` to ISO 8601. For that one I guess yeah we should just throw an error if it is missing.

nwb_conversion_tools/nwbconverter.py
```diff
@@ -60,25 +61,27 @@ def get_metadata(self):
...
+        conversion_data_routed = dict()
+        for interface_name, interface in self.data_interface_classes.items():
+            conversion_schema = interface.get_conversion_options_schema()
+            conversion_data_routed[interface_name] = get_schema_data(
+                conversion_options[interface_name],
+                conversion_schema
+            )
         for name, data_interface in self.data_interface_objects.items():
-            data_interface.convert_data(nwbfile, metadata_dict, stub_test)
+            data_interface.convert_data(nwbfile, metadata_dict, **conversion_data_routed[name])
```
bendichter, line 84: Can you combine these loops?

nwb_conversion_tools/utils.py
```diff
@@ -7,6 +7,10 @@
 import pynwb
 
 
+def get_schema_data(in_data, data_schema):
+    return {k: in_data.get(k, None) for k in data_schema['properties'].keys()}
```
bendichter, line 11:
```suggestion
    return {k: in_data[k] for k in data_schema['properties'] if k in in_data}
```

How about this, which doesn't auto-fill `None` for missing data

nwb_conversion_tools/nwbconverter.py
CodyCBakerPhD: Combined the occurrence of these loops here, as well as the similar structure in the input arguments

nwb_conversion_tools/utils.py
CodyCBakerPhD: Yeah, that's probably for the best. If there are missing input arguments or conversion options, if they are (a) optional, the downstream functions will use their own defaults, which aren't always `None`, or (b) required, best to not attempt to auto-population them anyway. Fixing now.

nwb_conversion_tools/nwbconverter.py
```diff
@@ -60,25 +54,27 @@ def get_metadata(self):
...
         """Build nwbfile object, auto-populate with minimal values if missing."""
         nwbfile_kwargs = dict(
-                session_description="no description",
-                identifier=str(uuid.uuid4()),
-                session_start_time=datetime.now()
-            )
-
+            session_description="Auto-generated description from NWBConverter.",
+            identifier=str(uuid.uuid4()),
+            session_start_time=datetime.now()
```
bendichter, line 62:
```suggestion
```

PR #77, opened 2020-11-10
define get_conversion_options_schem

bendichter:


CodyCBakerPhD:
@bendichter Thanks for the sub-branch; it really helps with these sort of technical changes.

One issue I have, even prior to these util changes is, the only arguments found by `get_schema_from_method_signature` on this are 'args' and 'kwargs'. We may need a sub-function that reads the next level from 'args'?

bendichter:
What do you mean?

CodyCBakerPhD:
I mean, when I call .get_conversion_options_schema() for the recording interface class, the only properties it finds are 'args' and 'kwargs'.

Also, should this be a class method like the .get_input_schema is?

CodyCBakerPhD:
Yeah, it definitely needs to be a class method to use it in the routing way we intend. I'll investigate the param nesting issue a bit more here

CodyCBakerPhD:
Ahhh, yeah no wonder. get_schema_from_method_signature is tailored for use on class initializers, not individual functions, see https://github.com/catalystneuro/nwb-conversion-tools/blob/auto_get_conversion_options_schema/nwb_conversion_tools/utils.py#L85

nwb_conversion_tools/basedatainterface.py
```diff
@@ -25,8 +25,8 @@ def get_metadata(self):
     @classmethod
     @abstractmethod
     def get_conversion_options_schema(cls):
-        return get_schema_from_method_signature(cls.convert_data, exclude=['nwbfile', 'metadata_dict'])
+        return get_schema_from_method_signature(cls.convert_data, exclude=['nwbfile', 'metadata_dict', 'stub_test'])
```
bendichter, line 28: why is stub_test removed?

nwb_conversion_tools/utils.py
```diff
@@ -82,7 +82,7 @@ def get_schema_from_method_signature(class_method, exclude=None):
     if exclude is None:
         exclude = []
     input_schema = get_base_schema()
-    for param in inspect.signature(class_method.__init__).parameters.values():
```
bendichter, line 83: ah, good call

nwb_conversion_tools/basedatainterface.py
CodyCBakerPhD: Trying to test at the moment if it can or can't be, but just about all interfaces expect it. It's mostly as you use this in an actual conversion framework (debugging with tank-lab atm), passing the conversion_options dictionary in the right way, you would have to repeat stub_test on a lot of the interfaces if it isn't handled globally.

CodyCBakerPhD:
Won't be able to 100% debug this until I implement the 0.4.0 version of SpikeGLX metadata, since we no longer have a way to force spikeextractors to autopopulate recording metadata (I used to just specify the recording metadata=None, but now that we pass the entire object it sees 'NWBFile', 'Subject' and hence thinks there ought to be an 'Ecephys', too). I'll do that after our meeting

CodyCBakerPhD:
But, the conversion option routing part largely appears to be working as intended now; under the hood, it is finding the options like 'sync_with_ttl' and others, at least.

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -9,16 +9,15 @@ class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
...
             rising_times = ttl[states == 1]
             self.recording_extractor = SubRecordingExtractor(self.recording_extractor, start_frame=rising_times[0])
         super().convert_data(nwbfile=nwbfile, metadata_dict=metadata_dict, stub_test=stub_test)
 
-    def get_metadata():
+    def get_metadata(self):
         """Retrieve Ecephys metadata specific to the SpikeGLX format."""
         # TODO
-        raise NotImplementedError
+        return dict()
```
bendichter, line 23: You are better off just deferring this function to the `BaseRecordingExtractorInterface`

CodyCBakerPhD (approved):

PR #81, opened 2020-11-11
get_metadata() for SpikeGLX

CodyCBakerPhD:
Though in principle, it appears as if the [SpikeGLX can have multiple shanks](https://github.com/billkarsh/SpikeGLX/blob/master/Markdown/UserManual.md#shank-map), the exact formatting of the channel mapping and extraction thereof would be unreliable without a good prototype file to test on.

The current implementation uses the default maps of single shanks.

### DIFF
--- a/nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
@@ -6,12 +6,12 @@ from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
 
 
 class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
-    """Primary data interface class for converting the SpikeGLX format."""
+    """Primary data interface class for converting the high-pass (ap) SpikeGLX format."""
 
     RX = SpikeGLXRecordingExtractor
 
     def get_metadata(self):
-        """Auto-populate as much metadata as possible from the SpikeGLX format."""
+        """Auto-populate as much metadata as possible from the high-pass (ap) SpikeGLX format."""
         file_path = Path(self.input_args['file_path'])
         session_id = file_path.parent.stem
 
@@ -27,7 +27,7 @@ class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
             Ecephys=dict(
                 Device=[
                     dict(
-                        description=session_id + '.ap.meta'
+                        description=f"More details for the high-pass (ap) data found in {session_id}.ap.meta!"
                     )
                 ],
                 ElectrodeGroup=[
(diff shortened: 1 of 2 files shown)


nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -1,15 +1,63 @@
...
+                    dict(
+                        name='group',
+                        description="A reference to the ElectrodeGroup this electrode is a part of.",
+                        data=shank_group_name
+                    ),
+                    dict(
+                        name='group_name',
+                        description="The name of the ElectrodeGroup this electrode is a part of.",
+                        data=shank_group_name
+                    )
```
bendichter, line 55: So `group` is a reference to ElectrodeGroup object, and `group_name` is the name of that object. I don't see a reason to pass both of these. You should be able to get the `group` value from the `group_name` value.

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -1,15 +1,63 @@
...
-        # TODO
-        raise NotImplementedError
+    def get_metadata(self):
+        """Auto-populate as much metadata as possible from the SpikeGLX format."""
+        file_path = Path(self.input_args['file_path'])
+        session_id = file_path.parent.stem
+
+        n_shanks = int(self.recording_extractor._meta['snsShankMap'][1])
+        if n_shanks > 1:
+            raise NotImplementedError("SpikeGLX metadata for more than a single shank is not yet supported.")
```
bendichter, line 20: Nice error catching

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
CodyCBakerPhD: Good catch, this may be a relic from prior to the add_electrode refactor, when passing both was necessary to create the link in the dynamic table. In theory just group_name should work currently, but I'll test it

CodyCBakerPhD: Right you are, this is from back when the process of adding electrode metadata required one key for each column of non-default information, so group and group_name had to exist separately and correspond one-to-one.

Since the refactor, though, only group_name is required and auto-populates the group link based only on that.

Removed this from both spikeglx and neuroscope.

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -1,15 +1,58 @@
...
+                    ),
+                    dict(
+                        name='group_name',
+                        description="The name of the ElectrodeGroup this electrode is a part of.",
+                        data=shank_group_name
+                    )
+                ],
+                ElectricalSeries=dict(
+                    name='ElectricalSeries',
+                    description="Raw acquisition traces."
```
bendichter, line 54: I believe spikeGLX records two different datasets, ap, which is high-passed, and lf, which is low-passed. Could you reflect this in the metadata structure with descriptions?

CodyCBakerPhD: Updated the Series and device (since there are two different meta files) descriptions as well as docstrings; I'm assuming the lf would use the same electrodes, so haven't modified those.

PR #82, opened 2020-11-12
simplified pickle interface to rely on base recording/sorting extractors

CodyCBakerPhD:
@alejoe91 Significant reduction in duplicated code for running the interface. I agree that long-term we should look into mix-ins for separating the internal class configuration (to adhere better to inheritance) from the common set of conversion functionality, but that will likely take a bit of debugging to streamline.

### DIFF
--- a/nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
+++ b/nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
@@ -1,18 +1,16 @@
 """Authors: Alessio Buccino."""
 import spikeextractors as se
-from pynwb import NWBFile
-from pynwb.device import Device
-from pynwb.ecephys import ElectrodeGroup, ElectricalSeries, SpikeEventSeries
 
-from ..basedatainterface import BaseDataInterface
-from ..utils import get_base_schema, get_schema_from_method_signature, \
-    get_schema_from_hdmf_class
+from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
+from ..basesortingextractorinterface import BaseSortingExtractorInterface
 
 
-class SIPickleRecordingExtractorInterface(BaseDataInterface):
+class SIPickleRecordingExtractorInterface(BaseRecordingExtractorInterface):
+    """Primary interface for reading and converting SpikeInterface objects through Pickle files."""
 
     @classmethod
     def get_input_schema(cls):
+        """Return partial json schema for expected input arguments."""
         return dict(
             required=['pkl_file'],
             properties=dict(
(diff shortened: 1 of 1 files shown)


nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
```diff
@@ -21,51 +19,16 @@ def get_input_schema(cls):
         )
 
     def __init__(self, **input_args):
-        super().__init__(**input_args)
+        self.input_args = input_args
```
bendichter, line 22: why is this line needed?

nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
```diff
@@ -74,45 +37,5 @@ def get_input_schema(cls):
...
```
bendichter, line 40: why is this line needed?

nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
CodyCBakerPhD: Previously, when this inherited from BaseDataInterface, calling super().__init__ calls self.input_args = input_args.

The other alternative here is to call super().super().__init__ but that seems a little convoluted (let me know if you'd prefer it, though).

Having the record of received input arguments is very useful for both debugging and retrieving information for stuff like metadata.

bendichter: why do we need input_args to be assigned as an attribute of the class?

nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
```diff
@@ -1,18 +1,18 @@
...
-    get_schema_from_hdmf_class
+from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
+from ..basesortingextractorinterface import BaseSortingExtractorInterface
 
 
-class SIPickleRecordingExtractorInterface(BaseDataInterface):
+class SIPickleRecordingExtractorInterface(BaseRecordingExtractorInterface):
+    """Primary interface for reading and converting SpikeInterface objects through Pickle files."""
+
+    RX = None
```
bendichter, line 11: Why is this needed?

CodyCBakerPhD: With this version that inherits from the BaseRecordingExtractorInterface, I assume it would only be proper to contain the RX class object common to all RecordingExtractorInterfaces, with the None value indicating it is not by default any particular class from SpikeExtractor.

That said, since we're not relying on the class for instantiation of the actual instance of the extractor in the __init__, I suppose it wouldn't actually get used unless something downstream wanted to test whether this interface has an RX object that is None or something.

We could also update the status of RX perhaps after loading the recording object during the __init__ here, I think.

nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
CodyCBakerPhD: All inheritors of the [base data interfaces](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/basedatainterface.py) have this. Downstream stuff may or may not wish to access it as a shortcut for retrieving file paths and stuff of that nature

bendichter: ok sure

nwb_conversion_tools/datainterfaces/sipickledatainterfaces.py
bendichter: This class would already inherit the `RX = None` from the `BaseRecordingExtractorInterface` so this line is redundant

CodyCBakerPhD: Oh, that's true. Fixing that now

PR #83, opened 2020-11-12
Add save si object function

CodyCBakerPhD:
@alejoe91 @bendichter This is the type of initial structure I was thinking to reduce the burden on the primary pipeline.

Basically, the client would do all the processing they want to, and in that ipython environment they would have at least one recording and/or sorting extractor they would want to convert to nwb. They do this first by saving the intermediate file via the si_object_location = save_si_object(this_object, this_folder), and we then set up the custom converter class for that pipeline to expect the cached objects at the location returned (passed as an input argument along with the behavioral data folder/file locations, etc).

The conversion then runs as if all the processed data (i.e., sorting and synchronized recording) had been saved in some particular format (but we now control that format ourselves via SpikeInterface).


CodyCBakerPhD:
There's several ways this can be improved though which is why I'm just opening this as a draft; there's also opportunity for metaprogramming since all the class names are built-in and ought to be automatable in some way.

CodyCBakerPhD:
And I left the other plethora of caching types in comments to indicate the scope this can cover, but only focusing on recording and sorting for now.

nwb_conversion_tools/conversion_tools.py
```diff
@@ -0,0 +1,57 @@
...
+    elif isinstance(si_object, se.SortingExtractor):  # TODO: asssumes a single extractor at the moment (not  iterable)
+        if not si_object.is_dumpable or cache_raw:
+            cache = se.CacheSortingExtractor(si_object, spikeinterface_folder / "sorting.npz")
+        else:
+            cache = si_object
+    else:
+        raise ValueError("The 'si_object' argument shoulde be a SpikeInterface Extractor!")
+
+    json_file = spikeinterface_folder / "raw.json"
+    pkl_file = spikeinterface_folder / "raw.pkl"
```
CodyCBakerPhD, line 47: @alejoe91 Just to clarify, if we dump both a recording and sorting object to these files using this method, the single .json and .pkl files will contain all the information for both extractor objects? (since the name of these files is hard-coded and not different per object)

CodyCBakerPhD: Or does the cache.dump_to_* otherwise automatically make the file name unique per cache?

alejoe91: I think we should dump the json (we should move it before the caching option) because that keeps all the information, while after caching it's resetted to the new bin file.

bendichter:
@CodyCBakerPhD is this ready to merge?

CodyCBakerPhD:
@bendichter hasn't been integrated into the pipeline just yet, I'll do that in the next hour

CodyCBakerPhD:
@bendichter @alejoe91 OK this should be good to go. I changed a couple things with the `save_si_object function`;

1. The first argument is now a string defining the name the pickle file is saved under, in order to make it unique across multiple calls to the save function.
2. It no longer returns anything, just takes the action to save it to the location in the `spikeinterface_folder`.

CodyCBakerPhD:
@bendichter @alejoe91 No merge conflicts here, should be good to go

nwb_conversion_tools/baserecordingextractorinterface.py
```diff
@@ -42,20 +42,16 @@ def get_metadata_schema(self):
...
-                ElectricalSeries=dict(
-                    name='ElectricalSeries',
-                    description='raw acquired data'
-                ),
-                Device=dict(
-                    name='device',
-                    description='ecephys probe'
-                )
+                ElectricalSeries=dict(),
+                Device=[],
```
bendichter, line 49: why remove these? They seem like good defaults for a base recording extractor interface

CodyCBakerPhD: Because `NwbRecordingExtractor.write_recording()` is setup with similar defaults, but more robustly in the sense of interactions across `Electrodes` -> `ElectrodeGroups` -> `Device`. But  `NwbRecordingExtractor.write_recording()` can't fill with these defaults correctly unless it receives either `metadata=None`, or a metadata with each of these fields as empty.

bendichter: I see. I think we can handle that with options on deep_dict_update, but let's worry about that later.

CodyCBakerPhD: Agreed; there is probably a better way to set all these types of defaults but it will be easier to see and fix when all the other pieces currently in development are in place.

PR #87, opened 2020-11-16
Sorting interface to write electrode metadata without recording

CodyCBakerPhD:
@bendichter Extended sorting interface to include capability of writing electrode metadata in the absence of a true recording extractor. This removes the need for a NoRecording interface on the conversions that currently make use of it.

I did begin this approach by attempting to refactor the NwbRecordingExtractor.add_* methods into a general, non-class specific form so that they could be used in both that and the NwbSortingExtractor, and it worked well up until add_electrodes, which is intertwined with its recording extractor object at nearly every update to the keyword arguments for nwbfile.add_electrode, so was unable to generalize it on that repo.

I believe this current interface specific approach is far simpler, however, and even has the same capabilities for filling missing data as would be had by a true RecordingExtractor.

### DIFF
--- a/nwb_conversion_tools/basesortingextractorinterface.py
+++ b/nwb_conversion_tools/basesortingextractorinterface.py
@@ -1,5 +1,6 @@
 """Authors: Cody Baker and Ben Dichter."""
 import spikeextractors as se
+import numpy as np
 from pynwb import NWBFile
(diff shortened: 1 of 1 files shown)


bendichter:
This looks pretty good. there might be an even better way to do this with mixins, but lets go with this for now.

PR #91, opened 2020-11-18
New schema compliance

CodyCBakerPhD:
@bendichter @luiztauffer Results of our meeting, minus the json validation (on the TODO list) and lazy imports. Probably best to do those on a different branch, I'll get started on them tomorrow.

Mostly name change standardization, a couple comments below on anything other than that.

### DIFF
--- a/nwb_conversion_tools/basedatainterface.py
+++ b/nwb_conversion_tools/basedatainterface.py
@@ -1,5 +1,6 @@
 """Authors: Cody Baker and Ben Dichter."""
 from abc import abstractmethod
+import warnings
 
 from .utils import get_base_input_schema, get_metadata_schema, get_schema_from_method_signature
 
@@ -26,3 +27,9 @@ class BaseDataInterface:
     @abstractmethod
     def run_conversion(self, nwbfile_path, metadata):
         pass
+
+    @abstractmethod
+    def convert_data(self, nwbfile_path, metadata):
+        warnings.warn("The convert_data method should now be renamed to run_conversion "
+                      "as of nwb-conversion-tools v0.6.0", DeprecationWarning)
+        self.run_conversion(nwbfile_path, metadata)
--- a/nwb_conversion_tools/basedatainterface.py
+++ b/nwb_conversion_tools/basedatainterface.py
@@ -21,8 +21,8 @@ class BaseDataInterface:
 
     @classmethod
     def get_conversion_options_schema(cls):
-        return get_schema_from_method_signature(cls.convert_data, exclude=['nwbfile', 'metadata_dict'])
+        return get_schema_from_method_signature(cls.convert_data, exclude=['nwbfile', 'metadata'])
 
     @abstractmethod
-    def convert_data(self, nwbfile_path, metadata_dict):
+    def run_conversion(self, nwbfile_path, metadata):
         pass
--- a/nwb_conversion_tools/basedatainterface.py
+++ b/nwb_conversion_tools/basedatainterface.py
@@ -25,11 +25,11 @@ class BaseDataInterface:
         return get_schema_from_method_signature(cls.convert_data, exclude=['nwbfile', 'metadata'])
 
     @abstractmethod
-    def run_conversion(self, nwbfile_path, metadata):
+    def run_conversion(self, nwbfile_path, metadata, **conversion_options):
         pass
 
     @abstractmethod
-    def convert_data(self, nwbfile_path, metadata):
+    def convert_data(self, nwbfile_path, metadata, **conversion_options):
         warnings.warn("The convert_data method should now be renamed to run_conversion "
                       "as of nwb-conversion-tools v0.6.0", DeprecationWarning)
-        self.run_conversion(nwbfile_path, metadata)
+        self.run_conversion(nwbfile_path, metadata, **conversion_options)
--- a/nwb_conversion_tools/basedatainterface.py
+++ b/nwb_conversion_tools/basedatainterface.py
@@ -2,27 +2,27 @@
 from abc import abstractmethod
 import warnings
 
-from .utils import get_base_source_schema, get_base_metadata_schema, get_schema_from_method_signature
+from .json_schema_utils import get_base_schema, get_schema_from_method_signature
 
 
 class BaseDataInterface:
 
     @classmethod
     def get_source_schema(cls):
-        return get_base_source_schema()
+        return get_base_schema()
 
     def __init__(self, **source_data):
         self.source_data = source_data
 
     def get_metadata_schema(self):
-        return get_base_metadata_schema()
+        return get_base_schema()
 
     def get_metadata(self):
         return dict()
 
     @classmethod
     def get_conversion_options_schema(cls):
-        return get_schema_from_method_signature(cls.convert_data, exclude=['nwbfile', 'metadata'])
+        return get_schema_from_method_signature(cls.run_conversion, exclude=['nwbfile', 'metadata'])
 
     @abstractmethod
     def run_conversion(self, nwbfile_path, metadata, **conversion_options):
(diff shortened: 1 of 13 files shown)


nwb_conversion_tools/basedatainterface.py
```diff
@@ -21,8 +22,14 @@ def get_metadata(self):
...
 
     @abstractmethod
-    def convert_data(self, nwbfile_path, metadata_dict):
+    def run_conversion(self, nwbfile_path, metadata):
         pass
+
+    @abstractmethod
+    def convert_data(self, nwbfile_path, metadata):
+        warnings.warn("The convert_data method should now be renamed to run_conversion "
+                      "as of nwb-conversion-tools v0.6.0", DeprecationWarning)
```
CodyCBakerPhD, line 34: @bendichter Should this be a warning or a hard assertion on the deprication?

nwb_conversion_tools/utils.py
```diff
@@ -103,21 +103,27 @@ def get_schema_from_hdmf_class(hdmf_class):
...
 
         # type string
-        elif docval_arg['type'] is str or (isinstance(docval_arg['type'], tuple) and str in docval_arg['type']):
+        elif docval_arg['type'] is str \
+            or (isinstance(docval_arg['type'], tuple)
+                and str in docval_arg['type']):
             schema_arg[docval_arg['name']].update(type='string')
 
         # type datetime
-        elif docval_arg['type'] is datetime or (isinstance(docval_arg['type'], tuple) and datetime in docval_arg['type']):
```
CodyCBakerPhD, line 114: This line was over 120 characters long, which is the standard over at `spikeextractors` and also what I use, so shortened it and aligned the nearby evaluations in a similar style.

nwb_conversion_tools/nwbconverter.py
```diff
@@ -59,28 +63,28 @@ def get_metadata(self):
...
-                metadata_dict['NWBFile']['session_start_time']
+                metadata['NWBFile']['session_start_time']
             )
         nwbfile = NWBFile(**nwbfile_kwargs)
 
         for interface_name, data_interface in self.data_interface_objects.items():
-            these_conversion_options = get_schema_data(
-                conversion_options,
+            valid_conversion_options = get_schema_data(
+                conversion_options[interface_name],
```
CodyCBakerPhD, line 84: This is the only line that actually changes the functionality (and as a consequence bumps this up to v0.6.0). All downstream dependents will not have to pass conversion options on a per interface basis.

luiztauffer: @CodyCBakerPhD can you check the changes I proposed on 96fed83 and 1c7eec5? I think we won't need the `get_schema_data` step, since we should validate the data against schema for source, conversion_options and metadata

nwb_conversion_tools/baserecordingextractorinterface.py
```diff
@@ -75,14 +74,16 @@ def convert_data(self, nwbfile, metadata_dict: None, stub_test=False):
         else:
             stub_recording_extractor = self.recording_extractor
 
-        if metadata_dict is not None and 'Ecephys' in metadata_dict and 'subset_channels' in metadata_dict['Ecephys']:
-            recording_extractor = se.SubRecordingExtractor(stub_recording_extractor,
-                                                           channel_ids=metadata_dict['Ecephys']['subset_channels'])
+        if metadata is not None and 'Ecephys' in metadata and 'subset_channels' in metadata['Ecephys']:
+            recording_extractor = se.SubRecordingExtractor(
+                stub_recording_extractor,
+                channel_ids=metadata['Ecephys']['subset_channels']
```
bendichter, line 80: I don't really like `subset_channels` being in the metadata dictionary. Can this be set as an attribute in the `__init__`? (set to `None` by default). This would allow it to be easily changed programmatically if the user needs to.

nwb_conversion_tools/nwbconverter.py
```diff
@@ -17,42 +19,33 @@ class NWBConverter:
...
-        It automatically checks with the interface schemas which data belongs to each
+        Validate source_data against source_schema and initialize all data interfaces.
         """
-        self.data_interface_objects = {name: data_interface(**input_data[name])
-                                       for name, data_interface in
-                                       self.data_interface_classes.items()}
+        # Validate source_data against source_schema
+        validate(instance=source_data, schema=self.get_source_schema())
+
+        # If data is valid, procede to instantiate DataInterface objects
```
bendichter, line 41:
```suggestion
        # If data is valid, proceed to instantiate DataInterface objects
```

nwb_conversion_tools/basedatainterface.py
bendichter: This is good

nwb_conversion_tools/nwbconverter.py
```diff
@@ -17,42 +19,33 @@ class NWBConverter:
...
-        self.data_interface_objects = {name: data_interface(**input_data[name])
-                                       for name, data_interface in
-                                       self.data_interface_classes.items()}
+        # Validate source_data against source_schema
+        validate(instance=source_data, schema=self.get_source_schema())
+
+        # If data is valid, proceed to instantiate DataInterface objects
+        self.data_interface_objects = dict()
+        for interface_name, data_interface in self.data_interface_classes.items():
+            self.data_interface_objects.update({interface_name: data_interface(**source_data[interface_name])})
```
bendichter, line 44:
```suggestion
        self.data_interface_objects = {name: data_interface(**source_data[name])
                                       for name, data_interface in
                                       self.data_interface_classes.items()}
```

nwb_conversion_tools/datainterfaces/neuroscopedatainterface.py
```diff
@@ -14,6 +14,27 @@ class NeuroscopeRecordingInterface(BaseRecordingExtractorInterface):
 
     RX = se.NeuroscopeRecordingExtractor
 
+    def __init__(self, **input_args):
+        super().__init__(**input_args)
```
bendichter, line 18:
```suggestion
    def __init__(self, *args, **kwargsargs):
        super().__init__(*args, **kwargs)
```

nwb_conversion_tools/datainterfaces/neuroscopedatainterface.py
```diff
@@ -14,6 +14,27 @@ class NeuroscopeRecordingInterface(BaseRecordingExtractorInterface):
 
     RX = se.NeuroscopeRecordingExtractor
 
+    def __init__(self, *args, **kwargsargs):
```
bendichter, line 17:
```suggestion
    def __init__(self, *args, **kwargs):
```

nwb_conversion_tools/nwbconverter.py
```diff
@@ -62,21 +80,23 @@ def get_metadata(self):
...
+    def run_conversion(self, metadata: dict, nwbfile_path=None, save_to_file: bool = True,
+                       conversion_options: dict = None):
         """Build nwbfile object, auto-populate with minimal values if missing."""
         if conversion_options is None:
             conversion_options = dict()
-        nwbfile_kwargs = dict()
-        if 'NWBFile' in metadata:
-            nwbfile_kwargs.update(metadata['NWBFile'])
+        else:
+            validate(instance=conversion_options, schema=self.get_conversion_options_schema())
```
luiztauffer: This would be problematic in this specific situation: The conversion_options_schema has required properties, but when calling run_conversion() the users does not pass them

bendichter: I don't think any conversion options should be required

nwb_conversion_tools/nwbconverter.py
CodyCBakerPhD: Removed

CodyCBakerPhD:
@bendichter @luiztauffer All changes look good to me; should probably merge this and do any further changed on a different branch given this has reached critical levels of +/- 300 changes, making a holistic review quite difficult.

PR #94, opened 2020-11-18
debugging for nwb-conversion-tools v0.6.0

CodyCBakerPhD:
@bendichter @luiztauffer The new json schema validation, while very nice and general, produced a number of errors that I've begun to correct here and in the Tank repo first.

The biggest changes are related to the now very heavy usage of `get_schema_from_method_signature` coupled  with the fact that not all downstream `properties` are of type `string`, as was previously hard-coded. Extended this to at least include type `bool` and throw an error if not either of those types to let us know to implement it.

While the actual `param_type` checking is the biggest direct change, there are a number of indirect changes required across the repo as well as coding practices for downstream inheritances. This is due to the usage derived from `param.annotation` which requires much better specification of value types across downstream functions. If this is undesired, I suppose we could just change the error I mentioned to a soft warning and simply default to a `string` type as before, but who knows what issues this might cause in more complicated future situations.

Otherwise, added some other docstrings and styling corrections.

### DIFF
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -1,3 +1,4 @@
+"""Authors: Luiz Tauffer, Cody Baker, and Ben Dichter."""
 import collections.abc
 import inspect
 
@@ -18,6 +19,7 @@ def dict_deep_update(d, u):
 
 
 def get_base_schema(tag=None, root=False, id_=None, **kwargs):
+    """Return the base schema used for all other schemas."""
     base_schema = dict(
         required=[],
         properties={},
@@ -26,23 +28,19 @@ def get_base_schema(tag=None, root=False, id_=None, **kwargs):
     )
     if tag is not None:
         base_schema.update(tag=tag)
-
     if root:
         base_schema.update({
             "$schema": "http://json-schema.org/draft-07/schema#"
         })
-
     if id_:
         base_schema.update({"$id": id_})
-
     base_schema.update(**kwargs)
-
     return base_schema
 
 
 def get_schema_from_method_signature(class_method, exclude=None):
     """
-    Take a class method and return a jsonschema of the input args
+    Take a class method and return a jsonschema of the input args.
 
     Parameters
     ----------
@@ -59,9 +57,17 @@ def get_schema_from_method_signature(class_method, exclude=None):
     input_schema = get_base_schema()
     for param in inspect.signature(class_method).parameters.values():
         if param.name not in exclude + ['self']:
+            if param.name != 'self' and param.annotation:
+                if param.annotation is bool:
+                    param_type = "boolean"
+                if param.annotation is str:
+                    param_type = "string"
+            else:
+                raise NotImplementedError(f"The annotation type of '{param}' in function '{class_method}' "
+                                          "is not assigned! Please implement.")
             arg_spec = {
                 param.name: dict(
-                    type='string'
+                    type=param_type
                 )
             }
             if param.default is param.empty:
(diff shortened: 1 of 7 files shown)


nwb_conversion_tools/json_schema_utils.py
```diff
@@ -59,9 +57,17 @@ def get_schema_from_method_signature(class_method, exclude=None):
...
     for param in inspect.signature(class_method).parameters.values():
         if param.name not in exclude + ['self']:
+            if param.name != 'self' and param.annotation:
+                if param.annotation is bool:
+                    param_type = "boolean"
+                if param.annotation is str:
+                    param_type = "string"
+            else:
+                raise NotImplementedError(f"The annotation type of '{param}' in function '{class_method}' "
+                                          "is not assigned! Please implement.")
```
bendichter, line 67: Thanks for fixing this! Would you mind refactoring this into a dictionary? Something like

```python
annotation_json_type_map = dict(
    bool="boolean",
    str="string",
    int="number",
    float="number"
)

File: nwb_conversion_tools/nwbconverter.py [hunk 2]
```diff
@@ -110,6 +111,8 @@ def run_conversion(self, metadata: dict, nwbfile_path=None, save_to_file: bool =
 
         # If conversion_options is valid, procede to run conversions for DataInterfaces
         for interface_name, data_interface in self.data_interface_objects.items():
+            if interface_name not in conversion_options:
+                conversion_options.update({interface_name: dict()})
             data_interface.run_conversion(nwbfile, metadata, **conversion_options[interface_name])
```
bendichter, line 116:
```suggestion
            data_interface.run_conversion(nwbfile, metadata, **conversion_options.get(interface_name, dict()))
```

bendichter:
@CodyCBakerPhD looking really good! Thanks for working your way through all of this! Two stylistic suggestions, then I think it's ready to merge

nwb_conversion_tools/json_schema_utils.py
CodyCBakerPhD: Yeah, I was thinking something like that might be the best way to go... Are those all the other 'simple' `json` schema types? Dictionaries would also be good to include, I'll check if that's supported

nwb_conversion_tools/nwbconverter.py
CodyCBakerPhD: Good catch; as I wrote that my tired brain said there was probably a nicer way than what I was doing, but couldn't remember what it was

nwb_conversion_tools/json_schema_utils.py
CodyCBakerPhD: [http://json-schema.org/understanding-json-schema/reference/type.html](http://json-schema.org/understanding-json-schema/reference/type.html) looks like it ought to

CodyCBakerPhD:
@bendichter OK, suggestions implemented. Reconciliation with the `json` basic types is maxed out and generalized to just extract it direction from the annotation string.

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -57,11 +55,25 @@ def get_schema_from_method_signature(class_method, exclude=None):
...
+        str="string",
+        int="number",
+        float="number",
+        dict="object",
+        list="array"
+    )
+
     for param in inspect.signature(class_method).parameters.values():
         if param.name not in exclude + ['self']:
+            if param.name != 'self' and param.annotation:
```
bendichter, line 69:
```suggestion
            if param.annotation:
```
This is handled by the line above

CodyCBakerPhD: Ah, yes. I had it outside that statement then copied it in but forgot to remove it after. Thanks!

PR #97, opened 2020-11-23
unroot schema

bendichter:
@luiztauffer when a `DataInterface.get_metadata_schema` returns values like `$shema` and `id_`, these values should only be on the root level of a json-schema object. When it gets merged with other schemas, it should lose these identifiers. Would this PR work for you?


luiztauffer (approved):
looks good to me

PR #99, opened 2020-11-26
add cell explorer sorting interface

CodyCBakerPhD:
@bendichter Pretty self-explanatory; built to be as redundant as possible with regard to the existence of optional fields in the matlab files, as well as the existence of the files themselves.

depends on https://github.com/SpikeInterface/spikeextractors/pull/509


bendichter:
@CodyCBakerPhD The .mat files are not part of [the neuroscope data standard](http://neuroscope.sourceforge.net/UserManual/data-files.html). This one is part of [the cellexplorer/buzcode standard]((https://cellexplorer.org/datastructure/data-structure-and-format/#spikes)). I'd like to keep things named "neuroscope" about purely data types that are in that standard. This seems like a useful interface, as it would allow us to be more interoperable with the cellexplorer software, but it needs to be named and organized accordingly.

CodyCBakerPhD:
@bendichter Fair point; I'll make a separate CellExplorerSortingInterface for that then.

CodyCBakerPhD:
Interface now organized according to format type.

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
```diff
@@ -0,0 +1,66 @@
...
+                    )
+                )
+            if 'region' in cell_info_fields:
+                unit_properties.append(
+                    dict(
+                        name="region",
+                        description="Brain region where each unit was detected.",
+                        data=[str(x[0]) for x in cell_info['region'][0][0][0]]
+                    )
+                )
```
bendichter, line 51: This should  be the required field "location" in NWB

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
```diff
@@ -0,0 +1,66 @@
...
+        celltype_mapping = {'pE': "excitatory", 'pI': "inhibitory"}
+        celltype_filepath = session_path / f"{session_id}.CellClass.cellinfo.mat"
+        if celltype_filepath.is_file():
+            celltype_info = loadmat(celltype_filepath).get('CellClass', np.empty(0))
+            if 'label' in celltype_info.dtype.names:
+                unit_properties.append(
+                    dict(
+                        name="cell_type",
+                        description="Type of cell this has been classified as.",
+                        data=[str(celltype_mapping[x[0]]) for x in celltype_info['label'][0][0][0]]
```
bendichter, line 62: The `str()` wrapper seems unnecessary here

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
```diff
@@ -0,0 +1,66 @@
...
+                        name="region",
+                        description="Brain region where each unit was detected.",
+                        data=[str(x[0]) for x in cell_info['region'][0][0][0]]
+                    )
+                )
+
+        celltype_mapping = {'pE': "excitatory", 'pI': "inhibitory"}
+        celltype_filepath = session_path / f"{session_id}.CellClass.cellinfo.mat"
+        if celltype_filepath.is_file():
+            celltype_info = loadmat(celltype_filepath).get('CellClass', np.empty(0))
```
bendichter, line 56: What's the plan here? Is 'CellClass` ever missing? If so, what happens?

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
```diff
@@ -0,0 +1,66 @@
...
+            if 'label' in celltype_info.dtype.names:
+                unit_properties.append(
+                    dict(
+                        name="cell_type",
+                        description="Type of cell this has been classified as.",
+                        data=[str(celltype_mapping[x[0]]) for x in celltype_info['label'][0][0][0]]
+                    )
+                )
+
+        return dict(UnitProperties=unit_properties)
```
bendichter, line 66: Wouldn't this be

```python
return dict(
    Ecephys=dict(
        UnitProperties=unit_properties
    )
```
?

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
CodyCBakerPhD: `CellClass.cellinfo` is a separate file from the main `spikes.cellinfo`, and it could be conceivable that it is missing. All it contains is the actual assigned labels (pE, pI) that were classified to each unit.

If it's missing, you simply don't include cell types column to the units table.

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
CodyCBakerPhD: The `BaseSortingInterface` is currently setup to pull metadata from a first-level field `UnitProperties`.

This doesn't appear to have been added to metadata schema documentation so I'll look into that; however, these are also all completely custom/optional so it's not a validation/enforcement thing.

We can discuss it further, but I think it's conceivable that there is spiking activity inferred or included outside the context of `Ecephys`, so I would personally want to keep it at that first level.

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
CodyCBakerPhD: `location` is only required for electrodes though, I thought? This is basically that same information that could be obtained by mapping the electrode locations through the `group_id` this unit was detected by.

I'll definitely change the name to reflect this, though.

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
CodyCBakerPhD: `loadmat` reads it in as a `numpy.str_` type, not `str`, so I'll have to double check if that messes anything up during the conversion.

bendichter: Ah ok

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
bendichter: oh, yeah, you're right that field isn't specifically defined here, but it is in several NWB objects and I think we should be consistent.

nwb_conversion_tools/datainterfaces/cellexplorerdatainterface.py
bendichter: Would it be possible that a SortingInterface would add metadata to other fields of the metadata dict? e.g. session or subject info?

CodyCBakerPhD: I can't think of any use case like that from everything I've seen, except the open question of how best to allow non-recording interfaces to write/call the writing of `Ecephys` info into the metadata dictionary in the case where there is no actual `RecordingInterface` (some of those changes are proposed [on this separate PR](https://github.com/catalystneuro/nwb-conversion-tools/pull/102)) but it might be better to open this as a discussion on Slack to see if other people ever have odd use cases like that. I'll get that convo started in a bit

bendichter: I think that's a great use-case. Also 

* The `sessionName.spikes.cellinfo.mat` file has a `sessionName` attribute that could be used as an `NWBFile.session_id`
* The Vermin data contains subject id

I'm sure there are more. I don't really see a downside in passing back an entire metadata dict.

CodyCBakerPhD:
TODO: Enhance metadata to fill in as much info as possible, including in non-'UnitProperties' fields, and allow the central deep_dict update to merge these into the central converters metadata.

CodyCBakerPhD:
@bendichter CellExplorerSorting is now merged on spikeextractors, so this should be ready for final review.

It is now filling in any non-UnitProperty metadata that happens to be in the same files being accessed by the interface.

PR #100, opened 2020-11-26
Extend annotation checking to union

CodyCBakerPhD:
@bendichter Some of the annotation types in `spikeextractors` are fancier in their extensions using `typing.Union` to include things like `Path`, `None`, etc., in addition to something obvious like a `str`. This threw off the original annotation type checking, of course, so this fixes it in a _fairly_ general manner by finding the common base data type and validating against that (assuming there is one and only one match with the base data type).

One interesting case that would break this, which we _may_ run into in the future with the method attached here is an `ArrayLike` annotation, which can include more than one of the allowable json types (`int`, `float`, `list`) but could still be valid in that mapping as a json `array`. Depends on how the A`rrayLike` call is implemented as well, which also depends on the version of python in question.

But anyway, this works for now.

### DIFF
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -1,6 +1,7 @@
 """Authors: Luiz Tauffer, Cody Baker, and Ben Dichter."""
 import collections.abc
 import inspect
+import numpy as np
 
 
 def dict_deep_update(d, u):
@@ -69,12 +70,16 @@ def get_schema_from_method_signature(class_method, exclude=None):
     for param_name, param in inspect.signature(class_method).parameters.items():
         if param_name not in exclude:
             if param.annotation:
-                param_type = [
-                    annotation_json_type_map[arg_type] for arg_type in param.annotation.__args__
-                    if arg_type in annotation_json_type_map
-                ]
-                assert len(param_type) == 1, \
-                    f"There must be only one valid annotation type that maps to json! {len(param_type)} found."
+                valid_args = [x in annotation_json_type_map for x in param.annotation.__args__]
+                if any(valid_args):
+                    param_type = [annotation_json_type_map[x] for x in np.array(param.annotation.__args__)[valid_args]]
+                else:
+                    raise ValueError("There must be only one valid annotation type that maps to json! "
+                                     f"{param.annotation.__args__} found.")
+
+                if len(set(param_type)) > 1:
+                    raise ValueError("Conflicting json parameter types were detected from the annotation! "
+                                     f"{param.annotation.__args__} found.")
             else:
                 raise NotImplementedError(f"The annotation type of '{param}' in function '{class_method}' "
                                           "is not implemented! Please request it to be added at github.com/"
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -71,8 +71,8 @@ def get_schema_from_method_signature(class_method, exclude=None):
         if param_name not in exclude:
             if param.annotation:
                 param_type = [
-                    annotation_json_type_map[y] for x in param.annotation.__args__
-                    for y in [str(x).split("'")[1]] if y in annotation_json_type_map
+                    annotation_json_type_map[anno] for arg_types in param.annotation.__args__
+                    for anno in [str(arg_types).split("'")[1]] if anno in annotation_json_type_map
                 ]
                 assert len(param_type) == 1, \
                     f"There must be only one valid annotation type that maps to json! {len(param_type)} found."
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -83,7 +83,8 @@ def get_schema_from_method_signature(class_method, exclude=None):
             else:
                 raise NotImplementedError(f"The annotation type of '{param}' in function '{class_method}' "
                                           "is not implemented! Please request it to be added at github.com/"
-                                          "catalystneuro/nwb-conversion-tools/issues.")
+                                          "catalystneuro/nwb-conversion-tools/issues or create the json-schema"
+                                          "for this method manually.")
             arg_spec = {
                 param_name: dict(
                     type=param_type[0]
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -79,7 +79,7 @@ def get_schema_from_method_signature(class_method, exclude=None):
                 elif "'" in anno:
                     param_type = annotation_json_type_map[anno.split("'")[1]]
                 else:
-                    raise NotImplementedError("The annotation format of '{param}' in function '{class_method}' "
+                    raise NotImplementedError(f"The annotation format of '{param}' in function '{class_method}' "
                                               "is unrecognized!")
             else:
                 raise NotImplementedError(f"The annotation type of '{param}' in function '{class_method}' "
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -79,12 +79,13 @@ def get_schema_from_method_signature(class_method, exclude=None):
                 elif "'" in anno:
                     param_type = annotation_json_type_map[anno.split("'")[1]]
                 else:
-                    raise NotImplementedError(f"The annotation type of '{param}' in function '{class_method}' "
-                                              "is not implemented! Please request it to be added at github.com/"
-                                              "catalystneuro/nwb-conversion-tools/issues.")
+                    raise NotImplementedError("The annotation format of '{param}' in function '{class_method}' "
+                                              "is unrecognized!")
             else:
                 raise NotImplementedError(f"The annotation type of '{param}' in function '{class_method}' "
-                                          "is not assigned! Please implement.")
+                                          "is not implemented! Please request it to be added at github.com/"
+                                          "catalystneuro/nwb-conversion-tools/issues.")
+
             arg_spec = {
                 param.name: dict(
                     type=param_type
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -58,21 +58,21 @@ def get_schema_from_method_signature(class_method, exclude=None):
     else:
         exclude = exclude + ['self']
     input_schema = get_base_schema()
-    annotation_json_type_map = dict(
-        bool="boolean",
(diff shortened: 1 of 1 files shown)


bendichter:
Looks good. We can always just return a custom json-schema instead of trying to automatically generate them if we need

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -67,10 +68,24 @@ def get_schema_from_method_signature(class_method, exclude=None):
     for param in inspect.signature(class_method).parameters.values():
         if param.name not in exclude + ['self']:
             if param.annotation:
-                param_type = annotation_json_type_map[str(param.annotation).split("'")[1]]
+                anno = str(param.annotation)
+                if "typing.Union" in anno:
+                    types = re.search("typing.Union\[(.*)\]", anno).group(1).split(",")
+                    intersect_valid_keys = list(set(annotation_json_type_map.keys()).intersection(types))
```
bendichter, line 74: It would be better to use `isinstance(param.annotation, typing._GenericAlias)` and `param.annotation.__args__` rather than parsing the string

bendichter, line 74:
```suggestion
                anno = str(param.annotation)
                if "typing.Union" in anno:
                    types = re.search("typing.Union\[(.*)\]", anno).group(1).split(",")
                    intersect_valid_keys = list(set(annotation_json_type_map).intersection(types))
```

CodyCBakerPhD: Agree 100% on using `param.annotation.__args__` (as long as its not an `inspect._empty`) instead of the regex.

`typing._GenericAlias` might be a tad too general in terms of intended support since that includes other things we would have to add, like `List`; and would also require contingency for nesting stuff like `List[Set[Union[type, str]]] ` (which is just random example code I found related to use of  `_GenericAlias`, not code we use of course).

Another option would be the `get_origin()` function from `typing_inspect`; the syntax would be
`
typing_inspect.get_origin(param.annotation) == typing.Union
`
however, `typing_inspect` does not come in the python base and has to be `pip` installed, so maybe you would be against using that due to the extra `requirements`?

CodyCBakerPhD: Right you are; I keep forgetting `set `only casts the `keys` of a `dict`.

CodyCBakerPhD:
@bendichter Actually, if we just directly access the `param.annotation.__args__` directly, Union wraps these as a tuple so we can just iterate over it and we don't have to check the complicated typing stuff. See latest commit for this simpler implementation.

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -64,22 +67,29 @@ def get_schema_from_method_signature(class_method, exclude=None):
...
 
-    for param in inspect.signature(class_method).parameters.values():
-        if param.name not in exclude + ['self']:
+    for param_name, param in inspect.signature(class_method).parameters.items():
+        if param_name not in exclude:
             if param.annotation:
-                param_type = annotation_json_type_map[str(param.annotation).split("'")[1]]
+                param_type = [
+                    annotation_json_type_map[anno] for arg_types in param.annotation.__args__
+                    for anno in [str(arg_types).split("'")[1]] if anno in annotation_json_type_map
```
CodyCBakerPhD, line 75: Purposefully not using `.get()` here to avoid adding excess items to this list that would have to be removed or ignored afterward.

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -53,7 +54,9 @@ def get_schema_from_method_signature(class_method, exclude=None):
 
     """
     if exclude is None:
-        exclude = []
+        exclude = ['self']
+    else:
+        exclude = exclude + ['self']
```
bendichter, line 59: What is the advantage to doing it this way?

nwb_conversion_tools/json_schema_utils.py
bendichter: I don't understand this comprehension. Can you break it out for me? Why are we casting arg_types as a string?

nwb_conversion_tools/json_schema_utils.py
CodyCBakerPhD: Just clearer that it's always included by default, imo. Ultimately it's always included in the` exclude` list when it is invoked, but if someone looked at the code for the first time they'd have to examine the interiors of the loop to see how it always skips `self`.

nwb_conversion_tools/json_schema_utils.py
CodyCBakerPhD: It's essentially the same previous 3 lines of pre-Union version code. It always used to convert it to a string to extract the class name, but I just realized I could simply do `arg_types.__name__`. Fixed that and cleaned it up a tad on the latest commit.

But otherwise, the `Union` usage allows `.__args__` to be an iterable, so that's why that for loop is in the compression. For each argument in that iterable it passes it through the json type map, but only if that is a valid key (bypassing the need for the previous intersection stuff).

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -64,22 +67,29 @@ def get_schema_from_method_signature(class_method, exclude=None):
...
-    for param in inspect.signature(class_method).parameters.values():
-        if param.name not in exclude + ['self']:
+    for param_name, param in inspect.signature(class_method).parameters.items():
+        if param_name not in exclude:
             if param.annotation:
-                param_type = annotation_json_type_map[str(param.annotation).split("'")[1]]
+                param_type = [
+                    annotation_json_type_map[arg_type.__name__] for arg_type in param.annotation.__args__
+                    if arg_type.__name__ in annotation_json_type_map
+                ]
```
bendichter, line 76:
```suggestion
                param_type = [
                    annotation_json_type_map[arg_type] for arg_type in param.annotation.__args__
                    if arg_type in annotation_json_type_map
                ]
```

does this work?

CodyCBakerPhD: Unfortunately, no. The actual `arg_type` object in this case is a `type` class, and does not map properly onto the keys of the `annotation_json_type_map`. The `__name__ `of the `arg_type` does, however.

bendichter: ah, right, but we can fix that

bendichter: by changing the dict to

```python
    annotation_json_type_map = {
        bool: "boolean",
        str: "string",
        int: "number",
        float: "number",
        dict: "object",
        list: "array"
    }
```

CodyCBakerPhD: Ahh, clever... Adding now

CodyCBakerPhD:
@bendichter Working well on my test cases, and looks pretty clean. Let me know if you have any other suggestions remaining

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -53,33 +53,41 @@ def get_schema_from_method_signature(class_method, exclude=None):
...
+    for param_name, param in inspect.signature(class_method).parameters.items():
+        if param_name not in exclude:
             if param.annotation:
-                param_type = annotation_json_type_map[str(param.annotation).split("'")[1]]
+                param_type = [
+                    annotation_json_type_map[arg_type] for arg_type in param.annotation.__args__
+                    if arg_type in annotation_json_type_map
+                ]
+                assert len(param_type) == 1, \
+                    f"There must be only one valid annotation type that maps to json! {len(param_type)} found."
```
bendichter, line 77:
```suggestion
                if any(np.isin(param.annotation.__args__, annotation_json_type_map):
                    param_type = annotation_json_type_map[arg_type]
                else:
                    raiseValueError(f"There must be only one valid annotation type that maps to json! {len(param_type)} found.")
```

what about this?

CodyCBakerPhD: Not working out of the box. Here's a simple one liner for testing
`
np.isin((str,pathlib.Path), {str: "string"})
`

which is strange because
`
(str,pathlib.Path)[0] in {str: "string"}
`
works fine.

I'm thinking maybe numpy does some casting that interferes with the class types somehow...

CodyCBakerPhD: Could always replace the `np.isin` with
`
[x in annotation_json_type_map for x in param.annotation.__args__]
`
to the same effect, though. Would that work for you?

bendichter: how about now?

bendichter: [x in annotation_json_type_map for x in param.annotation.__args__]

yeah, that would work!

CodyCBakerPhD: Sadly, no. I had tried that as well, but pulling out the keys specifically doesn't change anything in the `np` version... Very, very strange.

I'll just use the manual compression then to try to get this down a couple more lines of code.

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -53,33 +54,45 @@ def get_schema_from_method_signature(class_method, exclude=None):
...
+        list: "array"
+    }
+
+    for param_name, param in inspect.signature(class_method).parameters.items():
+        if param_name not in exclude:
             if param.annotation:
-                param_type = annotation_json_type_map[str(param.annotation).split("'")[1]]
+                valid_args = [x in annotation_json_type_map for x in param.annotation.__args__]
+                if any(valid_args):
+                    param_type = [annotation_json_type_map[x] for x in np.array(param.annotation.__args__)[valid_args]]
```
CodyCBakerPhD, line 75: Casting `param.annotation.__args__` the second time as a `np.array` so it can be indexed by the boolean list `valid_args`. But the whole `np.array` can't be passed into the dictionary, so it's done elementwise in the compression.

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -53,33 +54,45 @@ def get_schema_from_method_signature(class_method, exclude=None):
...
+                else:
+                    raise ValueError("There must be only one valid annotation type that maps to json! "
+                                     f"{param.annotation.__args__} found.")
+
+                if len(set(param_type)) > 1:
+                    raise ValueError("Conflicting json parameter types were detected from the annotation! "
+                                     f"{param.annotation.__args__} found.")
```
CodyCBakerPhD, line 82: It's really easy at this point to test if multiple unique (i.e., conflicting) `json` output types were detected, which would be a problem.

PR #101, opened 2020-11-26
added lfp interface

CodyCBakerPhD:
@bendichter Finally added a general version (and a `Neuroscope` implementation) of the LFP interface; it's essentially a `RecordingExtractorInterface`, but with the `run_conversion` overloaded to put the electrical series in the right place (`NwbRecordingExtractors` always adds to acquisition).

Long term, I do yet again see some opportunity for mixin functionality to reduce code duplication; in particular I believe (but have not tested/debugged) most if not all specific LFP interfaces could simply directly inherit from their specific type of RecordingExtractorInterface, but have that `run_conversion` overload simply refer to a `LFPMixin` class that only has that one function associated with it.

But what is here works fine for now.

@alejoe91 Also, long term, we may way to consider the `NwbRecordingExtractor` requiring `nwb-conversion-tools`, and having a `util` function here for doing the `DataChunking` on the traces of a `memmap` object to reduce code duplication cross-ecosystems; the approach used in the `lfp` `run_conversion` is essentially the same as the important part of `add_electrical_series`, +/- the specific differences between where they go in the `NWBFile`.

P.S.: This is setup to merge into the `peyrache_debugging_and_refactoring` branch for readability, since it relies on the refactored auxiliary functions.


nwb_conversion_tools/baselfpextractorinterface.py
```diff
@@ -0,0 +1,99 @@
...
+        if 'LFP' not in ecephys_mod.data_interfaces:
+            ecephys_mod.add_data_interface(LFP(name='LFP'))
+
+        if isinstance(lfp_extractor.get_traces(), np.memmap):
+            n_bytes = np.dtype(lfp_extractor.get_dtype()).itemsize
+            buffer_size = int(buffer_mb * 1e6) // (lfp_extractor.get_num_channels() * n_bytes)
+            lfp_data = DataChunkIterator(
+                lfp_extractor.get_traces(),
+                buffer_size=buffer_size
+            )
```
bendichter, line 74:
```suggestion
        traces = lfp_extractor.get_traces()
        if isinstance(traces, np.memmap):
            n_bytes = np.dtype(lfp_extractor.get_dtype()).itemsize
            buffer_size = int(buffer_mb * 1e6) // (lfp_extractor.get_num_channels() * n_bytes)
            lfp_data = DataChunkIterator(
                data=traces,
                buffer_size=buffer_size
            )
```

CodyCBakerPhD:
TODO: await changes to RecordingExtractors resolving LFP disputes, most likely resulting in moving a lot of this code over to the `NwbRecordingExtractor.write_recording` function in spikeextractors.

CodyCBakerPhD:
@bendichter Ready for final review in conjunction with the [spikeextractor PR](https://github.com/SpikeInterface/spikeextractors/pull/507).

bendichter:
thanks, @CodyCBakerPhD 

1. can you handle the merge conflicts?
2. why use `subset_recording`? If we want to make this functionality shared across several classes, it would be better it just move it up a class. Is that possible?

CodyCBakerPhD:
@bendichter Yeah I'll fix these conflicts here really quick.

I did make the `subset_recording` function as a common function for both recording and lfp interfaces to reduce code duplication (otherwise, they both would have done that same 15-20 lines of code in each interface).

Let me take a minute here and consider if it could just be an internal method of BaseRecording, in which case lfp would just inherit it directly and it wouldn't be more broadly used across any other interfaces (Segmentation/Sorting, where it would be inappropriate). It ought to work out...

CodyCBakerPhD:
@bendichter Allrighty, merge conflicts fixed, and `subset_recording` moved as an internal method of the `BaseRecordingExtractor`, which makes it a lot easier to read and safer to use as well.

PR #102, opened 2020-11-26
Peyrache debugging and refactoring for lfp

CodyCBakerPhD:
@bendichter Various debugging and refactors for running with the new lfp interfaces as well as on the Buzsaki Peyrache set.

I'll go through and make specific comments on various changes.

The `NeuroscopeRecordingInterface` changes look a little scrambled but that's only because of the new auxiliary functions added at the top; these are intended to be used by all of the Recording/LFP/Sorting interfaces, hence why they aren't class specific but are still Neuroscope specific.


nwb_conversion_tools/basedatainterface.py
```diff
@@ -8,6 +8,10 @@
 
 class BaseDataInterface:
 
+    @staticmethod
+    def get_ecephys_metadata(data_file_path: str):
+        raise NotImplementedError("The get_ecephys_metadata method has not been configured for this class!")
```
CodyCBakerPhD, line 13: In generalizing the ability of Sorting extractors and LFP extractors to write ecephys metadata in the absence of an actual Recording extractor, I believe it would be useful to just have this as a static method for each specific Recording extractor, whose own `get_metadata()` will simply wrap to it.

nwb_conversion_tools/baserecordingextractorinterface.py
```diff
@@ -42,20 +42,14 @@ def get_metadata_schema(self):
...
-                    description='raw acquired data'
-                ),
-                Device=dict(
-                    name='device',
-                    description='ecephys probe'
-                )
+                Device=[],
+                ElectrodeGroup=[],
+                Electrodes=[],
+                ElectricalSeries=dict()
```
CodyCBakerPhD, line 50: I mentioned there currently isn't a way to force spikeextractors to automatically write default metadata since it relies on the entire metadata dictionary being passed as None. Well, if we just set the internals to empty lists it will have the same effect by autopopulating each required field with the defaults over there (which are similar to the defaults that were removed here).

nwb_conversion_tools/basesortingextractorinterface.py
```diff
@@ -37,24 +37,8 @@ def run_conversion(self, nwbfile: NWBFile, metadata: dict, stub_test: bool = Fal
...
-                nwbfile=nwbfile,
-                metadata=metadata
-            )
-            se.NwbRecordingExtractor.add_electrodes(
-                recording=recording,
-                nwbfile=nwbfile,
-                metadata=metadata
-            )
+        if write_ecephys_metadata:
+            add_ecephys_metadata(nwbfile, metadata)
```
CodyCBakerPhD, line 41: The core part of -40 to -57 now occurs in multiple parts of the code, making it easier to simply have an auxiliary util function (+40 to +41) for adding it to the nwbfile when/if needed

nwb_conversion_tools/basesortingextractorinterface.py
```diff
@@ -71,13 +55,14 @@ def run_conversion(self, nwbfile: NWBFile, metadata: dict, stub_test: bool = Fal
...
 
         for metadata_column in metadata['UnitProperties']:
-            property_descriptions.update({metadata_column['name']: metadata_column['description']})
-            for unit_id in sorting_extractor.get_unit_ids():
-                if metadata_column['name'] == 'electrode_group':
-                    data = nwbfile.electrode_groups[metadata_column['data'][unit_id]]
-                else:
-                    data = metadata_column['data'][unit_id]
-                sorting_extractor.set_unit_property(unit_id, metadata_column['name'], data)
+            if len(metadata_column['data']) == len(sorting_extractor.get_unit_ids()):
```
CodyCBakerPhD, line 58: This change block looks messy, but it's only the addition of this if condition. Basically, if there is disagreement between the dimensionality of processed spiking unit metadata (like cell type) and what is directly pulled out of the .res and .clu files, the metadata won't be added.

nwb_conversion_tools/nwbconverter.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, **source_data):
         self.data_interface_objects = {
             name: data_interface(**source_data[name])
             for name, data_interface in self.data_interface_classes.items()
+            if name in source_data
```
CodyCBakerPhD, line 59: Adding this allows a given Converter class to contain an optional data interface class (like a recording) that only gets instantiated if we also include it in the keys of the `source_data` input arguments (which is easier to control than adding `pop`-like functionality to the `init` of that converter).

CodyCBakerPhD:
TODO: as per discussion, adjusting the form of the `get_metadata` retrievals to operate near autonomously for return in unassociated custom NWBConverters.

CodyCBakerPhD:
Nearly all of this has been absorbed into other PRs more directly related to their necessity. Closing this branch.

PR #103, opened 2020-11-26
source_data at converter init

luiztauffer:
source_data is a dictionary with NameDataInterfaces as keywords, not a list to be unpacked


bendichter:
I think I'd rather it be defined by the signature of the constructor. What is the advantage of having it this way?

bendichter (approved):

PR #104, opened 2020-11-29
added and debugged neuroscope multi recording interface

CodyCBakerPhD:
@bendichter Simple wrapper for the [NeuroscopeMultiRecording extractor](https://github.com/SpikeInterface/spikeextractors/pull/512); inherits all relevant methods from its related single recording extractor, with the only conflict requiring generalization being the determination of the `session_path` from either the `file_path` or the `folder_path` argument.


CodyCBakerPhD:
@bendichter The NeurocopeMultiRecordingTimeExtractor is now merged on spike interface, so this PR ought to be ready to go.

PR #105, opened 2020-11-29
generalized the ecephys metadata for spikeglx so it can be called outside the class

CodyCBakerPhD:
@bendichter When using the `SIPickle` interfaces to read in a processed (synchronized) recording extractor, the electrode metadata format has to be directly added to the metadata prior to conversion to avoid filling with default values.

Specific to the use in the Tank lab, this PR generalizes this as a static method so it can be called in the conversion scripts prior to `run_conversion`.

PR pointing to `add_save_si_object_function` branch for readability.

### DIFF
--- a/nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
@@ -11,24 +11,20 @@ class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
 
     RX = SpikeGLXRecordingExtractor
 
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
ToDo: work out and implement the better way of retrieving `ecephys` metadata from cross-interface or cross-repo methods instead of static usage with manual parameter passing. Possible solution, as Ben suggested, could be to just always instantiate a recording extractor local to the `get_metadata` method that can use its internal workings in the way an extractor would if it were intended to be converted.

CodyCBakerPhD:
This is now unnecessary. Closing

PR #106, opened 2020-11-30
Using lazy imports to avoid having to append strict requirements

CodyCBakerPhD:
@bendichter Using a similar strategy to spikeextractors, this allows the format specific interfaces that use specific outside packages to import them without having to add the packages to the main requirements.

Currently the only two interfaces that do this are intan and neuroscope, but it sets the precedent early on do always use this strategy to avoid bloating the nwb-conversion-tools `setup`.


CodyCBakerPhD:
@bendichter Actually, let's hold off on reviewing this until we discuss the approach with `ecephys_metadata` and _static_ methods.

If we have any `static` methods for use across interfaces or what not we would want the install message and installation status to be set as a class attribute for each interface.

CodyCBakerPhD:
@bendichter Should be good for review now.

PR #109, opened 2020-12-01
fixed issue with reading annotations without __args__ attr

CodyCBakerPhD:
@bendichter @weiglszonja Quick hotfix needed for running nwb-conversion-tools when the annotation types don't have internal __args__ attribute.

### DIFF
--- a/nwb_conversion_tools/json_schema_utils.py
+++ b/nwb_conversion_tools/json_schema_utils.py
@@ -19,7 +19,7 @@ def dict_deep_update(d, u):
     return d
 
 
-def get_base_schema(tag=None, root=False, id_=None, **kwargs):
+def get_base_schema(tag=None, root=False, id_=None, **kwargs) -> dict:
     """Return the base schema used for all other schemas."""
     base_schema = dict(
         required=[],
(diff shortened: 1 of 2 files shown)


bendichter:
@CodyCBakerPhD thanks for fixing this. Two things here:
1. Let's start adding tests. I have written a small test illustrating how we should start. It does not cover every edge-case of the function, but ensure that the main usage is working.
2. I think there is a more elegant way to solve this- I don't like putting a single annotation in a 1-element list and then treating it as a list- but this code works and it's relatively urgent so I'll merge this and we can make it prettier later.

CodyCBakerPhD:
@bendichter

1. Are these tests currently setup for a continuous integration system, or would they just have to be run manually? Either way it's definitely a great thing to start doing for situations just like these.
2, Agreed, I'll open a draft to clean it up here in a bit.

bendichter:
@CodyCBakerPhD good question. They *would* work for CI, but importing `CEDRecordingExtractor` fails rn pending a release of spikeextractors. As soon as that is released, we can start using CI properly. Another option if we did't want to wait would be to point the pip install to the dev branch of spikeextractors. Alessio says he can get a release out within the week so I'd rather just wait in this case.

PR #110, opened 2020-12-05
add dynamic_nwb_converter and test

bendichter:
@luiztauffer and @CodyCBakerPhD 

Ultimately, I'd like to create a web interface where a user can specify their unique combination of inputs and we can generate an NWB converter for them on the fly. This would not allow for any custom data interfaces, but would at least allow users to mix and match the ones we have available. What do you think?


CodyCBakerPhD:
@bendichter The principle of the code and approach seems sound - I'll keep thinking and testing/debugging if there are ever oddities with particular interfaces that could cause any issues here

CodyCBakerPhD:
@bendichter Now absorbed into #328 - good to close?

PR #112, opened 2020-12-07
Annotation refactor; remove intan naming; fix default session start time

CodyCBakerPhD:
@bendichter This PR adds some stuff relevant to Syntalos

1. Allow timestamps arguments to be passed to recording extractor
2. Refactored that part of the new annotations fix you didn't like
3. Small bug fix to the default `session_start_time`; apparently if it's `datetime(1900,1,1)` you get really weird, hard to debug errors during the final `io.write` because the associated timestamp becomes negative. Updating this to `datetime(2000,1,1)` fixes that bug.

### DIFF
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -80,7 +80,8 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
             )
         return recording_extractor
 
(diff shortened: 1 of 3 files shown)


bendichter (changes requested):
Can we move the use_timestamps arg just to Syntalos? So far this seems like its pretty specific to them. If other conversions need to do this we can generalize later

CodyCBakerPhD:
@bendichter This branch should be good now, as per our meeting. The default session start time can indeed be pushed back to 1970, but not beyond that.

Branch now includes; the approved intan naming fix, session start time fix, and the annotation refactor.

PR #116, opened 2020-12-09
add possible config file

bendichter:
Can you guys take a look at this and let me know what you think?

### DIFF
--- a/neuroconv_setup.yaml
+++ b/neuroconv_setup.yaml
@@ -16,7 +16,7 @@ experiment_types:
 
     sessions:
       - output_path: filepath1.nwb
-        source_args:
+        source_data:
           plexon:
             filepath: rawdata1.plx
           deeplabcut:
@@ -53,4 +53,3 @@ experiment_types:
               session_start_time: "2020-11-10T21:19:09+00:00"
             Subject:
               subject_id: '002'
-
--- a/neuroconv_setup.yaml
+++ b/neuroconv_setup.yaml
@@ -8,7 +8,7 @@ experiment_types:
     metadata:
       NWBFile:
         session_description: "maze is shaped like a Y"
-    input_data_types:
+    source_data_types:
       - name: plexon
         data_interface: PlexonRecordingExtractorInterface
       - name: deeplabcut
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
@bendichter So this would be used to conjunction with a dynamically generated conversion class to both make that dynamic class, fill in a handful of required metainfo, and specify the required source_data arguments for the interfaces?

Would we support any conversion options like stub_test?

CodyCBakerPhD:
Also, this type of design would reflect an experiment where each session's data has been partitioned per maze; for situations where there is only one data file containing many mazes (usually extracted via custom interfaces as epochs, but we can't do that here since it doesn't support the custom interfaces), would we simply suggest that the user partition the data themselves?

neuroconv_setup.yaml
```diff
@@ -0,0 +1,56 @@
...
+  NWBFile:
+    lab: 'My Lab'
+    institution: 'My Institution'
+
+experiment_types:
+  - ymaze:
+    metadata:
+      NWBFile:
+        session_description: "maze is shaped like a Y"
+    input_data_types:
```
CodyCBakerPhD, line 11:
```suggestion
    source_data_types:
```

neuroconv_setup.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - name: plexon
+        data_interface: PlexonRecordingExtractorInterface
+      - name: deeplabcut
+        data_interface: DeepLabCutInterface
+
+    sessions:
+      - output_path: filepath1.nwb
+        source_args:
```
CodyCBakerPhD, line 19:
```suggestion
        source_data:
```

CodyCBakerPhD (commented):
Just a couple suggestions for naming styles; I haven't checked that everyone follows this 100%, but the stuff inside the main converters was standardized to "source_data" instead of "inputs" or "args". If we ever want to change that (which I'm perfectly fine with) we should update the standard everywhere else, too.

CodyCBakerPhD:
@bendichter I've moved this over to #280 as a part of the testing infrastructure there - good to close this?

CodyCBakerPhD:
@bendichter Now absorbed into #328 - good to close?

PR #117, opened 2020-12-10
exclude smrx_ch_inds from source schema

luiztauffer:
@CodyCBakerPhD would you mind checking if this doesn't break the conversions on your side?


CodyCBakerPhD:
@luiztauffer Works fine in conjunction with [mease-lab-to-nwb PR #21](https://github.com/catalystneuro/mease-lab-to-nwb/pull/21)

PR #118, opened 2020-12-12
Refactor recording subsets

CodyCBakerPhD:
@bendichter 

1. Cleaned up the recording subset approach a bit, so that it only calls the function if subsetting is to actually be performed

2. Improved optional annotation checking for primary converter `run_conversion`

3. Extended the primary converter `run_conversion` to allow file appending, and optional overwrite capability


bendichter:
@CodyCBakerPhD looks good. Do these changes rely on any other changes or are they good to merge?

CodyCBakerPhD:
@bendichter No reliance on other PRs for this one. Bumped version to indicate new feature

PR #119, opened 2020-12-14
updated data interface import structure to be similar to roi

CodyCBakerPhD:
@bendichter For the spikeextractor data interfaces with more than one class within them, changing the `__init__` imports to `*` to reduce code and mimic roiextractors


bendichter:
@CodyCBakerPhD From my perspective, either is fine here. If you prefer this, I'm happy to merge the change

CodyCBakerPhD:
@bendichter It does make it easier to add additional interfaces under an umbrella, as is often the case in these developmental periods. Otherwise, we have to think of what import lists to update each time, like in spikeextractors, and sometimes it's possible to forget* them.

*Actually, this PR is a simpler way of adding `NeuroscopeMultiRecordingInterface` to the list, which it wasn't originally.

bendichter:
Yeah, makes sense

PR #120, opened 2020-12-15
test interface schemas

bendichter:
* create list of all interfaces
* validate all interface schemas against json schema draft 7

### DIFF
--- a/nwb_conversion_tools/datainterfaces/__init__.py
+++ b/nwb_conversion_tools/datainterfaces/__init__.py
@@ -1,6 +1,37 @@
-from .neuroscopedatainterface import *
+from .neuroscopedatainterface import (
+    NeuroscopeRecordingInterface,
+    NeuroscopeLFPInterface,
+    NeuroscopeMultiRecordingTimeInterface,
+    NeuroscopeSortingInterface
+)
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@bendichter  These tests look good and fun when tried manually.

Did we want to...

1) use `unittest` to generate a test class that runs all of them and generates detailed reports in case of failure or warnings
2) use the toy data supported in spikeextractors to test the data functionality of those interfaces, similar to how the to-nwb-template repo works?

bendichter:
1. These tests run automatically when you run `pytest` at the root level of the project. That's how I would expect them to be called in a CI setting as well. Pytest can also run `unittest` classes. `unittest` offers some nice tools, but there isn't a reason to start using it until we want to use those tools.

2. Yes, this is far from complete testing for everything. I would actually prefer to set up a testing pipeline using [real data](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data) if we can. We are currently setting up similar pipelines [here](https://github.com/SpikeInterface/spikeextractors/pull/525) and [here](https://github.com/catalystneuro/roiextractors/pull/59), but waiting on [this issue](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/issues/29).

CodyCBakerPhD (approved):

PR #122, opened 2020-12-19
Add spikeglx lfp interface

CodyCBakerPhD:
@bendichter Added LFP interface for SpikeGLX format.

Also fixed a bug in `NwbConverter.run_conversion` where it always makes a new NWBFile object, even if you were trying to pass one in to be operated on and returned.


nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -56,4 +56,23 @@ def get_metadata(self):
...
+
+    def get_metadata(self):
+        metadata = dict(
+            Ecephys=dict(
+                LFPElectricalSeries=dict(
+                    name="LFP",
+                    description="Local field potential signal."
+                )
+            )
+        )
```
bendichter, line 76: This seems generic. Can it be put [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/452d456d15f0d2107077f28ad73c7b62512af454/nwb_conversion_tools/baselfpextractorinterface.py#L15) instead?

CodyCBakerPhD: Yeah, that's the better place for it. Updated!

nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -56,4 +56,11 @@ def get_metadata(self):
                 )
             )
         )
+
         return ecephys_metadata
+
+
+class SpikeGLXLFPInterface(BaseLFPExtractorInterface):
+    """Primary data interface class for converting the high-pass (ap) SpikeGLX format."""
```
bendichter, line 64: usually it's the low pass that's called "lfp"

CodyCBakerPhD: Whoops, that would left over from copying/pasting from the RecordingInterface. Fixed!

bendichter (approved):

PR #123, opened 2020-12-28
Ced metadata improvement

CodyCBakerPhD:
@bendichter @luiztauffer @alejoe91 Added electrode metadata to CED data interface.

Note that I have opted to simply include the probe file information as built-in here, since it was proving very awkward to include it as an extra data file associated with a given dataset, but not automatically co-occurring with the data files themselves.

If we ever need to extend support to other types of probes, we can deal with it again at that point - my suggestion would be to house the probe info here within the nwb-conversion-tools for use across multiple interfaces when/where needed, and add optional arguments for the type of probe the user wants to specify.


nwb_conversion_tools/datainterfaces/ceddatainterface.py
```diff
@@ -12,8 +15,63 @@ class CEDRecordingInterface(BaseRecordingExtractorInterface):
...
+
+    def get_metadata(self):
+        file_path = Path(self.source_data['file_path'])
+        session_id = file_path.stem
+
+        prb_channels = [
+            42, 36, 32, 38, 39, 47, 45, 37, 44, 46, 48, 35, 34, 40, 41, 43, 62, 60, 58, 56, 54, 52, 50, 33, 49, 51, 53,
+            55, 57, 59, 61, 63, 1, 3, 5, 7, 9, 11, 13, 30, 14, 12, 10, 8, 6, 4, 2, 0, 21, 27, 31, 25, 24, 16, 18, 26,
+            19, 17, 15, 28, 29, 23, 22, 20
+        ]
```
bendichter, line 40: What is going on here?

nwb_conversion_tools/datainterfaces/ceddatainterface.py
```diff
@@ -12,8 +15,63 @@ class CEDRecordingInterface(BaseRecordingExtractorInterface):
...
+                        name='Group1',
+                        description="Group1 electrodes."
+                    )
+                ],
+                Electrodes=[
+                    dict(
+                        name='group_name',
+                        description="The name of the ElectrodeGroup this electrode is a part of.",
+                        data=group_name
+                    ),
```
bendichter, line 64: You can't absorb this in the `SortingExtractor`?

nwb_conversion_tools/baserecordingextractorinterface.py
```diff
@@ -31,7 +31,8 @@ def get_metadata_schema(self):
         metadata_schema = super().get_metadata_schema()
 
         # Initiate Ecephys metadata
-        metadata_schema['properties']['Ecephys'] = dict(
+        metadata_schema['properties']['Ecephys'] = get_base_schema(tag='Ecephys')
```
bendichter, line 34: why is this tagged?

nwb_conversion_tools/datainterfaces/ceddatainterface.py
CodyCBakerPhD: SortingExtractors use strict `int`s to classify groups, not string group names.

nwb_conversion_tools/baserecordingextractorinterface.py
CodyCBakerPhD: Not sure, doesn't look familiar - this looks like a change inherited from elsewhere, I'll try to track it down.

nwb_conversion_tools/datainterfaces/ceddatainterface.py
CodyCBakerPhD: As per meeting, will be moved to the appropriate location on the `mease-lab-to-nwb`

PR #124, opened 2021-01-11
Add SortingExtractor object directly to NWBConverter object

CodyCBakerPhD:
@bendichter Finalized functionality and testing of the method where we take a SortingExtractor object that has already been instantiated, and add it into the internal interface classes and lists via the metaprogramming strategy.

One thing to discuss would be how `data_interface_classes` defaults to `None`, but it is now possible for someone to make a NWBConverter without any `source_data` with the intention of simply adding a SortingExtractor object to it. I've resolved this here with a bunch of `if` checks on the `None` status, but it might take less code to simply set the default `data_interface_classes` to an empty dictionary.

### DIFF
--- a/nwb_conversion_tools/nwbconverter.py
+++ b/nwb_conversion_tools/nwbconverter.py
@@ -7,7 +7,9 @@ from typing import Optional
 
 from pynwb import NWBHDF5IO, NWBFile
 from pynwb.file import Subject
+import spikeextractors as se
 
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
I'll keep this branch undeleted so the code is at least floating around somewhere if we ever decide we want it - but otherwise, the use of this has largely become the alternate method of appending a previously written NWBFile with new data types via on/off use of the `overwrite` argument.

PR #125, opened 2021-01-11
Dynamic sorting extarctor

bendichter:


CodyCBakerPhD:
This is being put on the long-term backburner for development/integration into pipelines. Sticking with append mode writing method for now

CodyCBakerPhD:
Closing this PR but keeping branch for future reference, if we ever want to revisit this approach.

Recommended approach is to use append more through `NwbSortingExtractor.write_sorting(sorting_extractor, ..., overwrite=False)` to add data from a custom `sorting_extractor` object.

PR #126, opened 2021-01-12
Data interface for Blackrock and OpenEphys

luiztauffer:
- [x] depending on https://github.com/SpikeInterface/spikeextractors/pull/538
- [x] depending on https://github.com/SpikeInterface/spikeextractors/issues/540 
- [x] depending on https://github.com/SpikeInterface/spikeextractors/pull/541
- [x] depending on https://github.com/SpikeInterface/spikeextractors/issues/543
- [x] depending on https://github.com/SpikeInterface/spikeextractors/issues/554
- [x] depending on https://github.com/SpikeInterface/spikeextractors/pull/557
- [x] depending on https://github.com/SpikeInterface/spikeextractors/pull/547
- [x] depending on https://github.com/SpikeInterface/spikeextractors/pull/571
- [x] depending on https://github.com/SpikeInterface/spiketoolkit/pull/440

### DIFF
--- a/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
@@ -117,3 +117,5 @@ class BlackrockSortingExtractorInterface(BaseSortingExtractorInterface):
 
     def get_metadata(self):
         """Auto-populates spiking unit metadata."""
+        metadata = super().get_metadata()
+        return metadata
--- a/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
@@ -66,7 +66,7 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
 
         return metadata
 
-    def run_conversion(self, nwbfile: NWBFile, metadata: dict = None, use_timestamps: bool = False, 
+    def run_conversion(self, nwbfile: NWBFile, metadata: dict = None, use_times: bool = False, 
                        write_as_lfp: bool = False, save_path: PathType = None, overwrite: bool = False, 
                        stub_test: bool = False):
         """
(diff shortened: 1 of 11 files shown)


nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
```diff
@@ -0,0 +1,25 @@
...
+        return get_schema_from_method_signature(
+            class_method=cls.RX.__init__,
+            exclude=['block_index', 'seg_index_index']
+        )
+
+    def get_metadata(self):
+        """Auto-fill as much of the metadata as possible. Must comply with metadata schema."""
+        metadata = super().get_metadata()
+    
+        return metadata
```
bendichter, line 25: If this is all that is there, you don't need the method at all. It will just default to the `get_metadata` of the super class

luiztauffer: working on this

luiztauffer: @bendichter se e.g. 2d2f2d9

bendichter:
looking good!

bendichter:
@luiztauffer do we really want `return_scaled` to default to `True`?

luiztauffer:
> @luiztauffer do we really want `return_scaled` to default to `True`?

hum, not really. It makes sense to leave True as default for SpikeExtractors, since its their current behavior, but not really for us, I'll change that

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
```diff
@@ -0,0 +1,121 @@
...
+
+    SX = se.BlackrockSortingExtractor
+
+    def __init__(self, filename: PathType, nsx_to_load: Optional[int] = None):
+        super().__init__(filename=filename, nsx_to_load=nsx_to_load)
+
+    def get_metadata(self):
+        """Auto-populates spiking unit metadata."""
+        metadata = super().get_metadata()
+        return metadata
```
bendichter, line 121: This seems like it should not be necessary

bendichter:
@luiztauffer this looks good but let's try to keep PRs modular. Can you separate blackrock and openephys?

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
bendichter: @luiztauffer this still seems unnecessary

bendichter (changes requested):
@luiztauffer Looks great! Just one minor point and could you fix the tests?

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
luiztauffer: you're right, I'll remove it

luiztauffer:
@bendichter locally the tests are passing, I believe this could be due to the pip-installed versions of some packages that still need to be released with updates (e.g. spikeextractors)

bendichter:
@luiztauffer can you try to push out releases or pin to branches so that the CI works?

bendichter:
why are you changing return_scaled to False?

PR #131, opened 2021-02-03
Add session_start_time to SpikeGLXRecordingInterface

weiglszonja:
Extract session_start_time from SpikeGLX meta file


bendichter:
@CodyCBakerPhD , @luiztauffer this isn't going to cause any problems, is it?

CodyCBakerPhD:
@bendichter Nope - the deep update in the base converter will merge the NWBFile fields together in the metadata, only using that one if it hasn't been set already.

PR #132, opened 2021-02-05
Imaging extractor interfaces

Saksham20:
adding interfaces for imaging extractors


bendichter (approved):

Saksham20:
@bendichter checks will pass when https://github.com/catalystneuro/roiextractors/pull/76 merges

PR #137, opened 2021-02-22
propagate buffer_mb argument from spikeextractors

CodyCBakerPhD:
@alejoe91 @luiztauffer Depends on https://github.com/SpikeInterface/spikeextractors/pull/578

The `buffer_mb` argument has been available at the `add_electrical_series` level for quite a while, but was never able to be specified at the highest level of `write_recording`.

Now that we are running multiple large-scale raw data conversions in parallel, a facet unlikely to soon change, this argument will give us a much greater degree of control over the processing resource specification.

### DIFF
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -77,7 +77,7 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
         )
         return recording_extractor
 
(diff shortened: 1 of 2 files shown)


bendichter:
I like it

bendichter (approved):

bendichter:
I like this but maybe we should wait for spikeexractors to release this feature?

CodyCBakerPhD:
@bendichter It's been merged to master on se - it should therefore be included in the big release of the return_scaled feature once we have that ironed out

alejoe91:
Looks good to me @CodyCBakerPhD !!

PR #139, opened 2021-02-28
add movie interface

CodyCBakerPhD:
@bendichter @luiztauffer @alejoe91 I recently encountered some `.mpg` files in my Buzsaki conversions, and after some work figured out how to add the data to the NWBFile in a way that can work with DANDI (DANDI does not accept non-NWB file types, so the old solution of writing in `external` mode doesn't work).

When developing the interface over there, I realized it was very generalizable to other file types, so figured it would be worth including on the main repo here. Let me know what you all think.

A couple things before I open this for final review:

- [x] - [Should the pytest for github actions utilize a tempfile location?]
- [x] - [Add a DataChunkIterator for larger files. Otherwise it tries to read the entire movie into RAM as the array.]


bendichter:
looks good, @CodyCBakerPhD!

CodyCBakerPhD:
# data=H5DataIO(mov, compression="gzip", chunks=(1, frame_shape[0], frame_shape[1], 1)) # took 2:21 and 15x filesize
                data=H5DataIO(mov, compression="gzip", chunks=(1, frame_shape[0], frame_shape[1], 3)) # took 1:45 and 13.8x filesize
                # data=H5DataIO(mov, compression="gzip", chunks=(5, frame_shape[0], frame_shape[1], 3)) # took 6:33 and 13.8x filesize
                # data=H5DataIO(mov, compression="gzip", chunks=(30, frame_shape[0]/2, frame_shape[1]/2, 1)) # took 42:55 and 21.65x filesize
                # data=H5DataIO(mov, compression="gzip", chunks=(5, frame_shape[0]/2, frame_shape[1]/2, 1)) # took 7:57 and 20.8x filesize
                # data=H5DataIO(mov, compression="gzip", chunks=(total_frames/50, 1, 1, 3))  # took 1:30:00 for 78 iter
                # data=H5DataIO(mov, compression="gzip")  # predicted to take 8 hours

CodyCBakerPhD:
@bendichter @alejoe91 @luiztauffer Hey all - the MovieInterface PR is now ready for full review and final approval. No rush, take your time - feel free to ask questions and leave comments on anything that could be improved.

In particular, one thing I wanted to bring up is the structuring of our data interfaces folder on this repo - now that I have a separate util file (due to lazy import) just for movie stuff, which is actually quite useful more generally for downstream conversion stuff acting on movies, perhaps it would be best to move the MovieInterface class and the utils.py file into their own folder within the data interfaces folder? So that the two are more obviously linked/dependent.

alejoe91 (approved):
Looks good to me @CodyCBakerPhD Haven't tested on real data though. My only question is why you need to use a `while` to loop through frames when you know the number of frames beforehand

nwb_conversion_tools/datainterfaces/moviedatainterface.py
```diff
@@ -0,0 +1,175 @@
...
+            maxshape.extend(frame_shape)
+            best_gzip_chunk = (1, frame_shape[0], frame_shape[1], 3)
+            tqdm_pos, tqdm_mininterval = (0, 10)
+            if chunk_data:
+                def data_generator(file, count_max):
+                    cap = cv2.VideoCapture(str(file))
+                    n_frames = 0
+                    success, frame = cap.read()
+                    # with tqdm(total=total_frames, desc="Writing movie data", position=0) as pbar:
+                    while success and n_frames < count_max:
```
alejoe91, line 122: Can't you use a for loop knowing the number of frames?

CodyCBakerPhD: Hmm, a fair point since I do extract the total_frames (to get the timestamps outside a data generator) prior to those later data copies, so yes that could be simplified. I'll work on that right now

CodyCBakerPhD:
Given Satra's approval of the current ImageSeries method of upload for DANDI, this should be clear for integration with the main branch.

I do still need to add capability to specify the module to write the series to, but this will make use of the `get_module` function from PR #156, and so I'll patch it in after both are merged.

bendichter (approved):

PR #141, opened 2021-03-04
Spikeglx metadata fix

CodyCBakerPhD:
@bendichter Small fix required for Manuel's use of the pipeline.

It appears as if, when the fileCreateTime field was added for the session_start_time extraction, it was not checked against the possibility of a SubRecording being the extractor type.


bendichter:
I don't understand why the session_start_time is related to the file created time.

bendichter:
oh this is the created time of the original data! got it, ok.

PR #142, opened 2021-03-15
version update

luiztauffer:


CodyCBakerPhD:
@luiztauffer I think we should finish review/approval/merge of some of the more valuable hanging PRs before a new release, so it's not just a synch update with spikeextractors. I'll go through some with Ben later to see if we can't clean up the 10 that are posted as of 3/15/21.

@luiztauffer If you in particular could [help review this one](https://github.com/catalystneuro/nwb-conversion-tools/pull/139), it's the biggest feature update to supported data type conversions to NWB.

luiztauffer:
@CodyCBakerPhD there's not just a sync with spikeextractors for this new release, there are other fixes and datainterfaces as well
But if you want to use it as a boost to merge many PRs that are left hanging, that's fine with me

I still think though it is useful to have more frequent releases than make massive merges of several PRs, like you propose. This allows us to know specific versions for which we are certain our converters are working

nwb_conversion_tools/nwbconverter.py
```diff
@@ -96,18 +96,26 @@ def get_metadata(self):
...
         assert (not save_to_file and nwbfile_path is None) or nwbfile is None, \
             "Either pass a nwbfile_path location with save_to_file=True, or a nwbfile object, but not both!"
+
         if conversion_options is None:
             conversion_options = dict()
         else:
             validate(instance=conversion_options, schema=self.get_conversion_options_schema())
 
-        if nwbfile is None:
+        if nwbfile is None and not Path(nwbfile_path).is_file():
```
CodyCBakerPhD, line 118: This will cause an error when `nwbfile_path` defaults, since Path(None) expects `str `not `NoneType`.

But I see we do not currently support the use case of reading an existing NWBFile, and then returning the object post conversion (i.e., nwbfile=None, nwbfile_path="my_nwbfile.nwb", save_to_file=False).

Also we do not currently support passing an NWBFile object, then writing it post conversion (i.e., nwbfile=my_nwbfile_object, save_to_file=True, nwbfile_path="my_nwbfile.nwb").

While Ben has made it clear that adding support for such things can stretch us thin since each use-case represents an additional mode we have to provide support for with increasing complexity along each dimension, we should at a bare minimum improve the docstring for this function to make it clear how it is intended to be used. Granted if a user is simply using it as intended in the standard template pipeline, with most of the arguments being defaulted, there shouldn't be any issue.

CodyCBakerPhD:
@luiztauffer With the merging of those other PRs this should be good to go if you can improve the `run_conversion` docstring as per my comment above. Then we should be good for a new release!

alejoe91:
Guys what's the ETA for this PR?

CodyCBakerPhD:
@alejoe91 I went ahead and made the changes myself - once the tests verify it didn't interfere with anything I'll merge and start the release process

CodyCBakerPhD (approved):

CodyCBakerPhD:
@luiztauffer OK, if you need to make something like it (that doesn't also break default behavior) go ahead and make a hotfix for v0.7.0, bump it to v0.7.1 in that PR and we'll release after that

Issue #140, opened 2021-03-01
new version release


luiztauffer:
Depending on:
- [x] https://github.com/SpikeInterface/spikeextractors/pull/571
- [x] https://github.com/SpikeInterface/spiketoolkit/pull/440
- [x] https://github.com/catalystneuro/nwb-conversion-tools/pull/126
- [x] https://github.com/SpikeInterface/spikeextractors/issues/590
- [x] https://github.com/SpikeInterface/spiketoolkit/issues/463
- [x] new release of roiextractors (https://github.com/catalystneuro/nwb-conversion-tools/issues/135)
- [x] new release of spikeextractors (https://github.com/SpikeInterface/spikeextractors)
- [x] new release of spikesorters (https://github.com/SpikeInterface/spikesorters)
- [x] new release of spiketoolkit (https://github.com/SpikeInterface/spiketoolkit)

@bendichter @alejoe91 @CodyCBakerPhD @Saksham20 
trying to keep track here of what changes should be incorporated in other packages version releases before the most recent changes in **nwb-conversion-tools** can be released. Feel free to edit this list

Saksham20:
@alejoe91 there is a small dependency issue with spikeextractors package on pip:
https://github.com/SpikeInterface/spikeextractors/issues/590

Saksham20:
will be releasing the latest roiextractors on pypi soon https://github.com/catalystneuro/roiextractors/pull/80

CodyCBakerPhD:
v0.7.0 was released a few days ago - closing this

PR #144, opened 2021-03-17
Limit creation of initial nwbfile object

luiztauffer:
fix for append to existing file

### DIFF
--- a/nwb_conversion_tools/nwbconverter.py
+++ b/nwb_conversion_tools/nwbconverter.py
@@ -133,7 +133,7 @@ class NWBConverter:
         else:
             validate(instance=conversion_options, schema=self.get_conversion_options_schema())
 
-        if nwbfile is None:
+        if nwbfile is None and nwbfile_path is not None and not Path(nwbfile_path).is_file():
             nwbfile_kwargs = metadata['NWBFile']
             if 'Subject' in metadata:
                 # convert ISO 8601 string to datetime
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ with open(path.join(here, 'README.md')) as f:
 
 setup(
     name='nwb-conversion-tools',
-    version='0.7.0',
+    version='0.7.1',
     description='Convert data to nwb',
     long_description=long_description,
     long_description_content_type='text/markdown',


luiztauffer:
@CodyCBakerPhD this should solve for the last PR problem

luiztauffer:
fix #140

CodyCBakerPhD:
That does fix it, but it seems hacky isn't very thin - what use case does this come up in and is there no way around it in the downstream stuff?

CodyCBakerPhD:
@luiztauffer What problem does this create in your conversion? All the possibilities I can image can be resolved by setting overwrite appropriately - is that the case or am I missing something?

luiztauffer:
this piece of code should not be executed if we're just adding data to an existing nwb file:
https://github.com/catalystneuro/nwb-conversion-tools/blob/c7a2989243d9077e4db0c4f4468fbff92fee9daf/nwb_conversion_tools/nwbconverter.py#L136-L151

CodyCBakerPhD:
This piece of code will overwrite the nwbfile object in that case, so I don't see how this could 'cause a problem' in any of our conversions:

https://github.com/catalystneuro/nwb-conversion-tools/blob/update_version/nwb_conversion_tools/nwbconverter.py#L158-L165

luiztauffer:
@CodyCBakerPhD  you're right, in principle it wouldn't. Nevertheless, we should not run that piece of code, even if that will be overwritten, a matter of good practice imo.
I see now that we could indeed use `or overwrite` in that first if statement (instead of `nwbfile_path is not None and not Path(nwbfile_path).is_file()`), maybe that would actually be better

CodyCBakerPhD:
Yeah, we were aware of that fact when we wrote this code, but it's such a small inefficiency it doesn't really matter (one dictionary, one completely minimal NWB object). While it's a matter of practice certainly, it's also (I would argue) a better matter of practice to not expand a single logical `if` check out to four logical places (for doing what we want here, you'd need `nwbfile_path is not None and not Path(nwbfile_path).is_file() and not overwrite`).

A better way would be to make a `conversion_tool` function for generating a default NWBFile, and call it appropriately inside the `save_to_file` blocks.

Point being, this patch is listed as a 'fix' for a current conversion, not a performance enhancement, and is delaying our latest release. Since it changes nothing in terms of functionality I'll go ahead with the new release while you work on improving the code checks for this.

luiztauffer:
@CodyCBakerPhD 

When appending data to an existing file we should not require metadata, and this for example will break the code. Example:

```
converter.run_conversion(
    metadata={},
    nwbfile_path=path_to_output_nwb, 
    overwrite=False,
)
```

Again, we are running a piece of code unnecessary in this case. Please accept the proposed fix or make a fix more of your taste here and let’s release it. The current nwb-conversion-tools release does not work for Movshon

CodyCBakerPhD:
@luiztauffer Thanks for providing the source of  your issues.

Ultimately, that's a failure of our (a) logical checks on the passed metadata structure prior to use and (b) non-safe dictionary retrieval. While your previous suggested changes would have forced your Movshon conversion through, they wouldn't have resolved that core issue so it would have come up again at a later point in time. And also would have broken other conversions without the proper `overwrite` interaction.

But of course I am always happy to implement these changes myself! Working on it now.

CodyCBakerPhD:
> @CodyCBakerPhD
> 
> When appending data to an existing file we should not require metadata

We should _always_ require metadata, and encourage that it be retrieved from the call to `my_converter.get_metadata()`. You never know what contained interfaces in that converter might contain information that isn't already in the NWBFile you're intending to append. In fact, two of our recent conversions (Tank, Mease) support this as an intended usage. I'm pretty sure that if you call `get_metadata()` in your conversion, instead of passing an empty dictionary, this error should not occur since `get_metadata()` from the super `NWBConverter` class will never return an empty dictionary - an importantly, will contain the keys at the source of this issue.

Anyway, [I've raised this as an issue with v0.7.0](https://github.com/catalystneuro/nwb-conversion-tools/issues/147) so we can refer back it it, and [made a separate PR that fixes that](https://github.com/catalystneuro/nwb-conversion-tools/pull/148). If it fixes your issue, then I'll close this PR.

Also, remember that we encourage (as a code style) the use of dictionary literals whenever possible. I.e., `metadata=dict()`.

PR #145, opened 2021-03-22
small fix to arguments for recording and lfp interfaces

CodyCBakerPhD:
@luiztauffer [When I propagated the `buffer_mb` option](https://github.com/catalystneuro/nwb-conversion-tools/pull/137) from spikeextractors to the recording class, I forgot to also propagate it to the LFP interfaces that inherit from it, but override that function in order to specify the lfp specific option.

### DIFF
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -1,6 +1,6 @@
 """Authors: Cody Baker and Ben Dichter."""
 from abc import ABC
-from typing import Union
+from typing import Union, Optional
 from pathlib import Path
 
 import spikeextractors as se
@@ -12,7 +12,7 @@ from .basedatainterface import BaseDataInterface
 from .utils import get_schema_from_hdmf_class
 from .json_schema_utils import get_schema_from_method_signature, fill_defaults, get_base_schema
 
-PathType = Union[str, Path, None]
+OptionalPathType = Optional[Union[str, Path]]
 
 
 class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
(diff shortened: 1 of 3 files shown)


nwb_conversion_tools/baserecordingextractorinterface.py
```diff
@@ -75,14 +75,13 @@ def run_conversion(
...
-      write_as_lfp: bool = False,
-      save_path: PathType = None, 
+      use_times: bool = False,
+      save_path: OptionalPathType = None,
       overwrite: bool = False,
       buffer_mb: int = 500
     ):
         """
-        Primary function for converting recording extractor data to nwb.
+        Primary function for converting high-pass recording extractor data to nwb.
```
luiztauffer, line 84: not necessarily `high-pass`, right? More like raw data

luiztauffer (approved):
looks good!

nwb_conversion_tools/baserecordingextractorinterface.py
CodyCBakerPhD: Is there any type of `RecordingExtractor` data that is not either high-pass or low-pass? If high-pass, this is the interface to inherit from and the `run_conversion` to use. If low-pass, then they should use the LFPDataInterface.

Note since these interfaces are specific to those in spikeextractors, it foregoes those types of raw (first level) data that are sampled at frequencies that could be considered low-pass, like ecog.

Correct me if I'm wrong though, I'm just basing it off what I've seen so far in spikeextractors.

luiztauffer: This is a minor comment on nomenclature, high-pass is filtered data, and most cases I believe the source data will be raw data (non-filtered).
High-pass filtered data is valuable for spike sorting for example, so maybe we would also like to provide an interface for that, storing in the `pynwb.ecephys.FilteredEphys` container.
we could think about generalizing the RecordingInterfaces to map ecephys traces like this:

raw data → acquisition
low-pass → pynwb.ecephys.LFP
high-pass (or any other filtered data) → pynwb.ecephys.FilteredEphys

CodyCBakerPhD: Ahh, I see - thanks! I'll adjust the docstring then and add the high-pass interface as a to-do (I know it can come up sometimes in the spikeinterfac pipelines)

PR #146, opened 2021-03-22
Add tutorials

CodyCBakerPhD:
@bendichter I adapted @luiztauffer's Movshon tutorials into two notebook tutorials for using NWB Conversion Tools for use in both the hackathon, and also in general. I've tested them and they work out of the box, but I was thinking it could also help if I added them to run with the continuous integration as well for a kind of 'first test' that a NWBConverter with simulated data (and later, the GIN data too of course!) for a RecordingInterface and SortingInterface works (kind of similar to the MovieInterface tests on artificial video, in that regard).

I'll also add a section in the readthedocs for that as well, but that will probably be in another PR devoted to readthedocs improvements.

- [x] Depends on [spikeextractors PR #606](https://github.com/SpikeInterface/spikeextractors/pull/606).


luiztauffer (approved):
looks good to me!

requirements-dev.txt
```diff
@@ -0,0 +1,14 @@
+pynwb==1.4.0
+tqdm==4.45.0
+natsort==7.0.1
+numpy==1.18.2
+pandas==1.0.3
+scipy==1.4.1
+h5py==2.10.0
+PyYAML==5.3.1
```
CodyCBakerPhD, line 8:
```suggestion
PyYAML==5.4
```

PR #149, opened 2021-03-23
Refactoring Interfaces for maintainability, inspectability, and documentability

sneakers-the-rat:
Hey everyone, hope this isn't out of line as a first PR. Disregard the large number of lines and files changed, this is a refactoring that should keep all functionality intact. 

The interfaces as structured are reasonably daunting for someone trying to familiarize themselves with the package, with a lot of long filenames & implicit structure. It took me a bit to realize that there was an inheritance structure such that there was a base type, then subtypes like recording extractors, etc. I refactored the code to reflect that, so now rather than 

`nwb_conversion_tools.datainterfaces.openephysdatainterface. OpenEphysRecordingExtractorInterface` 

one would import

`nwb_conversion_tools.interfaces.recording.open_ephys.OpenEphysRecordingExtractorInterface`

This should make things easier to maintain, inspect, and document. Re: documentation -- it could be pretty challenging to manually maintain the kind of API-level documentation structure one would need to document each of these, but now that structure can fall out of the file hierarchy using sphinx's autodoc. 

I also refactored some of the intra-package imports: all interfaces were imported in the top-level `datainterfaces.__init__.py` file, and then they were further imported with a `from datainterfaces import *` in the top-level module initialization. This is discouraged in python because it makes it challenging to know what is being imported & what modules the package will have in its root namespace without documentation, which is hard to maintain manually. It also requires those manual imports to be maintained. 

I've replaced that, and the `interface_list` that was used in the tests with 'type'-level imports: eg all recording types are imported in `recording.__init__.py` so one can import 
```python
from nwb_conversion_tools import interfaces
interfaces.recording.OpenEphysRecordingExtractorInterface
```

And then the helper function ``list_interfaces()`` programmatically returns all available interfaces. As a bonus it also can list available interfaces by type, like ``list_interfaces('recording')``

I left the object names alone because there seems to be both "interface" objects and "extractor" objects and i wasn't sure how those terms were being used.

One downside to this arrangement is that interfaces of different types from the same format (eg. open ephys sorting interface vs recording) are now in separate files. This could be solved with some simple namespace aliasing like a `interfaces.format.open_ephys.py` that imports the relevant classes, or with some additions to `list_interfaces`. 

Anyway, hopefully useful. going to work on some abstraction around specifying schemas now.


luiztauffer:
I like this:
> the helper function list_interfaces() programmatically returns all available interfaces. As a bonus it also can list available interfaces by type, like list_interfaces('recording')

As for the difference between DataInterface and Converter objects and other useful info:
https://github.com/catalystneuro/nwb-conversion-tools/blob/master/documentation/conversion_tools_structure.md

sneakers-the-rat:
It's a little implicit/fragile, would probably be better implemented by giving class attributes for things we want to index over, or make better use of the `_recurse_subclasses` function to be able to specifically find interfaces that are subclasses of a particular class (iirc i hardcoded it to use the base class), but it's the start of programmatic access to the interfaces, the eventual goal (for me! we should talk!) is to implement an `add_interface` method that lets you create an abstract notion of your data structure without having to manually alter the class, eg.
``add_interface('recording', 'open_ephys', file="data/recordings/rec_{subject_id}_{recording_id}.bin", **kwargs)``

where the user is able to specify how to pull metadata fields from file names (ie. by using [parse](https://github.com/r1chardj0n3s/parse)), so internally the object would look relative to some root directory for files that match the template pattern like `glob.glob('data/recordings/rec_*_*.bin')` and then for each of them `subject_id, recording_id = parse(template, "data/recordings/rec_mindy_0001.bin")` . The same could be done for metadata stored in other common formats like csv, json, .mat file like `subject_id = spec('some/matlab/metadata_file.mat', field=('subject', 'id'))` if, for example, some value was stored in a nested matlab struct (the bane of my existence).

CodyCBakerPhD:
PR has diverged widely in many respects - each one has been modularized, distributed, and raised along respective channels. Moving these changes to those PRs.

PR #150, opened 2021-03-23
Api docs

sneakers-the-rat:
Built on https://github.com/catalystneuro/nwb-conversion-tools/pull/149 , some API docs from docstrings <3


CodyCBakerPhD:
@sneakers-the-rat Would it be possible to uncouple this from #149? We try to keep PRs extremely modular when possible since different aspects and features can move and develop at different rates.

CodyCBakerPhD:
Closed via #216

PR #156, opened 2021-03-27
Add get_module

CodyCBakerPhD:
Given how often we make use of this function in individual conversions, I'd say its a valuable addition to our `conversion_tools`.

### DIFF
--- a/nwb_conversion_tools/conversion_tools.py
+++ b/nwb_conversion_tools/conversion_tools.py
@@ -3,12 +3,28 @@ from pathlib import Path
 import numpy as np
 import uuid
 from datetime import datetime
+from warnings import warn
 
 from pynwb import NWBFile
 from pynwb.file import Subject
 import spikeextractors as se
 
 from .json_schema_utils import dict_deep_update
+
+
+def check_module(nwbfile: NWBFile, name: str, description: str = None):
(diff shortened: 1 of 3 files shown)


luiztauffer (commented):
I agree this has potential to be re-used several times. Some comments:
- why use `description` instead of `name` to identify the processing module? Then if it doesn't exist we create a module with the name passed as description?
- Should we rename `check_module`, for sake of specificity? Examples: `get_processing_module`, `get_or_create_processing_module`...

CodyCBakerPhD:
@luiztauffer Thanks for the review!

1. The `name` is always used to identify the module; the `description` is checked against the module `description` if it exists (and if differs, throws warning), but I have now changed the default to "No description." in the case when the module is created without a specified `description`.
2. I agree on the name - `check_module` sounds like something that would return a boolean for the existence or non-existence of the module in the nwbfile. I've renamed to `get_module`, since all modules are `ProcessingModules` adding processing seems unnecessary, and the `_or_create` feature is a descriptive property that can be conveyed in the docstring. Suffice to say, this function will _always give_ you a module if you ask to _get_ one, so that ought to be appropriate.

luiztauffer (approved):
good to go!

PR #164, opened 2021-04-08
added module control for movie interface

CodyCBakerPhD:
## Motivation

Users of the movie data interface are unable to alter the location in the nwbfile where the data is added. This gives them control over whether it is added to `acquisition` (default) or a custom processing module. 

## How to test the behavior?

Tests have been added to the unit testing framework.

## Checklist


CodyCBakerPhD:
@luiztauffer If you can take a look when you get a chance

Saksham20:
@CodyCBakerPhD  I was facing an issue where the opencv package is a required install while doing `from nwb_conversion_tools import NWBConverter`. Since the init imports all the datainterfaces and the moviedatainterface raises an `ImportError`. I think this movieinterface should be implemented like the other datainterfaces where only while using the interface it check for the installation of open cv. I've pushed the changes corresponding to this https://github.com/catalystneuro/nwb-conversion-tools/pull/164/commits/6bc506300d48c05c6ed050d86cfb44c46327f015

nwb_conversion_tools/utils/json_schema.py
```diff
@@ -312,7 +312,7 @@ def get_schema_for_NWBFile():
             "description": "Experimental protocol, if applicable. E.g., include IACUC protocol"
         },
         "related_publications": {
-            "type": "string",
+            "type": ["string", "array"],
```
Saksham20, line 315: this is unrelated but a minor error that I found. Nwb specs have `related publications` as a str, list, tuple
https://pynwb.readthedocs.io/en/stable/pynwb.file.html#pynwb.file.NWBFile

CodyCBakerPhD: @Saksham20 Nice catch, can you work with Luiz on a separate PR for this?

CodyCBakerPhD:
> @CodyCBakerPhD I was facing an issue where the opencv package is a required install while doing `from nwb_conversion_tools import NWBConverter`. Since the init imports all the datainterfaces and the moviedatainterface raises an `ImportError`. I think this movieinterface should be implemented like the other datainterfaces where only while using the interface it check for the installation of open cv. I've pushed the changes corresponding to this [6bc5063](https://github.com/catalystneuro/nwb-conversion-tools/commit/6bc506300d48c05c6ed050d86cfb44c46327f015)

Nice catch, looks like that was introduced [19 days ago](https://github.com/catalystneuro/nwb-conversion-tools/blame/master/nwb_conversion_tools/datainterfaces/__init__.py#L36) with the addition of the `MovieInterface` to the general init. Thanks for the fix, too!

@luiztauffer or @bendichter, can one of you take a look and review the PR so we can get the new feature + this bug fix in?

CodyCBakerPhD:
@Saksham20 I've added you as a reviewer if you take a look at this as soon as you can - I'd like the fix you included in the next release.

Saksham20 (approved):

Issue #165, opened 2021-04-08
Support Axona data conversion to NWB (add AxonaDataInterface)


sbuergers:
I want to create an AxonaDataInterface to support converting [Axona data](http://space-memory-navigation.org/DacqUSBFileFormats.pdf) to NWB format. 

This should be possible for both raw, continuous data and unit data from Axona. 
I will implement them in that order.

## Checklist

- [x] Have you ensured the feature or change was not already [reported](https://github.com/catalystneuro/nwb-conversion-tools/issues)?
- [x] Have you included a brief and descriptive title?
- [x] Have you included a clear description of the problem you are trying to solve?
- [x] Have you checked our [Contributing](https://nwb-conversion-tools.readthedocs.io/en/master/developer_guide.html) document?

CodyCBakerPhD:
@sbuergers Sounds great! We'll take it one PR at a time.

- [ ] Raw ([PR #166](https://github.com/catalystneuro/nwb-conversion-tools/pull/166))
- [ ] Processed units

CodyCBakerPhD:
@sbuergers So on the processed units part, will that be in a separate PR? I think I might have seen that it was still in progress on spikeextractors

sbuergers:
@CodyCBakerPhD Yes, that will be in a separate PR. For the moment it might not even be needed, since the conversion to NWB should work with the standard converter from spikeextractors. Will have to see. I will update it here if I start it.

CodyCBakerPhD:
Is it a spikeinterface pipeline that appends file with write_sorting at the end? Or are they data files that have to be read and converted (even with just write_sorting)? I ask because occasionally that second option may be worthy of an interface of its own if there are unit properties (similar to channel properties for ecephys) to be included as metadata.

sbuergers:
Yes, there is a spikeinterface pipeline with a `write_sorting` and `write_recording` at the end. I agree that it would be nice to have the UNIT data also be part of axonadatainterface to 1. keep everything together, 2. keep it independent of spikeinterface if that is ever needed or useful, and 3. add additional metadata that the spikeinterface method does not provide like you say. 

That being said, there are other parts of this conversion project that are more pressing, and this might not strictly be necessary. Hence, I will get to this possibly after finalizing the TINT conversion if there are still hours left and/or if we decide that it is really needed after all.

CodyCBakerPhD:
[...]

PR #166, opened 2021-04-08
Axonadatainterface

sbuergers:
## Motivation

Incorporating data for the Axona system (only raw data for now, no unit data yet), as part of the Hussaini-to-nwb-conversion. I opened a corresponding issue here: https://github.com/catalystneuro/nwb-conversion-tools/issues/165#issue-853985052

__Depends on the following PRs__
- [x] https://github.com/SpikeInterface/spikeextractors/pull/605 (this is also specified in `requirements-dev.txt` for tests to pass, should be changed back once the SE PR is in master)
- [x] https://github.com/catalystneuro/nwb-conversion-tools/pull/174 (waiting for tests before merging to master)

## Checklist

- [x] Have you ensured the PR description clearly describes problem and the solution?


sbuergers:
The test actually passes if I run it locally, only for `AxonaRecordingExtractorInterface`. It crashes for the following (see `interfaces_that_crash`), but runs through for the rest. The test function is `test_interface_schemas()`:

```
interfaces_that_crash = [nwb_conversion_tools.datainterfaces.tutorialdatainterface.TutorialRecordingInterface,
                         nwb_conversion_tools.datainterfaces.tutorialdatainterface.TutorialSortingInterface]
```

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,132 @@
...
+        metadata['Ecephys'] = dict(
+            Device=dict(
+                    name="Axona",
+                    description="Axona DacqUSB, sw_version={}".format(par['sw_version']),
+                    manufacturer="Axona"
+                ),
+            ElectrodeGroup=dict(
+                name='Group0',
+                location='',
+                device='',
```
bendichter, line 120:
```suggestion
                device="Axona",
```

bendichter: the device field should link to the name of the device. This linkage is required for NWB

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+        
+        metadata['Ecephys'] = dict(
+            Device=dict(
+                    name="Axona",
+                    description="Axona DacqUSB, sw_version={}".format(par['sw_version']),
+                    manufacturer="Axona"
+                ),
+            ElectrodeGroup=dict(
+                name='Group0',
```
bendichter, line 118: Are there multiple tetrodes? If so, each should be its own group

sbuergers: I thought so, too. What threw me off was that making ElectrodeGroup a list of dicts failed the validation, and I wasn't exactly sure how to incorporate it then. Presumably by changing the metadata_schema, but that seems somewhat crude. I am also not sure why lists of dicts are not accepted, since those are used in other examples. For instance [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py).

And since the datatype is "object" in the schema, list should qualify. This is the relevant part of the metadata_schema:

```
      "ElectrodeGroup": {
        "required": [
          "location",
          "description",
          "device",
          "name"
        ],
        "properties": {
          "name": {
            "description": "the name of this electrode",
            "type": "string"
          },
          "description": {
            "description": "description of this electrode group",
            "type": "string"
          },
          "location": {
            "description": "description of location of this electrode group",
            "type": "string"
          },
          "device": {
            "description": "the device that was used to record from this electrode group",
            "type": "string",
            "target": "pynwb.device.Device"
          }
        },
        "type": "object",
        "tag": "pynwb.ecephys.ElectrodeGroup"
      },
```

This is the error when using a list:

```
ValidationError                           Traceback (most recent call last)
<ipython-input-45-27393e1000b0> in <module>
      1 # Validate metadata with metadata_schema from nwbconverter
      2 
----> 3 validate(
      4     instance=converter.get_metadata(),
      5     schema=converter.get_metadata_schema()

/mnt/d/spikeinterface/hussaini-to-nwb/venv/lib/python3.8/site-packages/jsonschema/validators.py in validate(instance, schema, cls, *args, **kwargs)
    932     error = exceptions.best_match(validator.iter_errors(instance))
    933     if error is not None:
--> 934         raise error
    935 
    936 

ValidationError: [{'name': 'Group0', 'location': '', 'device': '', 'description': 'Group 0 electrodes.'}] is not of type 'object'

Failed validating 'type' in schema['properties']['Ecephys']['properties']['ElectrodeGroup']:
    {'additionalProperties': False,
     'properties': {'description': {'description': 'description of this '
                                                   'electrode group',
                                    'type': 'string'},
                    'device': {'description': 'the device that was used to '
                                              'record from this electrode '
                                              'group',
                               'target': 'pynwb.device.Device',
                               'type': 'string'},
                    'location': {'description': 'description of location '
                                                'of this electrode group',
                                 'type': 'string'},
                    'name': {'description': 'the name of this electrode',
                             'type': 'string'}},
     'required': ['location', 'description', 'device', 'name'],
     'tag': 'pynwb.ecephys.ElectrodeGroup',
     'type': 'object'}

On instance['Ecephys']['ElectrodeGroup']:
    [{'description': 'Group 0 electrodes.',
      'device': '',
      'location': '',
      'name': 'Group0'}]
```

bendichter:
> The test actually passes if I run it locally, only for `AxonaRecordingExtractorInterface`. It crashes for the following (see `interfaces_that_crash`), but runs through for the rest. The test function is `test_interface_schemas()`:
> 
> ```
> interfaces_that_crash = [nwb_conversion_tools.datainterfaces.tutorialdatainterface.TutorialRecordingInterface,
>                          nwb_conversion_tools.datainterfaces.tutorialdatainterface.TutorialSortingInterface]
> ```

@sbuergers Please always include the script you ran and the full traceback when reporting bugs

nwb_conversion_tools/datainterfaces/axonadatainterface.py
bendichter: I see

1. An `object` is mapped to a dictionary in json-schema (see [json-schema docs](https://json-schema.org/understanding-json-schema/reference/object.html))
2. You are right, this should really accept a list of groups. @CodyCBakerPhD and  @luiztauffer , do you know what's going on here? Have we passed multiple groups through in other conversion pipelines? How did you accommodate that?

CodyCBakerPhD:
The schema error is from not using master branch of spikeextractors, which the requirements-dev is told to use for the CI tests until next release with the annotation fixes.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+                    'description': 'Path to Axona files.'
+                }
+            },
+            'type': 'object',
+            'additionalProperties': True
+        }
+        return source_schema
+
+    def __init__(self, **source_data):
+        super().__init__(**source_data)
```
CodyCBakerPhD, line 83: FYI These lines have no effect, all functions in class inheritance pass up to their `super` class automatically. You would only manually override such a function here if you wanted to do something different before or after calling the `super()`, which while not unheard of is not all that typical

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+                'filename': {
+                    'type': 'string',
+                    'format': 'file',
+                    'description': 'Path to Axona files.'
+                }
+            },
+            'type': 'object',
+            'additionalProperties': True
+        }
+        return source_schema
```
CodyCBakerPhD, line 80: @luiztauffer We should consider an automated way to form these schemas for neo based extractors. The reason the default `get_source_schema` of the `BaseRecordingExtractorInterface` doesn't work is because the spikeextractors `kwargs` are all dynamically passed down to neo without annotations, so only the docstrings contain that information (and sometimes not even that). Since this is something that is going to become more common as time goes on, would rather not have 15 extra lines of code in every interface if we can avoid it.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, **source_data):
+        super().__init__(**source_data)
+
+    def get_metadata(self):
+        """
+        Auto-fill metadata where possible. Must comply with metadata schema.
+        """
```
CodyCBakerPhD, line 88: @sbuergers You can safely remove docstrings for all interface functions unless you are overriding with new/extra arguments that need describing. This will let your interface inherit the hopefully more robust/descriptive (but maybe not quite yet) docstrings from the base class.

CodyCBakerPhD, line 88: @sbuergers You can safely remove docstrings for all interface functions unless you are overriding with new/extra arguments that need describing. This will let your interface inherit the hopefully more robust/descriptive (but maybe not quite yet) docstrings from the base class.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+        params_of_interest = [
+            'experimenter',
+            'comments',
+            'duration',
+            'sw_version',
+            'tracker_version',
+            'stim_version',
+            'audio_version'
+        ]
+        set_file = self.source_data['filename']+'.set'
```
CodyCBakerPhD, line 101: Note that if `filename=my_data_file.bin`, this will make `set_file=my_data_file.bin.set`. I usually do 
```
self.source_data['filename'].replace(".bin", ".set")
```
for that sort of thing. There's also more sophisticated ways using the pathlib `suffixes` vs. `stems`, but we probably don't need that.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+        par = parse_generic_header(set_file, params_of_interest)
+
+        # Add available metadata
+        # NOTE that this interface is meant to be used within an NWBconverter,
+        # which contains a much larger metadata_schema. As such a datainterface
+        # metadata by itself does not necessarily validate with its own schema!
+        metadata = super().get_metadata()
+        metadata['NWBFile'] = dict(
+            session_start_time=read_iso_datetime(set_file),
+            session_description=par['comments'],
```
CodyCBakerPhD, line 111: Oh, nice. I've never seen an acquisition system with comments in the headers before - one thing I might triple check on the example data files is that their data type and what not is read in correctly as a string type for the `par['comments']` and not something binary (I do see that there is a `.decode()` function in the parsing function, but haven't completely stepped through the logic).

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+            identifier=''.join(random.choices(string.ascii_uppercase +
+                                              string.digits, k=12)),
```
CodyCBakerPhD, line 113: Should really never have to specify this as it's handled with `uuid` in the main converter class (the thing that actually makes the NWBFiles)

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+            experimenter=[par['experimenter']]
+        )
+
+        metadata['Ecephys'] = dict(
+            Device=dict(
+                    name="Axona",
+                    description="Axona DacqUSB, sw_version={}".format(
+                        par['sw_version']),
+                    manufacturer="Axona"
```
CodyCBakerPhD, line 122: Ah hah, interesting. This won't actually get added to the NWBFile since we don't have it built in to do so. Could you raise an issue on spikeextractors and @ me to get a PR that allows it to get passed to the metadata?

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,136 @@
...
+                ),
```
CodyCBakerPhD, line 123: The value of the "Device" key in the dictionary should be a `list` containing as many dictionaries of `(name=..., description=...)` as devices you have. 

For properly formatted metadata at all steps, check the docstrings on spikeextractors: [like here, for devices](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L361-L362)

nwb_conversion_tools/datainterfaces/axonadatainterface.py
CodyCBakerPhD: @sbuergers See [Neuroscope](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/datainterfaces/neuroscopedatainterface.py#L81-L87) for an example of properly formatted (and schema passing) electrode groups.

Now, there's a lot of parts to that error so I'll start with some of the easier ones. Is that first output the result of asking for the metadata schema through our interface? If so, we'll have to think about a certain field that is 

Please try specifying only 'name' and 'description' in that dictionary in the list and see if it complains then. The schema errors aren't always that specific about _what part_ failed validation and I suspect it's the names of the keys.

@bendichter Yeah, for #2 there the ElectrodeGroups always takes a list of dicts. This is actually [hard enforced on the spikeextractors side](https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L428-L429), but schema validation is failing before it gets there.

CodyCBakerPhD:
@sbuergers Looking good so far! Let us know of any gaps in the documentation for contributing interfaces too since you're the first to use it ;)

Also, to get the continuous integration passing, simply modify the requirements-dev.txt to include an @<name of your axons branch> to the end of the spikeextractors line. It's not finding it right at the moment since it's only installing master.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: Ok, good to know.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: That makes sense.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: Right!

nwb_conversion_tools/datainterfaces/axonadatainterface.py
CodyCBakerPhD: Oh, but please keep the **class** docstring since that is specific to each interface.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: There is a comment line in the .set file, but I am not sure how often it actually contains comments. in my example it is just an empty string. 

If the decoding stays the same for all .set files (cp1252) then it should work. This was assumed by the Hussaini lab code as well.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: That makes sense, too. I think I just took this from some other datainterface class that I used as template: https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py

CodyCBakerPhD: Hmm, IDK what it's specified there. 

@luiztauffer do you know why you need t[hat identifier specified at the interface level](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py#L51) instead of the automated uuid in the converter class

nwb_conversion_tools/datainterfaces/axonadatainterface.py
luiztauffer: The schema is formed [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/54608fd49a7eff977cfb9d69d5cd73e90e78e927/nwb_conversion_tools/baserecordingextractorinterface.py#L42), which right now returns the definition of an object, not an array of objects (like [this example](https://json-schema.org/learn/miscellaneous-examples.html)).
Same thing will happen for `Device`.

When I wrote the first schemas, I had a key for each Device / ElectrodeGroup, instead of an array for all of them. But it seems like we’ve moved from that to have arrays in a single key instead, I guess it makes sense. 
I’ll make a PR that changes this for `BaseRecordingExtractorInterface`

nwb_conversion_tools/datainterfaces/axonadatainterface.py
luiztauffer: @CodyCBakerPhD thanks for pointing that out, I am extracting session_start_time and comment with the data interface, the identifier apparently was inherited from an older implementation. With uuid is better indeed, I modified that

luiztauffer: @sbuergers if you don't need to extract that info with your datainterface, you don't need to overwrite the basic metadata automatically generated

nwb_conversion_tools/datainterfaces/axonadatainterface.py
luiztauffer: https://github.com/catalystneuro/nwb-conversion-tools/pull/168

nwb_conversion_tools/datainterfaces/axonadatainterface.py
luiztauffer: @CodyCBakerPhD We could in principle form it from type annotation, getting the Path types.
There’s still a limitation to use the conversion GUI though, since we need to state if the source is a file or directory

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: @CodyCBakerPhD 

I currently have the following code:
```
metadata['NWBFile'] = dict(
            session_start_time=read_iso_datetime(set_file),
            session_description=par['comments'],
            session_duration=par['duration']+'s',
            experimenter=[par['experimenter']]
)
```

And they do seem to get added as "defaults", with the exception of `session_description` - is that what you meant?:

```
{
  "NWBFile": {
    "required": [
      "session_description",
      "identifier",
      "session_start_time"
    ],
    "properties": {
      "session_description": {
        "type": "string",
        "format": "long",
        "description": "a description of the session where this data was generated",
        "default": ""
      },
      "identifier": {
        "type": "string",
        "description": "a unique text identifier for the file",
        "default": "8d4cf566-4041-4518-ada7-61a36b1466b9"
      },
      "session_start_time": {
        "type": "string",
        "description": "the start date and time of the recording session",
        "format": "date-time",
        "default": "2020-10-04T11:07:07"
      },
      "experimenter": {
        "type": "array",
        "items": {
          "type": "string",
          "title": "experimenter"
        },
        "description": "name of person who performed experiment",
        "default": [
          "Abid"
        ]
      },
     ...
```

sbuergers: @CodyCBakerPhD this one took me a while to get my head around, but I think I understand what you are saying now. 

When I omit some of this information and follow for instance [neuroscope](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/datainterfaces/neuroscopedatainterface.py), then I am able to convert to an NWB file.

But if I leave the additional attributes, it crashes with the following error.

```
data_interface.run_conversion(nwbfile, metadata)
```

```
<ipython-input-107-71ef00e12343> in <module>
----> 1 data_interface.run_conversion(nwbfile, metadata)

/mnt/d/spikeinterface/nwb-conversion-tools/nwb_conversion_tools/baserecordingextractorinterface.py in run_conversion(self, nwbfile, metadata, stub_test, use_times, save_path, overwrite, buffer_mb)
    124             recording = self.recording_extractor
    125 
--> 126         se.NwbRecordingExtractor.write_recording(
    127             recording=recording,
    128             nwbfile=nwbfile,

/mnt/d/spikeinterface/spikeextractors/spikeextractors/extractors/nwbextractors/nwbextractors.py in write_recording(recording, save_path, overwrite, nwbfile, buffer_mb, use_times, metadata, write_as_lfp, write_scaled)
   1045                 io.write(nwbfile)
   1046         else:
-> 1047             se.NwbRecordingExtractor.add_all_to_nwbfile(
   1048                 recording=recording,
   1049                 nwbfile=nwbfile,

/mnt/d/spikeinterface/spikeextractors/spikeextractors/extractors/nwbextractors/nwbextractors.py in add_all_to_nwbfile(recording, nwbfile, buffer_mb, use_times, metadata, write_as_lfp, write_scaled)
    910         )
    911 
--> 912         se.NwbRecordingExtractor.add_electrode_groups(
    913             recording=recording,
    914             nwbfile=nwbfile,

/mnt/d/spikeinterface/spikeextractors/spikeextractors/extractors/nwbextractors/nwbextractors.py in add_electrode_groups(recording, nwbfile, metadata)
    440                         )
    441                     )
--> 442                     se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
    443                     warnings.warn(f"Device \'{device_name}\' not detected in "
    444                                   "attempted link to electrode group! Automatically generating.")

/mnt/d/spikeinterface/spikeextractors/spikeextractors/extractors/nwbextractors/nwbextractors.py in add_devices(recording, nwbfile, metadata)
    376                 )
    377             )
--> 378         assert all([isinstance(x, dict) for x in metadata['Ecephys']['Device']]), \
    379             "Expected metadata['Ecephys']['Device'] to be a list of dictionaries!"
    380 

AssertionError: Expected metadata['Ecephys']['Device'] to be a list of dictionaries!
```

When going through this I found that in spikeextractors, `nwbextractors.add_electrode_groups` 
it looks for the dictionary key "device_name" in `metadata['Ecephys']['ElectrodeGroup']`, which
contains "device" instead (for each list item). Since the key is not found, it assigns the default
value of 'Device' and after a few more steps it tries to add a "default" dictionary for device,
but this is not compatible with the schema. 

```
defaults = dict(
    name="Electrode Group",
    description="no description",
    location="unknown",
    device_name="Device"
)

...

for grp in metadata['Ecephys']['ElectrodeGroup']:
    device_name = grp.get('device_name', defaults['device_name'])
    if grp.get('name', defaults['name']) not in nwbfile.electrode_groups:
        if device_name not in nwbfile.devices:
            new_device = dict(
                Ecephys=dict(
                    Device=dict(
                        name=device_name
                    )
                )
            )
            se.NwbRecordingExtractor.add_devices(recording, nwbfile, metadata=new_device)
```

It might be sufficient to change `device_name = grp.get('device_name', ...)` to
`device_name = grp.get('device', ...)`.

sbuergers:
@CodyCBakerPhD @luiztauffer I followed Cody's advice and changed the `requirements-dev.txt` to pip install the `axonaextractor` branch so the CI does not fail. Checks pass now, so feel free to have a look.

I wasn't sure whether to try to write tests for Axona specifically and have not done so for now, because maybe this would anyway overlap with what Cody was working on regarding tests in `spikeextractors`? https://github.com/catalystneuro/spikeextractors/pull/55#issue-637961848

Let me know what you think.

luiztauffer:
@sbuergers can you check if [those changes](https://github.com/catalystneuro/nwb-conversion-tools/pull/174) will break anything for your interface?

luiztauffer (changes requested):
looking good, a few comments here and there

luiztauffer (approved):

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,316 @@
+"""Authors: Steffen Buergers"""
+import os
+import re
+import datetime
+import contextlib
+import mmap
+import numpy as np
+
+import spikeextractors as se
+
```
CodyCBakerPhD, line 10:
```suggestion
```
You're really close to the import style we use - first block being things that come with Python naturally, second block being manually installed one, and third block being local project imports. So `spikeextractors `and `pynwb` go together here.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,316 @@
...
+from pynwb import NWBFile
+from pynwb.behavior import Position, SpatialSeries
+
+from ..basedatainterface import BaseDataInterface
+from ..baserecordingextractorinterface import (
+    BaseRecordingExtractorInterface
+)
+
+from spikeextractors.extractors.nwbextractors.nwbextractors import check_module
```
CodyCBakerPhD, line 19: We actually just added a `get_module` util function to nwb-conversion-tools a while back. Can you refactor to include that instead?

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,316 @@
...
+    params (list or set): parameter names to search for.
+
+    OUTPUT
+    header (dict): dictionary with keys being the parameters that
+                   were found & values being strings of the data.
+
+    EXAMPLE
+    parse_generic_header('myset_file.set', ['experimenter', 'trial_time'])
+    """
+    header = {}
```
CodyCBakerPhD, line 40:
```suggestion
    header = dict()
```
Please use dictionary literals whenever possible. The time and place to use `{}` is under compression with values that can't easily be mapped as strings, which is somewhat rare.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,316 @@
...
+    params = set(params)
+    with open(filename, 'rb') as f:
+        for bin_line in f:
+            if b'data_start' in bin_line:
+                break
+            line = bin_line.decode('cp1252').replace('\r\n', '').\
+                replace('\r', '').strip()
```
CodyCBakerPhD, line 47:
```suggestion
            line = bin_line.decode('cp1252').replace('\r\n', '').replace('\r', '').strip()
```
Really minor, but similar to the `spikeextractors` changes, try to use a 120 line break for all wrapping.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,316 @@
...
+
+def read_iso_datetime(set_file):
+    """
+    Creates datetime object (y, m, d, h, m, s) from .set file header
+    and converts it to ISO 8601 format
+    """
+    with open(set_file, 'r', encoding='cp1252') as f:
+        for line in f:
+            if line.startswith('trial_date'):
+                date_string = re.findall(r'\d+\s\w+\s\d{4}$', line)[0]
```
CodyCBakerPhD, line 64: For things like this I always like to make 100% sure that this `date_string` format is something that is hard-coded by the acquisition system and not just something the user inputs?

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,316 @@
...
+                replace('\r', '').strip()
+            parts = line.split(' ')
+            key = parts[0]
+            if key in params:
+                header[key] = ' '.join(parts[1:])
+
+    return header
+
+
+def read_iso_datetime(set_file):
```
CodyCBakerPhD, line 56:
```suggestion
def read_axona_iso_datetime(set_file):
```
Very minor, but a name like `read_iso_datetime` implies it has a general purpose of converting any arbitrary ISO into a datetime, whereas this seems very particular to axona. Which is why it's in the axona interface itself, and not the general utils, but still, uniquely named functions avoid future confusion.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
```diff
@@ -0,0 +1,316 @@
...
+            'properties': {
+                'filename': {
+                    'type': 'string',
+                    'format': 'file',
+                    'description': 'Full path to Axona .set file.'
+                }
+            },
+            'type': 'object',
+            'additionalProperties': True
+        }
```
CodyCBakerPhD, line 91:
```suggestion
        source_schema = dict(
            required=['filename'],
            properties=dict(
                filename=dict(
                    type='string',
                    format='file',
                    description='Full path to Axona .set file.'
                )
            ),
            type='object',
            additionalProperties=True
        )
```

CodyCBakerPhD:
@sbuergers Only minor comments above - awaiting final review and merge of the numerous dependencies this PR has to move forward

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: Ok!

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: Good point. I noted it down for something to check with the Hussaini lab. More generally, it might then be safer to use something like this:

```
import dateutil.parser
yourdate = dateutil.parser.parse(datestring)
```

I tried it with a few different datestrings and it works nicely. Though even here, there can be ambiguous cases. 4.5.2021 can be read as 4th of May or 5th of April.

nwb_conversion_tools/datainterfaces/axonadatainterface.py
sbuergers: Ah, ok. Will do.

CodyCBakerPhD:
@sbuergers Any last additions before I merge this? Looks good to me

sbuergers:
@CodyCBakerPhD we can merge. :-)

PR #168, opened 2021-04-10
define type of arrays for ecephys schema

luiztauffer:
## Motivation

@sbuergers [found out](https://github.com/catalystneuro/nwb-conversion-tools/pull/166#discussion_r610653741) we were not defining `Device` and `ElectrodeGorup` as arrays on the base schema.

## How to test the behavior?
```python
from jsonschema import validate
#import any Converter using BaseRecordingExtractorInterfaces
converter = MyNWBConverter()
validate(instance=converter.get_metadata(), schema=converter.get_metadata_schema())
```

## Checklist

- [x] Have you ensured the PR description clearly describes problem and the solution?

### DIFF
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -37,13 +37,24 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
 
         # Initiate Ecephys metadata
         metadata_schema['properties']['Ecephys'] = get_base_schema(tag='Ecephys')
+        metadata_schema['properties']['Ecephys']['required'] = ['Device', 'ElectrodeGroup']
         metadata_schema['properties']['Ecephys']['properties'] = dict(
-            Device=get_schema_from_hdmf_class(Device),
-            ElectrodeGroup=get_schema_from_hdmf_class(ElectrodeGroup),
+            Device=dict(
+                type="array",
+                items={"$ref": "#/properties/Ecephys/properties/definitions/Device"}
+            ),
+            ElectrodeGroup=dict(
+                type="array",
+                items={"$ref": "#/properties/Ecephys/properties/definitions/ElectrodeGroup"}
+            ),
             ElectricalSeries=get_schema_from_hdmf_class(ElectricalSeries)
         )
-        metadata_schema['properties']['Ecephys']['required'] = ['Device', 'ElectrodeGroup', 'ElectricalSeries']
-        fill_defaults(metadata_schema, self.get_metadata())
+        # Schema definition for arrays
+        metadata_schema['properties']['Ecephys']['properties']["definitions"] = dict(
+            Device=get_schema_from_hdmf_class(Device),
+            ElectrodeGroup=get_schema_from_hdmf_class(ElectrodeGroup),
+        )
+        #fill_defaults(metadata_schema, self.get_metadata())
         return metadata_schema
 
     def subset_recording(self, stub_test: bool = False):
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -41,10 +41,12 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
         metadata_schema['properties']['Ecephys']['properties'] = dict(
             Device=dict(
                 type="array",
+                minItems=1,
                 items={"$ref": "#/properties/Ecephys/properties/definitions/Device"}
             ),
             ElectrodeGroup=dict(
                 type="array",
+                minItems=1,
                 items={"$ref": "#/properties/Ecephys/properties/definitions/ElectrodeGroup"}
             ),
             ElectricalSeries=get_schema_from_hdmf_class(ElectricalSeries)
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -48,8 +48,7 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
                 type="array",
                 minItems=1,
                 items={"$ref": "#/properties/Ecephys/properties/definitions/ElectrodeGroup"}
-            ),
-            ElectricalSeries=get_schema_from_hdmf_class(ElectricalSeries)
+            )
         )
         # Schema definition for arrays
         metadata_schema['properties']['Ecephys']['properties']["definitions"] = dict(
--- a/nwb_conversion_tools/basesortingextractorinterface.py
+++ b/nwb_conversion_tools/basesortingextractorinterface.py
@@ -24,7 +24,7 @@ class BaseSortingExtractorInterface(BaseDataInterface, ABC):
 
     def get_metadata_schema(self):
         metadata_schema = get_base_schema(
-            required=['SpikeEventSeries'],
+            # required=['SpikeEventSeries'],
             properties=dict(
                 SpikeEventSeries=get_schema_from_hdmf_class(SpikeEventSeries)
             )
--- a/nwb_conversion_tools/basesortingextractorinterface.py
+++ b/nwb_conversion_tools/basesortingextractorinterface.py
@@ -24,7 +24,6 @@ class BaseSortingExtractorInterface(BaseDataInterface, ABC):
 
     def get_metadata_schema(self):
         metadata_schema = get_base_schema(
-            # required=['SpikeEventSeries'],
             properties=dict(
                 SpikeEventSeries=get_schema_from_hdmf_class(SpikeEventSeries)
             )
--- a/nwb_conversion_tools/conversion_tools.py
+++ b/nwb_conversion_tools/conversion_tools.py
@@ -40,7 +40,7 @@ def get_default_nwbfile_metadata():
     metadata = dict(
         NWBFile=dict(
             session_description="no description",
-            session_start_time=datetime(1970, 1, 1),
+            session_start_time=datetime(1970, 1, 1).isoformat(),
             identifier=str(uuid.uuid4())
         )
     )
--- a/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
@@ -2,7 +2,6 @@
 import random
 import string
 import pytz
-import uuid
 from typing import Union, Optional
 from pathlib import Path
 import spikeextractors as se
--- a/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
@@ -2,6 +2,7 @@
 import random
 import string
 import pytz
+import uuid
 from typing import Union, Optional
 from pathlib import Path
 import spikeextractors as se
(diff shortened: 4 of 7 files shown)


luiztauffer:
@CodyCBakerPhD @sbuergers this should be tested with the already implemented converters to make sure they won't break. With Blackrock it's working

sbuergers:
@luiztauffer thanks for implementing this. I just tried it with axonadatainterface and that worked fine.

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -28,7 +28,7 @@ def get_base_schema(tag=None, root=False, id_=None, **kwargs) -> dict:
         required=[],
         properties={},
         type='object',
-        additionalProperties=False
+        additionalProperties=True
```
CodyCBakerPhD, line 31: Can you explain why this flag needs to be flipped?

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
```diff
@@ -46,9 +47,9 @@ def get_metadata(self):
         comment = nsx_file.basic_header['Comment']
 
         metadata['NWBFile'] = dict(
-            session_start_time=session_start_time_tzaware,
+            session_start_time=session_start_time_tzaware.strftime('%Y-%m-%dT%H:%M:%S'),
             session_description=comment,
-            identifier=''.join(random.choices(string.ascii_uppercase + string.digits, k=12))
+            identifier=str(uuid.uuid4())
```
CodyCBakerPhD, line 52: Again, [these lines called by the primary NWBConverter](https://github.com/catalystneuro/nwb-conversion-tools/blob/54608fd49a7eff977cfb9d69d5cd73e90e78e927/nwb_conversion_tools/conversion_tools.py#L44) mean that any other use of the identifier field (especially within interfaces) is redundant.

And in general principle, is really something that we should always leave to the main converter anyway.

nwb_conversion_tools/basesortingextractorinterface.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, **source_data):
 
     def get_metadata_schema(self):
         metadata_schema = get_base_schema(
-            required=['SpikeEventSeries'],
+            # required=['SpikeEventSeries'],
```
CodyCBakerPhD, line 27: Might as well just remove this entirely

nwb_conversion_tools/baserecordingextractorinterface.py
```diff
@@ -37,13 +37,26 @@ def get_metadata_schema(self):
...
+        )
+        # Schema definition for arrays
+        metadata_schema['properties']['Ecephys']['properties']["definitions"] = dict(
             Device=get_schema_from_hdmf_class(Device),
             ElectrodeGroup=get_schema_from_hdmf_class(ElectrodeGroup),
-            ElectricalSeries=get_schema_from_hdmf_class(ElectricalSeries)
         )
-        metadata_schema['properties']['Ecephys']['required'] = ['Device', 'ElectrodeGroup', 'ElectricalSeries']
-        fill_defaults(metadata_schema, self.get_metadata())
+        #fill_defaults(metadata_schema, self.get_metadata())
```
CodyCBakerPhD, line 59: If this is not needed, best to simply remove instead of comment out.

CodyCBakerPhD:
@luiztauffer Good catch, thanks for finding this! Ultimately the reason this has never been an issue before is because the NWBConverter does not run the validation as it does for source schemas. Hence they only way this is discovered is by manually attempting the validation, which few have a reason to do outside of thoroughly studying/learning the repo.

Can you add a validation step to the main converter as the final solution of this issue? Granted, writing actual unit tests for ensuring it works on data formats will have to wait until we get the GIN setup at this location, too.

nwb_conversion_tools/json_schema_utils.py
sbuergers: In my case, I believe I would not be able to specify `Electrodes` without that flag being True, because it is not in the schema. Of course it could be added to it instead. 

```
"Ecephys": {
    "required": [
      "ElectrodeGroup",
      "Device"
    ],
    "properties": {
      "Device": {
        "type": "array",
        "minItems": 1,
        "items": {
          "$ref": "#/properties/Ecephys/properties/definitions/Device"
        },
        "default": [
          {
            "name": "Axona",
            "description": "Axona DacqUSB, sw_version=1.2.2.16",
            "manufacturer": "Axona"
          }
        ]
      },
      "ElectrodeGroup": {
        "type": "array",
        "minItems": 1,
        "items": {
          "$ref": "#/properties/Ecephys/properties/definitions/ElectrodeGroup"
        },
        "default": [
          {
            "name": "Group0",
            "location": "",
            "device": "Axona",
            "description": "Group 0 electrodes."
          }
        ]
      },
      "ElectricalSeries": {
        "required": [
          "name"
        ],
        "properties": {
          "name": {
            "description": "The name of this TimeSeries dataset",
            "type": "string",
            "default": "ElectricalSeries"
          },
          "resolution": {
            "description": "The smallest meaningful difference (in specified unit) between values in data",
            "type": "number",
            "default": -1.0
          },
          "conversion": {
            "description": "Scalar to multiply each element in data to convert it to the specified unit",
            "type": "number",
            "default": 1.0
          },
          "starting_time": {
            "description": "The timestamp of the first sample",
            "type": "number"
          },
          "rate": {
            "description": "Sampling rate in Hz",
            "type": "number"
          },
          "comments": {
            "description": "Human-readable comments about this TimeSeries dataset",
            "type": "string",
            "default": "no comments"
          },
          "description": {
            "description": "Description of this TimeSeries dataset",
            "type": "string",
            "default": "Raw acquisition traces."
          }
        },
        "type": "object",
        "tag": "pynwb.ecephys.ElectricalSeries"
      },
      "definitions": {
        "Device": {
          "required": [
            "name"
          ],
          "properties": {
            "name": {
              "description": "the name of this device",
              "type": "string"
            },
            "description": {
              "description": "Description of the device (e.g., model, firmware version, processing software version, etc.)",
              "type": "string"
            },
            "manufacturer": {
              "description": "the name of the manufacturer of this device",
              "type": "string"
            }
          },
          "type": "object",
          "tag": "pynwb.device.Device"
        },
        "ElectrodeGroup": {
          "required": [
            "device",
            "description",
            "name",
            "location"
          ],
          "properties": {
            "name": {
              "description": "the name of this electrode",
              "type": "string"
            },
            "description": {
              "description": "description of this electrode group",
              "type": "string"
            },
            "location": {
              "description": "description of location of this electrode group",
              "type": "string"
            },
            "device": {
              "description": "the device that was used to record from this electrode group",
              "type": "string",
              "target": "pynwb.device.Device"
            }
          },
          "type": "object",
          "tag": "pynwb.ecephys.ElectrodeGroup"
        }
      }
    },
    "type": "object",
    "additionalProperties": true,
    "tag": "Ecephys"
  }
```

CodyCBakerPhD: If it's something required for a specific interface, it should be flipped within that interface. Given this is such a core function, there has to be a good general reason for flipping it (which there could very well be, but one example isn't enough for that). Note this isn't just for metadata schemas either, but anything that ever calls the base schema retrieval.

luiztauffer:
there will be a small incompatibility between this and https://github.com/catalystneuro/nwb-conversion-tools/pull/172
on metadata_schema

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
CodyCBakerPhD, line 52:
```suggestion
```

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
```diff
@@ -2,6 +2,7 @@
 import random
 import string
 import pytz
+import uuid
```
CodyCBakerPhD, line 5:
```suggestion
```

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
```diff
@@ -46,9 +47,9 @@ def get_metadata(self):
         comment = nsx_file.basic_header['Comment']
 
         metadata['NWBFile'] = dict(
```
CodyCBakerPhD, line 49: Does `pytz.timezone('EST').localize(session_start_time)` do something better than `.astimezone()` from `datetime`?

nwb_conversion_tools/json_schema_utils.py
CodyCBakerPhD, line 31:
```suggestion
        additionalProperties=False
```

nwb_conversion_tools/nwbconverter.py
```diff
@@ -121,11 +121,15 @@ def run_conversion(
...
             "Either pass a nwbfile_path location with save_to_file=True, or a nwbfile object, but not both!"
 
+        # Validate conversion options
         if conversion_options is None:
             conversion_options = dict()
         else:
             validate(instance=conversion_options, schema=self.get_conversion_options_schema())
 
+        # Validate metadata
+        validate(instance=metadata, schema=self.get_metadata_schema())
```
CodyCBakerPhD, line 131: @luiztauffer I'm wondering now if we should actually decouple this and move this one line to a PR of its own accompanied by a checklist of each supported data format to be tested/confirmed by us, and then followed by a version bump+release. B/c just adding this here could potentially cause a lot of breakages in previous uses

luiztauffer: @CodyCBakerPhD makes sense to me. Given that the user can also modify metadata, it would be better to have this separated as a class method as well

bendichter:
How do you guys feel about having a `validate_json` flag that is default `True` but can be turned to `False`? Would that ever be useful?

nwb_conversion_tools/json_schema_utils.py
luiztauffer: @sbuergers @CodyCBakerPhD I accepted Cody's turning that back to False, I can't remember anymore why I needed that as True. We can change it back in the next PR if we spot a reason why

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
luiztauffer: localize will get a tz-naive datetime and add the tz info to that. Not sure if astimezone would do something else

luiztauffer:
@CodyCBakerPhD I commented out the specific line that validates metadata at run_conversion, it's ready to be put back in place once we make sure other converters won't break. Let's merge it? other PRs are already dependent on those changes

nwb_conversion_tools/conversion_tools.py
```diff
@@ -40,7 +40,7 @@ def get_default_nwbfile_metadata():
     metadata = dict(
         NWBFile=dict(
             session_description="no description",
-            session_start_time=datetime(1970, 1, 1),
+            session_start_time=datetime(1970, 1, 1).isoformat(),
```
CodyCBakerPhD, line 43: @luiztauffer Why is this change necessary? @bendichter can correct me if I'm wrong, but the pynwb module specifically acts like the session_start_time should be a datetime object, not a string: https://github.com/NeurodataWithoutBorders/pynwb/blob/66ef3abeaa5bd9200ad926ddb82c0d53eba11281/src/pynwb/file.py#L207

Though the NWB Schema itself seems to specify ISO (https://nwb-schema.readthedocs.io/en/latest/format.html#type-specifications), I therefore assume that pynwb handles the translation.

luiztauffer: json schemas accept times only as strings, this needs to be a string to pass validation

CodyCBakerPhD:
The validation changes in the base converter could use some refactoring and improvement, but we'll do that in the validation focused PR.

CodyCBakerPhD (approved):

PR #172, opened 2021-04-12
updated options from se.NwbRecordingExtractor.write_recording()

luiztauffer:
Depends on:
- [x] https://github.com/catalystneuro/nwb-conversion-tools/pull/168
- [x] https://github.com/SpikeInterface/spikeextractors/pull/621

Problem: if data for conversion contains multiple ElectricalSeries objects, we are not able to to write it to nwb file using the current implementation.

This doesn’t work because on `se.NwbRecordingExtractor.write_recording()` we were only giving two options `raw` or `lfp` and fixing the metadata dictionary keys to be either `LFPElectricalSeries` or `ElectricalSeries`.
I proposed changes on this function on SpikeExtractors [here](https://github.com/SpikeInterface/spikeextractors/pull/621). Now we can have more freedom on the metadata side, example

```python
metadata['Ecephys'] = {
    'ElectricalSeries_raw': {'name': ‘raw data’, 'description': ‘description’},
    'ElectricalSeries_lfp': {'name': ‘lfp data’, 'description': ‘description’},
    'ElectricalSeries_thetaband': {'name': ‘theta data’, 'description': ‘description’},
    'ElectricalSeries_highpass': {'name': ‘filtered data’, 'description': ‘description’},
}
```

if we have one data interface for each, the converter will be able to store all those traces on the same nwbfile, all we need to do is to use conversion_options:

```python
conversion_options = dict(
    MyRecordingInterfaceRaw=dict(write_as='raw', es_key='ElectricalSeries_raw'),
    MyRecordingInterfaceLFP=dict(write_as='lfp', es_key=' ElectricalSeries_lfp'),
    MyRecordingInterfaceTheta=dict(write_as='processed', es_key=' ElectricalSeries_theta'),
    MyRecordingInterfaceHighpass=dict(write_as='processed', es_key=' ElectricalSeries_highpass'),
)
```

`se.NwbRecordingExtractor.write_recording()` recognizes three options for type of electrical series:
- `raw` will save it in acquisition
- `processed` will save it as FilteredEphys
- `lfp` will save it as LFP

### DIFF
--- a/nwb_conversion_tools/baselfpextractorinterface.py
+++ b/nwb_conversion_tools/baselfpextractorinterface.py
@@ -68,7 +68,8 @@ class BaseLFPExtractorInterface(BaseRecordingExtractorInterface):
             nwbfile=nwbfile,
             metadata=metadata,
             use_times=use_times,
-            write_as_lfp=True,
+            write_as='lfp',
+            es_key='LFPElectricalSeries',
             save_path=save_path,
             overwrite=overwrite,
             buffer_mb=buffer_mb
--- a/nwb_conversion_tools/baselfpextractorinterface.py
+++ b/nwb_conversion_tools/baselfpextractorinterface.py
@@ -10,8 +10,8 @@ from .baserecordingextractorinterface import BaseRecordingExtractorInterface
 OptionalPathType = Optional[Union[str, Path]]
 
 
-class BaseLFPExtractorInterface(BaseProcessingExtractorInterface):
-    """Previous class, for backward compatibility"""
+class BaseLFPExtractorInterface(BaseRecordingExtractorInterface):
+    """Primary class for all LFP data interfaces."""
 
     def get_metadata(self):
         metadata = dict(
@@ -25,10 +25,6 @@ class BaseLFPExtractorInterface(BaseProcessingExtractorInterface):
 
         return metadata
 
-
-class BaseProcessingExtractorInterface(BaseRecordingExtractorInterface):
-    """Primary class for all processed data interfaces."""
-
     def run_conversion(
       self,
       nwbfile: NWBFile,
@@ -41,7 +37,6 @@ class BaseProcessingExtractorInterface(BaseRecordingExtractorInterface):
     ):
         """
         Primary function for converting low-pass recording extractor data to nwb.
-
         Parameters
         ----------
         nwbfile: NWBFile
@@ -73,7 +68,7 @@ class BaseProcessingExtractorInterface(BaseRecordingExtractorInterface):
             nwbfile=nwbfile,
             metadata=metadata,
             use_times=use_times,
-            write_as_processed=True,
+            write_as_lfp=True,
             save_path=save_path,
             overwrite=overwrite,
             buffer_mb=buffer_mb
--- a/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
@@ -48,9 +48,8 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
         comment = nsx_file.basic_header['Comment']
 
         metadata['NWBFile'] = dict(
-            session_start_time=session_start_time_tzaware,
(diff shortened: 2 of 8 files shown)


nwb_conversion_tools/baselfpextractorinterface.py
```diff
@@ -69,8 +68,9 @@ def run_conversion(
             nwbfile=nwbfile,
             metadata=metadata,
             use_times=use_times,
-            write_as_lfp=True,
+            write_as='lfp',
+            es_key='LFPElectricalSeries',
```
CodyCBakerPhD, line 72: This default value for `es_key` does not align with your proposed default name on the spikeextractors PR

luiztauffer: this is specific for your baselfpinterface, it will overwrite the spikeinterface defaults

CodyCBakerPhD: I see now, hopefully that should work then

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
```diff
@@ -13,6 +13,7 @@
 from .interface_utils.brpylib import NsxFile
 
 PathType = Union[str, Path, None]
+OptionalPathType = Optional[Union[str, Path]]
```
CodyCBakerPhD, line 16: Both these lines are redundant (`Optiona[x] = Union[x, None]`). Use `PathType=Union[str, Path]` for non-optional path annotations and `OptionalPathType = Optional[PathType]` for non-required ones.

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
```diff
@@ -94,17 +105,23 @@ def run_conversion(self, nwbfile: NWBFile, metadata: dict = None, use_times: boo
...
             If True, will truncate the data to run the conversion faster and take up less memory.
         """
-        if self.source_data['nsx_to_load'] != 6:
-            write_as_lfp = True
+        if self.source_data['filename'].split('.')[-1][-1] == '6':
+            write_as = 'raw'
+        elif write_as not in ['processed', 'lfp']:
+            write_as = 'processed'
+
+        print(f'Converting Blackrock {write_as} traces...')
```
CodyCBakerPhD, line 113: Are you 100% sure that this is a general feature of the Blackrock format and not just something specific to the Movshon labs way of formatting filenames? In general, I don't see a problem having `write_as` specified in the `conversion_options` rather than a lower level auto-configuration, but not a fan of overriding `run_converion` if it can ever be avoided..

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
luiztauffer: I'm not sure what's your suggestion here

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
luiztauffer: https://github.com/NeuralEnsemble/python-neo/blob/072adfdaeffcae3b4f2bc4e8eb0be971a38874c5/neo/rawio/blackrockrawio.py#L24-L42

nwb_conversion_tools/datainterfaces/blackrockdatainterface.py
sbuergers: I think Cody was suggesting to put `PathType` in the brackets of `Optional`, and to remove `None` from `PathType = ...`. 
So part of the second line is actually redundant then. I didn't test this or know what this code is about to be honest, just interpreting while browsing through. Maybe there is a good reason for the `None` in the first line and it should not be in the second?

So:
```
PathType = Union[str, Path]
OptionalPathType = Optional[PathType]
```

CodyCBakerPhD, line 16:
```suggestion
PathType = Union[str, Path]
OptionalPathType = Optional[PathType]
```

CodyCBakerPhD (approved):

PR #174, opened 2021-04-28
fixes for dicts and schema

### LINKED ISSUE #173: dict_deep_update() not removing duplicated dictionaries
Problem: `dict_deep_update()` not removing duplicated dictionaries. Example:
```python
from nwb_conversion_tools.json_schema_utils import dict_deep_update

m1 = converter.data_interface_objects['BlackrockRecordingInterfaceRaw'].get_metadata()
m2 = converter.data_interface_objects['BlackrockRecordingInterfaceProcessed'].get_metadata()
dict_deep_update(m1, m2)
```

```
{'NWBFile': {'session_start_time': '2019-11-28T22:35:54',
  'session_description': ''},
 'Ecephys': {'Device': [{'name': 'Device_ecephys',
    'description': 'no description'},
   {'name': 'Device_ecephys', 'description': 'no description'}],
  'ElectrodeGroup': [{'name': 'ElecetrodeGroup',
    'description': 'no description',
    'location': 'no description',
    'device': 'Device_ecephys'},
   {'name': 'ElecetrodeGroup',
    'description': 'no description',
    'location': 'no description',
    'device': 'Device_ecephys'}],
  'ElectricalSeries_raw': {},
  'ElectricalSeries_processed': {}}}
```

working on a PR for that

### LINKED ISSUE #175: Electrodes is not defined on schema,
Problem: `Electrodes` is not defined on schema, giving errors such as the ones @sbuergers found when passing `Electrodes` on metadata['Ecephys'] 

working on it

luiztauffer:
- [x] Fix #173 
- [x] Fix #175 

Depends on:
- [ ] https://github.com/SpikeInterface/spikeextractors/pull/634

### DIFF
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -75,6 +75,7 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
                     data=dict(
                         type="array",
                         description="values for each row in this electrodes column"
+                        default=[np.nan]*len(self.recording_extractor.get_channel_ids())
                     )
                 )
             )
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -2,6 +2,7 @@
 from abc import ABC
 from typing import Union, Optional
 from pathlib import Path
+import numpy as np
 
 import spikeextractors as se
 from pynwb import NWBFile
@@ -74,7 +75,7 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
                     ),
                     data=dict(
                         type="array",
-                        description="values for each row in this electrodes column"
+                        description="values for each row in this electrodes column",
                         default=[np.nan]*len(self.recording_extractor.get_channel_ids())
                     )
                 )
--- a/nwb_conversion_tools/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/baserecordingextractorinterface.py
@@ -48,12 +48,36 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
                 type="array",
                 minItems=1,
                 items={"$ref": "#/properties/Ecephys/properties/definitions/ElectrodeGroup"}
-            )
+            ),
+            Electrodes=dict(
+                type="array",
+                minItems=0,
+                items={"$ref": "#/properties/Ecephys/properties/definitions/Electrodes"}
+            ),
         )
         # Schema definition for arrays
         metadata_schema['properties']['Ecephys']['properties']["definitions"] = dict(
             Device=get_schema_from_hdmf_class(Device),
             ElectrodeGroup=get_schema_from_hdmf_class(ElectrodeGroup),
+            Electrodes=dict(
+                type="object",
+                additionalProperties=False,
+                required=["name"],
+                properties=dict(
+                    name=dict(
+                        type="string",
+                        description="name of this electrodes column"
+                    ),
+                    description=dict(
+                        type="string",
+                        description="description of this electrodes column"
+                    ),
+                    data=dict(
+                        type="array",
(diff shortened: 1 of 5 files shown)


bendichter:
Thanks for tackling this, @luiztauffer. What if a user added a dict that has the same name, but different other properties? It looks like this would append that to the list, but maybe it would be better to overwrite. Perhaps we could use a `key` arg, which we could set to `name` which can act as the primary key to find duplicate dictionaries

luiztauffer:
@bendichter that makes sense, Devices for example have `name` as their unique feature
that made me think of Electrodes. which have `id` as their unique feature...

luiztauffer:
@sbuergers check these changes on Blackrock, it might help you see how to make the changes you need on your data interfaces

luiztauffer:
> Thanks for tackling this, @luiztauffer. What if a user added a dict that has the same name, but different other properties? It looks like this would append that to the list, but maybe it would be better to overwrite. Perhaps we could use a `key` arg, which we could set to `name` which can act as the primary key to find duplicate dictionaries. Also, if the entry does exist, I'd rather overwrite instead of skipping it

@bendichter it should be working in this way now

nwb_conversion_tools/json_schema_utils.py
```diff
@@ -5,6 +5,30 @@
...
+    Append a dictionary to a list of dictionaries.
+    If some dictionary already contains the same value as d[k], it gets 
+    replaced by the new dict.
+    Returns the updated list.
+    """
+    if k in d and len(l) > 0:
+        # Index where the value dictionary[k] exists in the list of dicts
+        ind = np.where([d[k] == i[k] for i in l])[0]
+        if len(ind) > 0:
+            l[ind[0]] = d
```
bendichter, line 24: should we replace this, or do an `update_dict`?

nwb_conversion_tools/baserecordingextractorinterface.py
```diff
@@ -48,12 +48,36 @@ def get_metadata_schema(self):
...
+                        type="string",
+                        description="name of this electrodes column"
+                    ),
+                    description=dict(
+                        type="string",
+                        description="description of this electrodes column"
+                    ),
+                    data=dict(
+                        type="array",
+                        description="values for each row in this electrodes column"
```
CodyCBakerPhD, line 77:
```suggestion
                        description="values for each row in this electrodes column"
                        default=[np.nan]*len(self.recording_extractor.get_channel_ids())
```

Would have to add `numpy as np` to imports at the top - but I suspect it would cause an issue to both to have the `data` field as non-required but lacking any default values (based on lines [595-598](https://github.com/catalystneuro/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L595-L598)) of `add_electrodes`).

CodyCBakerPhD: Still doesn't solve the greater issue Saksham is raising regarding the mapping of electrode ids that aren't either in strictly incrementing order from 0, or that span multiple interfaces with mixing/matching ids, but that's a bigger issue to think about.

CodyCBakerPhD:
@luiztauffer Are the dictionary update changes here necessary for the metadata and schema parts? If not could you separate those into a different PR (the dictionary changes will take more investigation, but the metadata/schema changes seem fine to me)

luiztauffer:
> @luiztauffer Are the dictionary update changes here necessary for the metadata and schema parts? If not could you separate those into a different PR (the dictionary changes will take more investigation, but the metadata/schema changes seem fine to me)

@CodyCBakerPhD yes, if there is more than one DataInterface in a converter. 
Which dictionary changes you're concerned about?

CodyCBakerPhD:
Any and all changes to functions in the json utils tend to have far-reaching implications which can be dangerous in our current low testing infrastructure. I'll think through them carefully then.

CodyCBakerPhD (approved):
More tests will be added in the future verifying the intend and use of each function. Pushing through for now.

PR #178, opened 2021-05-18
Port se write nwb

luiztauffer:
## Motivation

1. Port SpikeExtractors `write_to_nwb()` functions to nwb-conversion-tools. Many of the recent changes we pushed to SpikeExtractors have to do with writing to nwb files in complex ways, most of which are not really the scope of SpikeExtractors, but the scope of nwb-conversion-tools.

2. A tiny refactor, creating the utils module

## Depends on
- [x] https://github.com/catalystneuro/nwb-conversion-tools/pull/174


luiztauffer:
@bendichter @CodyCBakerPhD @Saksham20 @alejoe91 
here's what we discussed in the last time group meeting

bendichter:
@Saksham20 the methods you are modifying are being moved to NWB Conversion Tools, so no need to push them through to SpikeExtractors

CodyCBakerPhD:
@alejoe91 @bendichter This is ready to go with combined fixes from other PRs as well.

I'm thinking the safest way to go about this right now might be to keep the `spikeextractors` code intact as is and warn people that it may not have all the features that the nwb-conversion-tools one does (maybe even add a deprication warning over there), and give it a couple weeks or so before replacing the spikeextractors version with a wrapper to nwb-conversion-tools write_recording.

But from here on out, this will be the place to make changes to the functions that write recording extractors to nwbfiles.

PR #181, opened 2021-05-21
add_electrodes changes

### LINKED ISSUE #177: NWBConverter's get_metadata() for DyanamicTable data interfaces
For current implementation of `NWBConverter.get_metadata()` , itloops over all the `datainterface `classes and updates a master metadata dict [like here](https://github.com/catalystneuro/nwb-conversion-tools/blob/918b408fc631329895154a2b767dc79cd5c83b5f/nwb_conversion_tools/nwbconverter.py#L165): I'm having a problem for datainterface child classes that want to update an existing DyanamicTable in nwbfile. 
Specifically, for `Electrodes`: I have 2 datainterface classes, the 1st of which creates an `Electrodes` with 96 columns with custom data like so: 
```python
metadata["Ecephys"]["Electrodes"] = [
            dict(name="filtering", description="filtering", data=["1000Hz"] * 192),
            dict(
                name="group_name",
                description="electrode group name",
                data=['PMd array'] * 96,
            ),
            dict(name='location',
                 description='location',
                 data=['Pre-Motor Cortex, dorsal'] * 96)
        ]
```
Creates okay, now the 2nd one has the same colnames with different data to append to the `Electrodes`:
```python
metadata["Ecephys"]["Electrodes"] = [
            dict(name="filtering", description="filtering", data=["1000Hz"] * 192),
            dict(
                name="group_name",
                description="electrode group name",
                data=['M1 array'] * 96, #difference here
            ),
            dict(name='location',
                 description='location',
                 data=['M1 Motor Cortex'] * 96) #difference here
        ]
```
this does not work as the  `NWBConverter.get_metadata()`'s `dict_deep_update()` creates: 
```python
etadata["Ecephys"]["Electrodes"] = [
            dict(name="filtering", description="filtering", data=["1000Hz"] * 192),
            dict(
                name="group_name",
                description="electrode group name",
                data=['PMd array'] * 96,
            ),
            dict(name='location',
                 description='location',
                 data=['Pre-Motor Cortex, dorsal'] * 96),
            dict(name="filtering", description="filtering", data=["1000Hz"] * 192),
            dict(
                name="group_name",
                description="electrode group name",
                data=['M1 array'] * 96, #difference here
            ),
            dict(name='location',
                 description='location',
                 data=['M1 Motor Cortex'] * 96) #difference here
        ]
```
So it essentially appends the list. And when the `NWBConverter.run_conversion()` calls the `data_interface_object.run_conversion()` with this master metadata, the child data interface sees this and does not convert correctly. I'm using the `BlackRockDataInterface `to populate the `Electrodes` and even with the offset `extractor._channel_ids`, [on this line](https://github.com/catalystneuro/spikeextractors/blob/7716cff0e64ac3f97bc85b445b1eeb97be4372a0/spikeextractors/extractors/nwbextractors/nwbextractors.py#L602) it pulls the 0:95 elements repeatedly. 
Essentially, the `Electrodes` in the nwbfile does have 192 columns but `nwbfile.electrodes['group_name'] is `['M1 array'] * 192` while I would like: `['PMd array'] * 96 + ['M1 array'] * 96`.

I could think of some possibilities to avert this:

1.  [On this line](https://github.com/catalystneuro/nwb-conversion-tools/blob/918b408fc631329895154a2b767dc79cd5c83b5f/nwb_conversion_tools/nwbconverter.py#L165), we could do: `metadata=data_interface.get_metadata()`. But this in a way reduces the functionality as the user cannot then override this. It may make sense thus to implement a `set_metadata()` method in data interfaces in general, which can be used for overrides in custom conversion scripts. 
2. Intelligent dict_deep_update() for a dynamic table: appends all data within the data field only (so in this case it would become something like: 
```python
metadata["Ecephys"]["Electrodes"] = [
            dict(name="filtering", description="filtering", data=["1000Hz"] * 192),
            dict(
                name="group_name",
                description="electrode group name",
                data=['PMd array'] * 96 + ['M1 array'] * 96,
            ),
            dict(name='location',
                 description='location',
                 data=['Pre-Motor Cortex, dorsal'] * 96 + ['M1 Motor Cortex'] * 96)
        ]
```
The datainterface could then look for the length of the existing `Electrodes` table and pick and populate correct data from `metadata["Ecephys"]["Electrodes"][0]['data']`. 1st datainterface: `metadata["Ecephys"]["Electrodes"][0]['data'][:96]` and the 2nd one : `metadata["Ecephys"]["Electrodes"][0]['data'][97:193]`

Saksham20:
[fix #177](https://github.com/catalystneuro/nwb-conversion-tools/issues/177)

Since we are porting all the si write_nwb() code here, the changes that I and @bendichter discussed are implemented here. 

This PR is specifically for the `add_electrodes()`  method.  
The problem this will solve: multiple data interfaces appending to a electrodes table (for the Shenoy labs dataset that I've been bugging everyone about lately :) )

**Previously**, 
`add_electrodes()` pulled all data from the 1) metadata and  2)defaults, combined it and called `nwbfile.add_electrode()`. After this it would pull all custom columns data from `extractor.channel_properties` and add custom columns to the existing table using `set_dynamic_table_property()`.
This would create a problem when we would want to append to the electrodes table, since the subsequent calls to `nwbfile.add_electrode()` would require the kwargs corresponding to the newly created columns from the  `set_dynamic_table_property()` . 

**Currently**,
 these changes combine the all operations into one: kwargs for `nwbfile.add_electrode()` is pulled from 1) metadata, 2) defaults 3) `extractor.channel_properties`. Thus any subsequent calls dont require the metadata to have the additional fields (columns that were added by `set_dynamic_table_property()`)


nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -343,13 +343,47 @@ def add_electrodes(
...
+    for x in metadata['Ecephys']['Electrodes']:
+        name = x.pop('name')
+        if name not in elec_columns:
+            elec_columns[name] = dict(description=name)
+        if 'description' in x:
+            elec_columns[name]['description'] = x['description']
+        if 'data' in x:
+            assert len(x['data']) == recording.get_num_channels()
+            elec_columns[name]['data'] = x['data']
+
```
Saksham20, line 379: `elec_columns` contains all info from both the `channel_properties `of `recording `and the metadata.

luiztauffer:
@Saksham20 could you write a test for this concatenation on the electrodes table with multiple data interfaces?

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
+            # property 'offset' should not be in the NWB electrodes_table as not officially supported by schema v2.2.5
+
+            data = []
+            for chan_id in recording.get_channel_ids():
+                if prop in recording.get_channel_property_names(channel_id=chan_id):
+                    data.append(recording.get_channel_property(channel_id=chan_id, property_name=prop))
+                else:
+                    if len(data) > 0 and isinstance(data[-1], str):
+                        data.append('')
+                    else:
```
luiztauffer, line 371:
```suggestion
                    elif len(data) > 0:
```

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
+
+    for x in metadata['Ecephys']['Electrodes']:
+        name = x.pop('name')
+        if name not in elec_columns:
+            elec_columns[name] = dict(description=name)
+        if 'description' in x:
+            elec_columns[name]['description'] = x['description']
+        if 'data' in x:
+            assert len(x['data']) == recording.get_num_channels()
+            elec_columns[name]['data'] = x['data']
```
luiztauffer, line 385:
```suggestion
            if 'description' in x:
                elec_columns[name]['description'] = x['description']
            if 'data' in x:
                assert len(x['data']) == recording.get_num_channels()
                elec_columns[name]['data'] = x['data']
```

luiztauffer: if name exists in elec_columns you won't need these lines, is that right?

nwb_conversion_tools/utils/spike_interface.py
luiztauffer: don't you need to test this again here?

luiztauffer:
@Saksham20 this looks good, from what I understood the appending you need done for that converter is in the columns dimension, is that correct?
If not, can you help me understand how it would perform in other specific cases? Imagine two DataInterfaces writing to the same file:

1. Those are two sets of electrodes, they will belong to different ElectrodeGroups and will have different ids in nwbfile.electrodes, although within their own recording objects they might have overlapping ids (0, 1, 2…). The appending here would be in the rows dimension, with attention to ids

2. The second DataInterface recording has some channel properties that are different from the first (or vice-versa). How can we account for that? (not sure if this is a realistic scenario though)

3. The simplest case I believe would be if these two recording objects were related to the same channels, but they read files with different processing results (Blackrock does this) and we would like to avoid re-creating extra electrodes and properties in the electrodes table. That I believe we had working already.

nwb_conversion_tools/utils/spike_interface.py
Saksham20: metadata can override the description already set by channel properties

Saksham20:
> @Saksham20 this looks good, from what I understood the appending you need done for that converter is in the columns dimension, is that correct?
> If not, can you help me understand how it would perform in other specific cases? Imagine two DataInterfaces writing to the same file:
> 
> 1. Those are two sets of electrodes, they will belong to different ElectrodeGroups and will have different ids in nwbfile.electrodes, although within their own recording objects they might have overlapping ids (0, 1, 2…). The appending here would be in the rows dimension, with attention to ids
> 2. The second DataInterface recording has some channel properties that are different from the first (or vice-versa). How can we account for that? (not sure if this is a realistic scenario though)
> 3. The simplest case I believe would be if these two recording objects were related to the same channels, but they read files with different processing results (Blackrock does this) and we would like to avoid re-creating extra electrodes and properties in the electrodes table. That I believe we had working already.

yes, these set of changes are only to append new rows but same columns.
1. For a second set of electrodes, the user is expected to override the `extractor._channel_ids` accordingly.
2.  Yes, this case is not taken care of as of now. Could be implemented though, i'll try to work on this possibility also. (case of adding new columns and rows)
3. I think this could the a similar case as in 2 but where the second data interface only wants to add new columns to the table

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
-                name=str(metadata_column['name']),
-                description=str(metadata_column['description'])
-            )
+    elec_columns = {}  # name: description
+    property_names = set()
+    for chan_id in recording.get_channel_ids():
+        for i in recording.get_channel_property_names(channel_id=chan_id):
+            property_names.add(i)
+    for prop in property_names:
+        if prop not in ['gain', 'offset', 'location', 'name'] + list(exclude):
```
CodyCBakerPhD, line 359:
```suggestion
    exclude_names = set(['gain', 'offset', 'location', 'name'] + list(exclude))
    for prop in property_names:
        if prop not in exclude_names:
```
a) This removes duplicated entries from the collection
b) Also prevents reconstruction of the object at each iteration of the loop

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
-                and metadata_column['name'] not in defaults:
-            nwbfile.add_electrode_column(
-                name=str(metadata_column['name']),
-                description=str(metadata_column['description'])
-            )
+    elec_columns = {}  # name: description
+    property_names = set()
+    for chan_id in recording.get_channel_ids():
+        for i in recording.get_channel_property_names(channel_id=chan_id):
+            property_names.add(i)
```
CodyCBakerPhD, line 357:
```suggestion
    property_names = recording.get_shared_channel_property_names()
```

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -269,7 +269,8 @@ def add_electrodes(
     recording: se.RecordingExtractor, 
     nwbfile=None, 
     metadata: dict = None,
-    write_scaled: bool = True
+    write_scaled: bool = True,
+    exclude: tuple = ()
```
CodyCBakerPhD, line 273:
```suggestion
    exclude: Iterable[str]= []
```

Will have to include from `typing import Iterable` above, ofc. Might want to think about naming the option `exclude_property` to make that more explicit, too.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -291,6 +290,10 @@ def add_electrodes(
                                                     'data': [my_electrode_data]}, ...]
         where each dictionary corresponds to a column in the Electrodes table and [my_electrode_data] is a list in
         one-to-one correspondence with the nwbfile electrode ids and RecordingExtractor channel ids.
+    write_scaled: bool (optional, defaults to True)
+        If True, writes the scaled traces (return_scaled=True)
+    exclude: tuple
+        TODO - Add description
```
CodyCBakerPhD, line 296:
```suggestion
    exclude: Iterable[str] (optional, defaults to [])
        An iterable containing the string names of channel properties in the RecordingExtractor object to ignore when writing to the NWBFile.
```

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
         nwb_elec_ids = nwbfile.electrodes.id.data[:]
 
-    for metadata_column in metadata['Ecephys']['Electrodes']:
-        if (nwbfile.electrodes is None or metadata_column['name'] not in nwbfile.electrodes.colnames) \
-                and metadata_column['name'] not in defaults:
-            nwbfile.add_electrode_column(
-                name=str(metadata_column['name']),
-                description=str(metadata_column['description'])
-            )
+    elec_columns = {}  # name: description
```
CodyCBakerPhD, line 353:
```suggestion
    elec_columns = dict()
```

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
+                    if len(data) > 0 and isinstance(data[-1], str):
+                        data.append('')
+                    elif len(data) > 0:
+                        data.append(np.nan)
+            prop = 'location' if prop == 'brain_area' else prop
+            prop = 'group_name' if prop == 'group' else prop
+            elec_columns[prop] = dict(description=prop, data=data)
+
+    for x in metadata['Ecephys']['Electrodes']:
+        name = x.pop('name')
```
CodyCBakerPhD, line 378:
```suggestion
        name = x["name"]
```
Note this will actually modify the metadata object, not a copy of the object. So after calling `run_conversion`, the metadata dictionary will not reflect what was passed into it. This should be avoided.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
+        if name not in elec_columns:
+            elec_columns[name] = dict(description=name)
+        if 'description' in x:
+            elec_columns[name]['description'] = x['description']
+        if 'data' in x:
+            assert len(x['data']) == recording.get_num_channels()
+            elec_columns[name]['data'] = x['data']
```
CodyCBakerPhD, line 385: Note that you don't have to test for the conditional existence of these fields. The format assertion from [these lines ](https://github.com/catalystneuro/nwb-conversion-tools/pull/181/files#diff-296417a9c87b534b7bc88ddf071347b6a0d5241391a80448db87082db15c55b5R342-R344)ensures they exist so you can pull from them directly.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
+
+    for name, dat in elec_columns.items():
+        if 'data' not in dat and name not in defaults:
+            raise ValueError(f"no data for electrode column {name}")
```
CodyCBakerPhD, line 389: Given we have complete control over the content and structure of the `elec_columns `object, I don't think we need to `raise` this. I'll keep thinking through it to be sure.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
+
+            data = []
+            for chan_id in recording.get_channel_ids():
+                if prop in recording.get_channel_property_names(channel_id=chan_id):
+                    data.append(recording.get_channel_property(channel_id=chan_id, property_name=prop))
+                else:
+                    if len(data) > 0 and isinstance(data[-1], str):
+                        data.append('')
+                    elif len(data) > 0:
+                        data.append(np.nan)
```
CodyCBakerPhD, line 372: I see that you are trying to fill missing data with auto-detected formats to keep the entire data vector the length of the channel ids, but there is an issue here that could occur if a property exists for any channel after the first one, since the data will not have been appended at least once (the `len `will still be zero) meaning no auto-fill occurred prior to observing that property.

I understand and appreciate this approach as it differs from relying on `get_shared_channel_properties`, and I definitely think there is a place for it, but it has to be thought of very carefully.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -331,27 +332,65 @@ def add_electrodes(
...
+            prop = 'location' if prop == 'brain_area' else prop
```
CodyCBakerPhD, line 373: Clever one-line adjustment for resolving differences in name mapping. Nice!

CodyCBakerPhD:
@Saksham20 Thanks for working on this. The add_electrodes method definitely needs these kinds of improvements.

Expanding on what Luiz said, we really really need to get into that habit of writing tests every time we make, refactor, or improve an existing method. I've started this for the earlier `add_*` methods but haven't gotten to the electrodes yet so if you can get that started with your refactor here that would be great.

Other than that and the various comments, I am a tad confused about one thing. Are you saying that these changes, altogether, resolve your issue of writing from different channel ids from different recording interfaces all to the same electrode table?

nwb_conversion_tools/utils/spike_interface.py
bendichter: @Saksham20 is right. The drawback of using channel properties from the extractor is that they do not come with descriptions, so a user can specify just the name and the description in the metadata for a column where the data was already retrieved as a channel property

nwb_conversion_tools/utils/spike_interface.py
bendichter: Should default to empty tuple, not empty list. Empty lists as default args have [unexpected behavior](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument)

nwb_conversion_tools/utils/spike_interface.py
bendichter: you could use 

args = dict(x)
name = args.pop("name")

nwb_conversion_tools/utils/spike_interface.py
bendichter: Well first of all, those lines should be replaced by just validating json-schema. Second of all, I'd like to relax this constraint and allow for 'data' to not be defined in the metadata dict.

nwb_conversion_tools/utils/spike_interface.py
bendichter: see comment above. I'd like it to be set-able in the properties of the extractor

Saksham20:
> @Saksham20 Thanks for working on this. The add_electrodes method definitely needs these kinds of improvements.
> 
> Expanding on what Luiz said, we really really need to get into that habit of writing tests every time we make, refactor, or improve an existing method. I've started this for the earlier `add_*` methods but haven't gotten to the electrodes yet so if you can get that started with your refactor here that would be great.

Yes, I'll write tests for this method. 

> Other than that and the various comments, I am a tad confused about one thing. Are you saying that these changes, altogether, resolve [your issue (#177)](https://github.com/catalystneuro/nwb-conversion-tools/issues/177) of writing from different channel ids from different recording interfaces all to the same electrode table? I just find it surprising if it does, since the fix I was imagining it would really need was something to resolve the mapping of channel_ids to data elements.

Each datainterface would need to explicitly change the channel ids saved under `recording._channel_ids()`, a hack in the absence of any set method for now. Case by case:
1. if channel_id is not set explicitly, the recording would default to same channel ids and subsequently the write_electrodes would not append any electrodes to the table since these ids already exist. 
2. the values of new channel_ids set by the datainterface have some new values and some overlapping values with the previous datainterfaces, then yes the assertion [line 384 here](https://github.com/catalystneuro/nwb-conversion-tools/blob/07071754ba513b671bba3541cb11daa356c26557/nwb_conversion_tools/utils/spike_interface.py#L384) will error out. I can change the assertion message to reflect this possibllity, but would want to enforce each datainterface gives all different channel ids if they're attempting to.
3. If all channel_ids are different, then it correctly extends the table. Though as @luiztauffer pointed out, we could also take care of the [cases mentioned here](https://github.com/catalystneuro/nwb-conversion-tools/pull/181#issuecomment-846434065)

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: Interesting - did not know that. Modified to empty tuple.

We have exclusion type arguments scattered throughout spikeextractors (now here, of course) and a couple other places that are all list based so I'll check if they do it safely.

CodyCBakerPhD:
As per Ben's comments this PR will address the long-standing issue of differentiating light vs. heavy metadata in ecephys

What needs to be outside of the `add_electrode` method is to enhance/modify the metadata schema for electrodes to have data as an optional field, and encourage users or interface developers through documentation to set that data (i.e., _heavy_ metadata) as a channel property. The purpose of the `metadata` object then is primarily to convey optional descriptions (i.e., `light `metadata) of these properties, similar to how the `property_descriptions` argument works for `write_units`.

Even though the data field of the metadata object is now optional with preference to avoid, I do think it would nonetheless be helpful to take an extra step here to enhance it from a mere list (expected to be of length == channel_ids) to a dictionary of arbitrary size but which maps the channel ids to the data values as {channel_id: data_value}, which is actually quite similar to the inner workings of the RecordingExtractor channel properties. This would completely reconcile any and all differences between the two methods of setting the heavy metadata as well as be the most direct way of resolving Sakshams particular issue with multiple interfaces.

Of course, we can also discuss tomorrow the potential of simply removing support for the data field entirely, since the very existence of having two different ways to set it is partially at the root of this issue and would make this refactor considerably simpler.

nwb_conversion_tools/utils/spike_interface.py
Saksham20: I've pushed [changes here ](https://github.com/catalystneuro/nwb-conversion-tools/pull/181/commits/d25494a34bcc8fc0dacb873a416482269e5ec1c8)that dont require checking for existence of 'description' as json-schema would take care of it. Checking for 'data' is still there though.

nwb_conversion_tools/utils/spike_interface.py
Saksham20: yes, thats a case where it could fail, can you review [these changes](https://github.com/catalystneuro/nwb-conversion-tools/pull/181/commits/70e13898d4ccd111fe3c7e9dc9e39a73ce52f434)

CodyCBakerPhD:
@Saksham20 So as we discussed thoroughly at the last meeting - feel free to remove the data field entirely from the metadata here. Hence the only column data to write is that from the channel properties, but still pulling the descriptions for those from the metadata object.

Should make this code a lot cleaner from what it used to be since merging the two was always the hardest problem.

Saksham20:
- [ ] create tests

CodyCBakerPhD:
I'll take a look at the tests and other changes tomorrow

luiztauffer (approved):
I had to push some fixes for blackrock, openephys and spikeglx:

- How metadata_schema and metadata are formed, since now there’s no need for Electrodes field

- baselfprecording (and inheriting classes) needed similar fixes. @CodyCBakerPhD here again we have examples of unnecessary duplicated code, just for having a separate class for LFP. We’re likely to find this repetition of again and again, since I expect most Raw/Processed/LFP interfaces for the same data type to usually perform the same operations, the only difference being the local of writing within nwbfile, which is solved by just two args in `write_recording()` and proper set of metadata. 

For blackrock, openephys and spikeglx the changes in this PR are working fine now, I tested it locally with all combinations of Raw / Processed / LFP available

CodyCBakerPhD:
@luiztauffer Can you explain how to set electrode column descriptions in this new schema you've set up?

luiztauffer:
Hum, that’s right, there is no description associated with recording_extractor properties as far as I know, so using the default method as it is would not add this info.
We should still be able to do it by metadata though, then we could have default values for description set in the DataInterface `.get_metadata_schema()` and `.get_metadata()`, and allow users to edit this. But I haven’t tried that out! I will today
@Saksham20 , what do you think?

luiztauffer:
Alright, so here's what I did now:
- `BaseRecordingExtractorInterface` and `BaseLFPExtractorInterface` have now `Electrodes` present in their schema, as an optional item
- `SpikeGLXRecordingInterface` and `SpikeGLXLFPInterface` now inherit this schema and define default values for `['Ecephys']['Electrodes'][:]['description']` in their metadata. Any other DataInterface that wants to write descriptions to new electrodes columns should do the same
- The values for `description` can be overridden by the user when they customize their metadata

The electrodes columns' descriptions seem to be correctly written in the final NWB file, even if we invert the order in the metadata list `['Ecephys']['Electrodes'][:]`, which is good, its matching by name. 

One thing we should consider is: what to do when the user edits metadata and includes an Electrodes column name that does not exist? 
1. We allow this, for adding extra user-specific data.
2. We prohibit this and throw an error before conversion.

Right now, we have the option 2 in place.

nwb_conversion_tools/datainterfaces/ecephys/baselfpextractorinterface.py
```diff
@@ -1,25 +1,93 @@
...
+                    ),
+                    description=dict(
+                        type="string",
+                        description="description of this electrodes column"
+                    )
+                )
+            )
+        )
+        return metadata_schema
+
```
Saksham20, line 69: is this not a repetition from the base class `BaseRecordingExtractorInterface`

nwb_conversion_tools/datainterfaces/ecephys/baselfpextractorinterface.py
```diff
@@ -1,25 +1,93 @@
...
+            ],
+            ElectrodeGroup=[
+                dict(
+                    name=str(group_id),
+                    description="no description",
+                    location="unknown",
+                    device='Device_ecephys'
                 )
+                for group_id in np.unique(self.recording_extractor.get_channel_groups())
+            ],
```
Saksham20, line 87: repetition ?

Saksham20:
> One thing we should consider is: what to do when the user edits metadata and includes an Electrodes column name that does not exist?
> 
> 1. We allow this, for adding extra user-specific data.
> 2. We prohibit this and throw an error before conversion.
> 
> Right now, we have the option 2 in place. 

Yes, correct, it will throw a ValueError if the passed electrode column name is not written as a SI channel property prior check here https://github.com/catalystneuro/nwb-conversion-tools/blob/b6516cc440971498f2176ef532b75639ed05efd7/nwb_conversion_tools/utils/spike_interface.py#L416

This is necessary else, we will have to allow a 'data' field in addition to the name and description. But we decided against it.

Saksham20:
also i think its a better practice to put the json schema definitions in the root. Check the commit here https://github.com/catalystneuro/nwb-conversion-tools/pull/181/commits/cda4934fa0010cf35f195e49e334ab4cf6b1d15f

luiztauffer:
@Saksham20 could we save this change of definitions in the schema to another PR? It won't change any behavior here, but it might break something in the GUI.

Saksham20:
> @Saksham20 could we save this change of definitions in the schema to another PR? It won't change any behavior here, but it might break something in the GUI.

okay reverted, we can cherry pick this in another PR

luiztauffer:
@Saksham20 @CodyCBakerPhD are we all good with this PR? Let's merge it?

Saksham20:
> @Saksham20 @CodyCBakerPhD are we all good with this PR? Let's merge it?

apart from the two unresolved conversations above ([here](https://github.com/catalystneuro/nwb-conversion-tools/pull/181#discussion_r649797513)), i think we're good to go

nwb_conversion_tools/datainterfaces/ecephys/baselfpextractorinterface.py
CodyCBakerPhD: @Saksham20 is correct - if one follows the inheritance structure it is duplicated from its super

nwb_conversion_tools/datainterfaces/ecephys/baselfpextractorinterface.py
CodyCBakerPhD: Same as above

CodyCBakerPhD:
@Saksham20 Fantastic work on these tests and the `add_electrodes` refactor! After a thorough review and resolution of those two hanging items, I can definitely approve of this in its current state.

If @luiztauffer agrees we can merge.

CodyCBakerPhD (approved):

PR #187, opened 2021-05-26
metadata as optional and validated

luiztauffer:
fix [#190](https://github.com/catalystneuro/nwb-conversion-tools/issues/190)

## Motivation
1. allow metadata to be optional for `Converter.run_conversion()`. If not passed by the user, Converter will use the basic metadata formed with `Converter.get_metadata()`
2. validate metadata before running conversion. This will help us catch metadata-related errors before conversion starts

## How to test the behavior?
All conversion tests will necessarily go through this

### DIFF
--- a/nwb_conversion_tools/nwbconverter.py
+++ b/nwb_conversion_tools/nwbconverter.py
@@ -146,8 +146,7 @@ class NWBConverter:
         # Validate metadata
         if metadata is None:
             metadata = self.get_metadata()
-        else:
-            self.validate_metadata(metadata=metadata)
+        self.validate_metadata(metadata=metadata)
 
         # Validate conversion options
         if conversion_options is None:
--- a/nwb_conversion_tools/nwbconverter.py
+++ b/nwb_conversion_tools/nwbconverter.py
@@ -113,7 +113,7 @@ class NWBConverter:
 
     def run_conversion(
         self,
-        metadata: dict,
+        metadata: Optional[dict] = None,
         save_to_file: Optional[bool] = True,
         nwbfile_path: Optional[str] = None,
         overwrite: Optional[bool] = False,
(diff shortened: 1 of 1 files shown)


nwb_conversion_tools/nwbconverter.py
```diff
@@ -143,15 +143,19 @@ def run_conversion(
         assert (not save_to_file and nwbfile_path is None) or nwbfile is None, \
             "Either pass a nwbfile_path location with save_to_file=True, or a nwbfile object, but not both!"
 
+        # Validate metadata
+        if metadata is None:
+            metadata = self.get_metadata()
+        else:
+            self.validate_metadata(metadata=metadata)
```
CodyCBakerPhD, line 150: @luiztauffer The key assumption of course being that everything from self..get_metadata() is ensured to pass validation? No edge cases where it could fail?

I know we have built-in tests for this that check each individual data interface metadata, but a converter can take deep unions of that, and I just wonder if that could ever have the potential to give a post-unification metadata that isn't valid?

luiztauffer: >@luiztauffer The key assumption of course being that everything from self..get_metadata() is ensured to pass validation? 

That's correct.

> but a converter can take deep unions of that, and I just wonder if that could ever have the potential to give a post-unification metadata that isn't valid?

in this case I believe the best thing to do would be override the run_conversion method of the edgy case converter
```
def run_conversion(…)
	# code that reorganizes metadata and make it schema compatible
	metadata = ….
	super().run_conversion(metadata, args...)
```

luiztauffer: or actually, just override the `Converter.get_metadata()` would do it.

CodyCBakerPhD: Seems more complicated than I was thinking - I was simply proposing that we remove the else and always validate metadata, regardless of if it was passed or auto-generated.

```suggestion
            metadata = self.get_metadata()
        self.validate_metadata(metadata=metadata)
```

luiztauffer: of course, that makes sense

CodyCBakerPhD (approved):

CodyCBakerPhD:
I'll merge once the tests complete

Issue #46, opened 2020-06-25
pip installation


CodyCBakerPhD:
Upon first installation, calling `pip install nwb-conversion-tools` without a cloned repository is unable to find the package. Is this posted to PyPI yet?

luiztauffer:
this is because the package name on pypi is still `nwbn-conversion-tools`, we dropped the n recently. I can create a new project on pypi with the proper name. 
Now, should we stop updating the `nwbn-` project right away, leaving it only as legacy? @bendichter

bendichter:
Yes please do that

luiztauffer:
@CodyCBakerPhD @bendichter 
there were a lot of changes since I last contributed to the repo, so I'm not 100% up to date with everything. For now, we have it published on test Pypi and I think it would be wise to test it before we publish on the official pypi. You can do this:

```
pip install -i https://test.pypi.org/simple/ nwb-conversion-tools
```

bendichter:
@luiztauffer test pypi is usually used for testing deployment, not testing of the package itself. If there are problems, we can always cut a new release. Is there a problem with that?

luiztauffer:
so, what do you guys think, should we put out the current version already?
@bendichter @CodyCBakerPhD

bendichter:
Yeah go for it!

CodyCBakerPhD:
It has been pretty stable since v0.6.0... might need to update/expand the examples in the to-nwb-template repo for demonstration however, I'll check

PR #189, opened 2021-05-26
Fixes for SpikeGLX

luiztauffer:
Accommodate the possibility of writing spikeglx traces as raw (`ap` extension) or lfp (`lf` extension). 

@CodyCBakerPhD I think this will be important for Brody lab [here](https://github.com/catalystneuro/brody-lab-to-nwb/pull/5).
I tested on MovshonSpikeGLXConverter and it is working over [there](https://github.com/catalystneuro/movshon-lab-to-nwb/blob/main/tutorials/spikeglx_nwb_conversion_simple.ipynb).

### DIFF
--- a/nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
@@ -38,7 +38,8 @@ class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
         """Compile metadata schema for the RecordingExtractor."""
         metadata_schema = super().get_metadata_schema()
         metadata_schema['properties']['Ecephys']['properties'].update(
-            ElectricalSeries_raw=get_schema_from_hdmf_class(ElectricalSeries)
+            ElectricalSeries_raw=get_schema_from_hdmf_class(ElectricalSeries),
+            ElectricalSeries_lfp=get_schema_from_hdmf_class(ElectricalSeries)
         )
         return metadata_schema
 
@@ -50,12 +51,12 @@ class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
         session_id = file_path.parent.stem
 
         if isinstance(self.recording_extractor, SubRecordingExtractor):
-            n_shanks = int(self.recording_extractor._parent_recording._meta['snsShankMap'][1])
+            n_shanks = int(self.recording_extractor._parent_recording._meta.get('snsShankMap', [1, 1])[1])
             session_start_time = datetime.fromisoformat(
                 self.recording_extractor._parent_recording._meta['fileCreateTime']
             ).astimezone()
         else:
-            n_shanks = int(self.recording_extractor._meta['snsShankMap'][1])
+            n_shanks = int(self.recording_extractor._meta.get('snsShankMap', [1, 1])[1])
             session_start_time = datetime.fromisoformat(self.recording_extractor._meta['fileCreateTime']).astimezone()
         if n_shanks > 1:
             raise NotImplementedError("SpikeGLX metadata for more than a single shank is not yet supported.")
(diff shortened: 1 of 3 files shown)


nwb_conversion_tools/datainterfaces/spikeglxdatainterface.py
```diff
@@ -94,15 +95,38 @@ def get_metadata(self):
                     description="The name of the ElectrodeGroup this electrode is a part of.",
                     data=shank_group_name
                 )
-            ],
-            ElectricalSeries_raw=dict(
-                name='ElectricalSeries',
+            ]
+        )
+
+        if self.source_data['file_path'].split('.')[-2] == 'ap':
```
CodyCBakerPhD, line 101:
```suggestion
        if ".ap" in Path(self.source_data["file_path"]).suffixes:
```
I always prefer working with `pathlib` instead of string operations when dealing with file aspects. Simpler, cleaner, more explicit syntax.

CodyCBakerPhD:
The LF changes should go under the SpikeGLXLFPInterface.

Interesting to note that the snsShankmap isn't always present in the files, nice addition of safe retrieval.

luiztauffer:
There's no need for another data interface just for lfp imo, but in any case it will inherit these changes and should automatically save the traces as lfp anyway

CodyCBakerPhD:
> There's no need for another data interface just for lfp imo, but in any case it will inherit these changes and should automatically save the traces as lfp anyway

This is the way the recording interfaces have been modularized for about half a year now - it really helps keep complexity down within any single interface. For you, Blackrock is almost a special case since it also triggered the expansion into general processed recording types, which is probably what's influencing your opinion. Either way if it bothers you that much raise an issue on it as a general system wide refactor and bring it up in next meeting. Otherwise I went ahead and made the appropriate changes and checked that it works for SpikeGLX formatted data.

CodyCBakerPhD (approved):

luiztauffer:
> This is the way the recording interfaces have been modularized for about half a year now - it really helps keep complexity down within any single interface.

What complexity exactly goes down with a child class that simply copies the code previously present in its parent? You’re just disposing of an automatic way of detecting raw/processed for no benefit, perhaps an aesthetic one.

> Either way if it bothers you that much raise an issue on it as a general system wide refactor and bring it up in next meeting. Otherwise I went ahead and made the appropriate changes and checked that it works for SpikeGLX formatted data.

Where do you see the necessity for any general system wide refactor here? And why bring it up in a meeting?
The way I see it @CodyCBakerPhD, the changes were not appropriate, since it will require changes in another repo to adjust to your special LFP-named Class. So you broke something (that was working for both uses) for no gain.

**But the important thing here is this:** please, next time you review a PR, do not enforce a change and merge it before the author’s review. It’s a collaborative project after all.

CodyCBakerPhD:
> The way I see it @CodyCBakerPhD, the changes were not appropriate, since it will require changes in another repo to adjust to your special LFP-named Class. So you broke something (that was working for both uses) for no gain.

I tested this on actual SpikeGLX data using the simplest code possible for going through the interfaces and converters - what exactly broke? Also we've been using this exact differentiating structuring for multiple projects over several months with nothing breaking at the core level, so there's nothing special about it. Any changes required to any repos would once again be those which did not (a) version pin, or (b) are currently in development and perhaps are not utilizing the proper interfaces for the proper data types (a _user_ error in that case).

However, your changes to the SpikeGLXRecordingInterface would have actually had no propagated effects to any properly used SpikeGLXLFPInterfaces currently in use, as they do not inherit from the SpikeGLXRecordingInterface. I simply corrected that. If it's broken something for one of your conversions raise a subsequent issue with detailed code outlining the precise problem and perhaps a subsequent PR fixing the specific issue. That keeps the average size of the changelogs down and leaves a nice trail of documentation for cause and effect.

> Where do you see the necessity for any general system wide refactor here?

A while back, these `es_key` naming changes you made to the metadata in the processed recording feature additions were adjusted at certain base levels but not entirely reconciled at each individual interface level. That's the refactor that would have to update and ensure that.

> And why bring it up in a meeting?

If you're questioning why there are distinct interfaces for distinct data types to begin with, and actively designing format APIs like SpikeGLX around them because you disagree with them, then that's an ecosystem wide consideration that should be brought up and agreed upon. Given how long that code has been in the base and in use in various projects, I had assumed it to have been agreed upon in its usage, but if that's not the case where better place to bring it up?

> What complexity exactly goes down with a child class that simply copies the code previously present in its parent?

I would pay close attention to which children are inheriting from which interfaces. This isn't the place to forward arguments analyzing the complexity, the place to do that would be in a discussion of whether to discontinue the other recording interfaces.

CodyCBakerPhD:
> **But the important thing here is this:** please, next time you review a PR, do not enforce a change and merge it before the author’s review. It’s a collaborative project after all.

Apologies! I figured it was a small enough change that I might as well do it quickly myself. For what it's worth, this is also why having each other **thoroughly review each others lab-specific PRs** would prevent potential issues - for example, note that [line #46 of your class specification for Movshon SpikeGLX](https://github.com/catalystneuro/movshon-lab-to-nwb/blob/main/movshon_lab_to_nwb/movshonnwbconverter.py#L46) should be a `SpikeGLXLFPInterface`.

Issue #153, opened 2021-03-24
DataInterface nesting structure


CodyCBakerPhD:
Pertaining to PR #149, I'd like to open a discussion between maintainers @luiztauffer @alejoe91 @bendichter for everyone's input on evaluating the nesting structure for the data interfaces. @sneakers-the-rat has proposed shifting from a `datainterfaces/formatname.py` approach to `interfaces/extractortype/formatname.py`. The `datainterfaces/formatname.py` approach was inherited originally from the spikeextractors nesting format, which was adopted by roiextractors, and then by us to keep things consistent.

Ultimately, we want to (a) minimize the code necessary to run the repo functional for easier maintainability but hopefully not at the cost of (b) being too hard for users to explore. While (b) could certainly be combated independently through enhanced documentation, it may be worth asking if this proposed methodology might help as well. One downside to the `extractortype` nesting is it duplicates the overall number of files for any format that has multiple extractor types (in cases like Neuroscope, this could generate 3-4x as many files, and the extra lines of import code required on the top).

There could also be a nesting structure of `interfaces/datatype/formatname.py` where `datatype in ["ecephys", "ophys"]` that could be worth throwing in the mix for discussion. Going this route could solve the issue of code file amplification (since no formats are both types, I think) and could also make managing imports from each package more convenient/localized as well. (ecephys imports drawing from only spikeextractors in the `init`, etc).

Anyway, just wanted to see what everyone's thoughts are.

sneakers-the-rat:
Yes, happy to discuss further :)

Here is the structure of the library in the current `master` branch:

```
./nwb_conversion_tools/
├── __init__.py
├── auto_qc.py
├── basedatainterface.py
├── baseimagingextractorinterface.py
├── baselfpextractorinterface.py
├── baserecordingextractorinterface.py
├── basesegmentationextractorinterface.py
├── basesortingextractorinterface.py
├── conversion_tools.py
├── datainterfaces
│   ├── __init__.py
│   ├── blackrockdatainterface.py
│   ├── ceddatainterface.py
│   ├── cellexplorerdatainterface.py
│   ├── imagingextractorinterface.py
│   ├── intandatainterface.py
│   ├── interface_utils
│   │   ├── __init__.py
│   │   └── brpylib.py
│   ├── neuroscopedatainterface.py
│   ├── openephysdatainterface.py
│   ├── phydatainterface.py
│   ├── roiextractordatainterface.py
[...]

PR #191, opened 2021-05-29
reorganize datainterfaces directory

CodyCBakerPhD:
[fix Issue #153](https://github.com/catalystneuro/nwb-conversion-tools/issues/153) 

## Motivation

The original method of appending newly made data interfaces under a nesting structure of simple `nwb-conversion-tools/datainterfaces/*` was becoming inflated, making it difficult to navigate. The new structure nests them by modality in the list `["ecephys", "ophys", "behavior"]`. In the future, planning to extend to `"icephys"` as well.

## How to test the behavior?

The testing infrastructure has either been altered along the new import structure, or is robust to those changes.

## Checklist


CodyCBakerPhD:
The one interesting special case is our most recent interface, Axona, which has a PositionInterface (`behavior` type) attached to the RawIO and thus is still contained in the same file as the `ecephys` RecordingExtractor.

Tests are currently failing because the CI is checking out the main branch rather than the new one - see this discussion which had some tips I've tried but so far no luck: [https://github.com/actions/checkout/issues/15](https://github.com/actions/checkout/issues/15)

CodyCBakerPhD:
a) Of course, there are plenty of other formats that might mirror that. For instance, the `mease-lab-to-nwb` project had some `intan` data that had accelerometer channels plugged in and thus contained in the 'raw' data format. If we hypothetically made a general  IntanAccelerometer interface, should that interface then belong within the `ecephys/intan` nesting or should both this and the Axona be separated under the `behavior/<format name>` (even though that format might also exist in `ecephys`)?

b) It's hard to imagine what movie files wouldn't reflect behavior in some way, but perhaps the MovieDataInterface should belong under a `misc` nesting header?

@bendichter @luiztauffer @sbuergers Any thoughts?

bendichter:
I see the issue, but I say if it's primarily ecephys let's put it in ecephys, even if it can contains behavioral data as well.

luiztauffer:
I also prefer this proposed organization, here are some points:

1. Shouldn’t we wait for #181 and #183 to be merged into master? Otherwise we might have bad conflicts to solve.

2. We should move the modality-specific base interfaces to their modality-specific folders. Examples: `basesegmentationextractorinterface.py` to `datainterfaces/ophys` and `baserecordingextractorinterface` to `datainterfaces/ecephys`. I think `basedatainterface.py` and `nwbconverter.py` can keep living where they are.

With this re-organization of the package I would include some other changes and clean up:

3. The file named `auto_qc.py` doesn’t seem to be used, let’s consider removing it. If there’s any use for the functions in it, we can port them to some utils-like file.

4. We should consider renaming the file named `conversion_tools.py` (maybe `misc.py`) and relocating to it `/utils`.

5. Folder `/schemas` to be renamed `/schema_examples` and the file `metafile.schema.json` to be renamed `metadata_schema.json` and moved there. Probably also this file content needs review, but that’s for later.

6. `/datainterfaces/interface_utils` renamed to `/datainterfaces/utils`

These changes will also requires fixing imports here and there. I could do these changes once we tackle the first point.

CodyCBakerPhD:
@luiztauffer Thanks for taking the time to write all that out, it helps a lot.

1. I'm not too concerned about that, would rather get this through sooner rather than later. And #181 still has a ways to go yet.
2. Agreed, I've made that change now.
3. Git Blame shows the file `auto_qc.py` is almost entirely from @bendichter from 2 years ago - since I have no idea what it's used for, I'll defer to him on what to do with it. I will point out one function there `check_constant_rate` has a slightly different version but the same functionality present in the `conversion_tools.py`.
4. Definitely agree on moving to `utils`, which is now changed. The naming convention is inherited from `spikeextractors`, etc. where it contains all general-purpose util like functions that one might find handy during conversions.
5. Can you make the schema naming changes since you're more familiar with them?
6. Actually, on this one, I wonder if we can go a step better. All the files in this folder pertain specifically to certain interfaces and are in no way general util-like functions. We could instead dismantle the `interface_utils` header and put those individual files under the same nesting as their respective interfaces. A question that remains here is between (i), `interface_utils/brpylib.py` -> `datainterfaces/ecephys/blackrock/brpylib.py`, leaving some of the interfaces under `/datainterfaces/ecephys/` in lone files and others that need supplementary functions in folders paired with those helper functions, or (ii) `interface_utils/brpylib.py` -> `datainterfaces/ecephys/brpylib.py` alongside all the other `ecephys` related files. Either way, this prevents these types of hyperspecific utils from becoming disjointed in the location across the repo. @luiztauffer which would you prefer?

luiztauffer:
Yes, I’m fine with following the SpikeInterface structure and having one extra folder per DataInterface, we can then store DataInterface-specific auxiliary files directly there. This organizes it better when the number of DataInterfaces grow too much, also helps separating them from the base shared modules. So I’d vote for `datainterfaces/ecephys/blackrock/brpylib.py` style

CodyCBakerPhD:
@luiztauffer Check out the latest commit distributing the ophys interfaces in an identical manner - how does it look to you?

luiztauffer (approved):
looks good to me

luiztauffer:
@CodyCBakerPhD you wan to wait for #181 or should we merge it now?

CodyCBakerPhD:
@luiztauffer It hopefully shouldn't cause any conflicts if merges go properly onto master and then from master onto the dev branches, so let's go ahead with it now

Saksham20:
lets merge this?

PR #194, opened 2021-06-14
API docstrings

### LINKED ISSUE #160: API docstrings in readthedocs
@sneakers-the-rat is proposing to use an automatic documentation method to put the docstrings for each function (in each interface) in the documentation so people can search for things that way.

A user wants to look up details on function arguments, but can't or doesn't want to type `nwb_conversion_tools.location.function_name?` in an python environment. 

Add the arguments and docstring details to the readthedocs on the web, so people can use the search functionality and `ctrl-f`.

**Describe alternatives you've considered**
Now, we should note that docstrings for class functions pass to inheritors automatically in current versions of python. Hence, we've intended for only the highest-level docstrings to include the richest most detailed docstrings for things like `run_conversion`, `get_metadata`, etc. but I admit we're probably not where we would like to be on that (also a different issue altogether there). But point being, we wanted the docstrings to really be felt top-down, so that anyone could go look for docstrings on a particular interface by simply using the `nwb_conversion_tools.location.function_name?` method.

One case where this breaks is if the `run_conversion` function has been overridden and additional arguments included, which we mostly did for older projects like the Mease lab to include custom things like the timestamps. Those were so valuable however that we ended up just integrating it into a big refactor of spikeextractors, so it's now in the base behavior regardless. But conceptually, a particular format interface might need some additional arguments at a particular stage, possible requiring user input, so those docstrings would always have to be updated, meaning there would also be deviations from the upstream classes at that point.

@luiztauffer Automatically generating the docstrings for availability on a 'API' subheader of the main docs would allow people to just go to their format directly to learn about the arguments, but it would result in a lot of text duplication since most interfaces will have the exact same . I'm not sure if it's worth the tradeoff, especially since NWB-conversion-tools is supposed to be very user friendly for people perhaps not so interested by such things and it's mostly devs who would want to discover various optional arguments, but I could be wrong about that as well. What's your opinion on that part?

## Checklist

CodyCBakerPhD:
[fix #160](https://github.com/catalystneuro/nwb-conversion-tools/issues/160) 

## Motivation

Allowing docstrings for all functions to be loaded and displayed in an API section of the readthedocs.

## Checklist


CodyCBakerPhD:
Opening as a draft to track progress, needs some slight rework of docstring inheritance since many of them are showing as empty on local `sphinx` compile.

CodyCBakerPhD:
@luiztauffer @bendichter After some tinkering, I was able to get this successfully building on readthedocs.

Entry point: [https://nwb-conversion-tools.readthedocs.io/en/modular_api_docstring/api/nwbconverter.html](https://nwb-conversion-tools.readthedocs.io/en/modular_api_docstring/api/nwbconverter.html)

bendichter (changes requested):
The Recording and Sorting APIs look great! For some reason the Imaging and Segmentation are not being rendered properly.

bendichter:
It looks like you might be missing a space between the ".." and the "automodule" on a few in the Segmentation section

CodyCBakerPhD:
Sorry I didn't catch that before - I  just saw the ecephys ones looked good so I figured there would be no difference or issue with the ophys ones.

Apparently, readthedocs is having issues importing the interfaces, I'll look into it after the brody meeting later this week. Reverting to draft PR.

CodyCBakerPhD:
Moving to a different branch

PR #201, opened 2021-06-24
load metadata from file

luiztauffer:
Convenient util function to load metadata from yaml and json files

The `load_metadata_from_file()` function will be helpful any time a script might load the user-customized metadata from files (which can cbe yaml or json). 
For example, when running the conversion scripts [from command line](https://github.com/catalystneuro/movshon-lab-to-nwb/blob/6b705c24615c098399ca0c48c0535f6564d67f1d/scripts/convert_blackrock.py#L23) that is the only feasible way of passing user metadata. Of course this can be used in any notebook or other script as well.

### DIFF
--- a/nwb_conversion_tools/utils/metadata.py
+++ b/nwb_conversion_tools/utils/metadata.py
@@ -1,0 +1,44 @@
+from pathlib import Path
+import yaml
+import json
+
+
+class NoDatesSafeLoader(yaml.SafeLoader):
+    @classmethod
+    def remove_implicit_resolver(cls, tag_to_remove):
+        """
+        Solution from here: https://stackoverflow.com/a/37958106/11483674
+        Remove implicit resolvers for a particular tag
+
+        Takes care not to modify resolvers in super classes.
+
+        We want to load datetimes as strings, not dates, because we
+        go on to serialise as json which doesn't have the advanced types
+        of yaml, and leads to incompatibilities down the track.
+        """
+        if not 'yaml_implicit_resolvers' in cls.__dict__:
+            cls.yaml_implicit_resolvers = cls.yaml_implicit_resolvers.copy()
+
+        for first_letter, mappings in cls.yaml_implicit_resolvers.items():
+            cls.yaml_implicit_resolvers[first_letter] = [(tag, regexp) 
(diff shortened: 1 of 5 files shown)


bendichter:
So this is to prevent us from parsing dates? Why wouldn't we want to parse dates?

luiztauffer:
we do want to parse them, but as strings. PyYaml reader automatically parses datetime fields as datetime objects

luiztauffer:
@bendichter the `load_metadata_from_file()` function will be helpful any time a script might load the user-customized metadata from files (which can cbe yaml or json), for example [here](https://github.com/catalystneuro/movshon-lab-to-nwb/blob/6b705c24615c098399ca0c48c0535f6564d67f1d/scripts/convert_blackrock.py#L23), calling the converter from command line

CodyCBakerPhD:
@luiztauffer Can you add tests demonstrating usage, and maybe also  complete the auto-filled PR description or something richer to make a case for the structure, use, design, etc.? Since this would count as a new supported feature altogether

Also, are you using this in a less generalized form on any current projects so that you would be able to point to examples that way?

bendichter:
@luiztauffer I'm still stuck on the date issue. I think we should parse them and represent them as datetimes in memory. Then they should appear on the form with a date selector, not a string. Then if we want to save them again into json or yaml, they should be re-serialized into strings.

luiztauffer:
@CodyCBakerPhD you’re right, we should have tests for it, I’ll do them. I’ve included one example usage in the PR description.

@bendichter datetime information should be strings in the metadata objects, that’s how it gets validated by json schemas (check [here](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times)). 
The UI forms already work the way you described: they identify (from the schema definitions) if the string is of a datetime format and creates a datetime selector element, not a string field.

bendichter:
@luiztauffer could you extend jsonschema to allow for datetime objects? https://stackoverflow.com/questions/52210425/validate-datetime-value-using-python-jsonschema

luiztauffer:
so we're good to merge this, right @bendichter ?
it would be important to have it in the coming release

bendichter (approved):

PR #209, opened 2021-07-07
axona style suggestions

bendichter:
in axondadatainterface.py:
* run BLACK
* change name of "generate_position_data" to "get_position_object"
* explicitly define __init__ for AxonaRecordingExtarctor
* generate source schema from explicit AxonaRecordingExtarctor.__init__

in conversion_tools.py, use nwb.processing instead of deprecated nwb.modules

## Motivation

## How to test the behavior?
```
```

## Checklist

- [ ] Is your contribution compliant with our coding style? Read about [PEP8](https://www.python.org/dev/peps/pep-0008/) and [black](https://black.readthedocs.io/en/stable/the_black_code_style.html).
- [ ] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number?

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/axona/axonadatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/axona/axonadatainterface.py
@@ -76,20 +76,10 @@ class AxonaRecordingExtractorInterface(BaseRecordingExtractorInterface):
 
     @classmethod
     def get_source_schema(cls):
+        return get_schema_from_method_signature(cls.__init__)
 
-        source_schema = dict(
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@bendichter For BLACK, I was wondering if there was any desire to implement a `lint` reporter like on `roiextractors` so it shows up as an extra PR condition?

bendichter:
@CodyCBakerPhD we could do that, or we could implement a pre-init hook to just always run BLACK auto-formatting on PRs

sbuergers:
Looks good to me.

PR #211, opened 2021-07-12
get version from release, update requirements

### LINKED ISSUE #210: Standardize releases pipeline
We should standardize our releases pipeline. Right now, Github releases are usually not made prior to PyPI publishing and the github action for automatic PiPY publishing is not working.
I'm opening a PR where we can discuss possible ways of doing that.

luiztauffer:
## Motivation
Fix #210 

Automatically releases a PyPI package for `nwb-conversion-tools` whenever we create a release on Github repository.
This will encourage us to add a text description with the latest updates on each release and will keep the binaries accessible in history if ever needed.

References:
https://youtu.be/U-aIPTS580s 
https://docs.github.com/en/actions/guides/building-and-testing-python#publishing-to-package-registries
https://github.com/pypa/gh-action-pypi-publish

## How to test the behavior?

Make a new release with increased version number.

## Checklist


luiztauffer:
@bendichter for this work, we need to add a secret `PYPI_API_TOKEN` in this repository (I don't have access to that)
The token can be created on the project management page, on PyPI.

luiztauffer:
@alejoe91 @CodyCBakerPhD 
I'm interested in hearing your input on this

CodyCBakerPhD:
@luiztauffer I love it, this would be great if it were working properly. Thanks for looking into it!

I also like pointing the setup `install_requires` to the `requirements.txt` file that has all the version pins rather than just the basic listing of dependencies.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Let's give it a whirl once the secret tokens are in place (I also can't do that myself).

bendichter:
I just created the token and added it to this repo

CodyCBakerPhD:
@luiztauffer Any idea why the tests are failing?

luiztauffer:
I'll check it tomorrow!

alejoe91:
@luiztauffer I think this is a great idea! We have something similar in SI for tags, but I agree 100% that we should document clearly with releases every incremental change. So good for me :)

alejoe91 (approved):

luiztauffer:
This pipeline requires fetching of tags from the repository. But Github Actions uses what is called a "shallow" clone, which does not contain this information. I fixed it with [this solution](https://stackoverflow.com/a/60883893/11483674)

Now the failure is in the tests, due to spikeinterface versions:
`AttributeError: module 'spikeextractors' has no attribute 'AxonaRecordingExtractor'`
this pops up again because now we're installing the versions from `requirements.txt` when running setup.py, which re-installs spikeinterface packages with the fixed versions. These versions are not updated with the fix for the Axona errors

luiztauffer:
@bendichter @CodyCBakerPhD @alejoe91 
I’ve separated the CI for development and production, the only difference being the requirements file used by setup.py. 
This allows us to fix versions for **production** releases and test new **development** environments separately. 

If we ever meet the case where we want the users to install the development version (e.g. if it requires github branch version from specific pakages) it can be controlled locally by setting an ENV variable before installation:
```
NWB_CONVERSION_INSTALL_MODE="development"
```

As you can see, the current **development** version is passing the tests, but the **production** version is not. We should only make releases if **production** CI ends successfully.

luiztauffer:
I added “safe importing” for `opencv` and `ndx-events` also on tests, where they didn’t exist and were failing. These should not be hard requirements for the nwb_conversion_tools. The tests involving these and other optional packages should run on ci-development.

CodyCBakerPhD:
LGTM - I really like separating the production/dev pipelines

CodyCBakerPhD (approved):

CodyCBakerPhD:
Is there a reason to keep the `requirements-dev-old.txt`?

luiztauffer:
> Is there a reason to keep the `requirements-dev-old.txt`?

just didn't want to delete what was being used before, but we can probably get rid of it

CodyCBakerPhD:
> > Is there a reason to keep the `requirements-dev-old.txt`?
> 
> just didn't want to delete what was being used before, but we can probably get rid of it

Yeah, removing should be fine since the PR will maintain a permanent changelog we can always refer back to if needed

PR #212, opened 2021-07-12
minor testing improvements

bendichter:
* offer more explicit error when type mapping cannot be found
* use parameterize to separate out tests for different interfaces

### DIFF
--- a/tests/test_interfaces.py
+++ b/tests/test_interfaces.py
@@ -16,7 +16,6 @@ from nwb_conversion_tools import MovieInterface
 
 @pytest.mark.parametrize("data_interface", interface_list)
 def test_interface_source_schema(data_interface):
-    # check validity of source schema
     schema = data_interface.get_source_schema()
     Draft7Validator.check_schema(schema)
 
--- a/tests/test_interfaces.py
+++ b/tests/test_interfaces.py
@@ -22,7 +22,6 @@ def test_interface_source_schema(data_interface):
 
 @pytest.mark.parametrize("data_interface", interface_list)
 def test_interface_conversion_options_schema(data_interface):
-    # check validity of conversion options schema
     schema = data_interface.get_conversion_options_schema()
     Draft7Validator.check_schema(schema)
 
--- a/tests/test_interfaces.py
+++ b/tests/test_interfaces.py
@@ -5,6 +5,8 @@ from shutil import rmtree
 from pathlib import Path
 from itertools import product
 
+import pytest
+
 import cv2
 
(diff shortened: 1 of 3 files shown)


tests/test_interfaces.py
```diff
@@ -5,22 +5,27 @@
...
-    for data_interface in interface_list:
-        # check validity of source schema
-        schema = data_interface.get_source_schema()
-        Draft7Validator.check_schema(schema)
+@pytest.mark.parametrize("data_interface", interface_list)
+def test_interface_source_schema(data_interface):
+    # check validity of source schema
+    schema = data_interface.get_source_schema()
+    Draft7Validator.check_schema(schema)
+
```
luiztauffer, line 22: @bendichter do we gain any speed by outsourcing the loop to this fixture? Otherwise it just feels less explicit to me

luiztauffer (approved):

tests/test_interfaces.py
CodyCBakerPhD: Even if there isn't a difference in speed, this is easier to read imo

tests/test_interfaces.py
```diff
@@ -5,22 +5,27 @@
...
 
 
-def test_interface_schemas():
-    for data_interface in interface_list:
-        # check validity of source schema
-        schema = data_interface.get_source_schema()
-        Draft7Validator.check_schema(schema)
+@pytest.mark.parametrize("data_interface", interface_list)
+def test_interface_source_schema(data_interface):
+    # check validity of source schema
```
CodyCBakerPhD, line 19:
```suggestion
```
The comments don't really say anything the code around them doesn't

tests/test_interfaces.py
```diff
@@ -5,22 +5,27 @@
...
+    schema = data_interface.get_source_schema()
+    Draft7Validator.check_schema(schema)
+
 
-        # check validity of conversion options schema
-        schema = data_interface.get_conversion_options_schema()
-        Draft7Validator.check_schema(schema)
+@pytest.mark.parametrize("data_interface", interface_list)
+def test_interface_conversion_options_schema(data_interface):
+    # check validity of conversion options schema
```
CodyCBakerPhD, line 26:
```suggestion
```

CodyCBakerPhD (approved):

tests/test_interfaces.py
bendichter: this separates out each data interface into its own test, which helps you identify which ones are failing.

PR #213, opened 2021-07-12
Cleanup

bendichter:
## Motivation

* Generate source schema from __init__ where possible
* remove unused imports

## How to test the behavior?
```
```

## Checklist


nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
```diff
@@ -24,14 +22,7 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
...
     def get_source_schema(cls):
-        """Compile input schema for the RecordingExtractor."""
-        source_schema = get_schema_from_method_signature(
-            class_method=cls.__init__,
-            exclude=['block_index', 'seg_index']
-        )
-        source_schema['properties']['filename']['format'] = 'file'
-        source_schema['properties']['filename']['description'] = 'Path to Blackrock file.'
-        return source_schema
+        return get_schema_from_method_signature(cls.__init__)
```
luiztauffer, line 25: we could replace the upper part with:
```python
	source_schema = super().get_source_schema()
	source_schema['properties']['filename']['format'] = 'file'
    source_schema['properties']['filename']['description'] = 'Path to Blackrock file.'
    return source_schema
```
this addition is relevant for the GUI Forms.

luiztauffer (changes requested):
It makes sense to me to have `get_schema_from_method_signature(cls.__init__)` already in the parent class `get_source_schema()`.
However, most of the class specific modifications in this method are still needed if the class will use the GUI Forms, more specifically the fields `format` and `description`.

nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
bendichter: @luiztauffer Good catch.

How about this syntax:
```python
	source_schema = super().get_source_schema()
	source_schema['properties']['filename'].update(format='file', description='Path to Blackrock file.')
    return source_schema
```

We could also update `get_schema_from_method_signature` to include `format='file'` if the typehint includes the `Path` type.

CodyCBakerPhD: Folder modes can also be `Path` type, but you could automatically determine the difference based on if the suffix is empty. _I.e._, `Path("./Test").suffix` returns `""`

CodyCBakerPhD:
I definitely agree that moving control of the annotation typing to our own class `__init__` is the best way to go here. Just need to find the best way to maintain the custom description fields per data interface, which is more of a thing for Luiz's forms so I'll defer to him on it.

bendichter:
I don't see a great way to automate the description field. The only way I can think of is to parse the docstring, which I think is more trouble than it's worth. I think @luiztauffer's solution to add it at the end of the init as a single line in that function is probably the best approach.

nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
luiztauffer: we couldn't really parse it because this is a definition, comes prior to the user inputs. In fact, the user inputs should be validated by that definition (we're not validating this right now)

CodyCBakerPhD:
@bendichter Between the import cleanups and PR #237, I think the content of the PR is covered by more recent branches. Good to close?

PR #214, opened 2021-07-19
Blackrock interface updated kwargs

Saksham20:
## Motivation

This is related to PR https://github.com/SpikeInterface/spikeextractors/pull/659

To handle the above problem, the same arguments (nsx_override/nev_override) provided by `neo.BlackrockIO` should also be exposed at the blackrockdatainterface arguments. 

## Checklist

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
@@ -33,9 +33,8 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
         source_schema['properties']['filename']['description'] = 'Path to Blackrock file.'
         return source_schema
     
-    def __init__(self, filename: PathType):
-        nsx_to_load = int(str(filename).split('.')[-1][-1])
-        super().__init__(filename=filename, nsx_to_load=nsx_to_load)
+    def __init__(self, filename: PathType, nsx_override: PathType=None, nev_override: PathType=None):
+        super().__init__(filename=filename, nev_override=nev_override,nsx_override=nsx_override)
 
     def get_metadata_schema(self):
         """Compile metadata schema for the RecordingExtractor."""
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Seems fine to me, but I don't work with Blackrock data atm. @luiztauffer Look good to you?

CodyCBakerPhD (approved):

nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
```diff
@@ -33,9 +33,12 @@ def get_source_schema(cls):
...
     
-    def __init__(self, filename: PathType):
-        nsx_to_load = int(str(filename).split('.')[-1][-1])
-        super().__init__(filename=filename, nsx_to_load=nsx_to_load)
+    def __init__(self, filename: PathType, nsx_override: PathType=None):
+        super().__init__(filename=filename,nsx_override=nsx_override)
+        if self.source_data['nsx_override'] is not None:
+            self.data_filename = self.source_data['nsx_override']
+        else:
+            self.data_filename = self.source_data['filename']
```
bendichter, line 41: why do we need this line?

nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
```diff
@@ -33,9 +33,12 @@ def get_source_schema(cls):
         source_schema['properties']['filename']['description'] = 'Path to Blackrock file.'
         return source_schema
     
-    def __init__(self, filename: PathType):
-        nsx_to_load = int(str(filename).split('.')[-1][-1])
-        super().__init__(filename=filename, nsx_to_load=nsx_to_load)
+    def __init__(self, filename: PathType, nsx_override: PathType=None):
+        super().__init__(filename=filename,nsx_override=nsx_override)
+        if self.source_data['nsx_override'] is not None:
+            self.data_filename = self.source_data['nsx_override']
```
bendichter, line 39: why not just pass `None`?

Saksham20: the method get_metadata_Schema() requires to check for the filename of the nsx file to know the version: ns2/3/4/5/6.  In the case when we need only to open the nsx file and not the nev and other header files (as was in my special case which caused me to propose this change), the path needs to be passed only as under the nsx_override argument : the arguments to the `BlackrockIO` have to be :
```python
bk = BlackrockIO(filename='',nsx_override='filepath-to-nsx.ns3')#would be called by spikeextractors internally. 
```

bendichter (approved):
OK, makes sense, but can you please leave a comment in the code that explains this?

bendichter (approved):

PR #216, opened 2021-07-20
API docstrings

### LINKED ISSUE #160: API docstrings in readthedocs

CodyCBakerPhD:
fix #160 

## Motivation

@bendichter Allowing docstrings for all functions to be loaded and displayed in an API section of the readthedocs.

Entry point to built html: [https://nwb-conversion-tools.readthedocs.io/en/modular_api_docstrings_10/api/interfaces.ecephys.html](https://nwb-conversion-tools.readthedocs.io/en/modular_api_docstrings_10/api/interfaces.ecephys.html)

One of the many annoyances in working with readthedocs is its refusal to fully wipe a previous html build from the same branch, hence the many copies of `modular_api_docstring_*` as changes have been requested.

Combined the Recording and Sorting as per the previous PR request, and also added some previously non-existent docstrings for the `BaseSortingExtractorInterface` class.

## Checklist


bendichter:
@CodyCBakerPhD what is the purpose of the TutorialInterfaces? Are they used in some type of tutorial? Do we reference them somewhere in the documentation?

bendichter:
Would it be possible to remove them from this PR and perhaps add them at a later point?

CodyCBakerPhD:
@bendichter The tutorials have been around for a while now: [https://github.com/catalystneuro/nwb-conversion-tools/tree/master/tutorials](https://github.com/catalystneuro/nwb-conversion-tools/tree/master/tutorials)

Can update them to the recent standards in a subsequent PR, though.

PR #218, opened 2021-07-21
DataChunkIterator with specified chunk shape

### LINKED ISSUE #217: allow user to specify chunk shape
This should be a generic function that can be used by multiple different types of DataInterfaces where ever there are large arrays of data (ElectricalSeries, TwoPhotonSeries, ImageSeries, etc.)

Reference implementation from @oruebel:

```python
from hdmf.data_utils import AbstractDataChunkIterator, DataChunk
import numpy as np
​
​
class EdfDataChunkIterator(AbstractDataChunkIterator):
    """
    DataChunkIterator for Edf data
    """
    def __init__(self, edf, num_timesteps, num_channels, buffer_size=1000):
        """
        :param edf: The edf file object to read from
        :param num_timesteps: The total number of timesteps
        :param num_channels: The total number of recording channels
        :param buffer_size: The number of timesteps to read at a time
        """
        self.edf = edf
        self.num_channels = num_channels
        self.num_timesteps = num_timesteps
        self.buffer_size = buffer_size
        self.__curr_timestep = 0
​
    def __iter__(self):
        """Return the iterator object"""
        return self
​
    def __read_block(self, start_index, stop_index):
        arr = np.empty(shape=(stop_index - start_index,  self.num_channels), dtype=self.dtype)
        for chan in range(self.num_channels):
          arr[:, chan] = self.edf.get_data(chan, start_index, stop_index)  # TODO check that this read call is the correct one
        return arr.astype(self.dtype)
​
    def __next__(self):
        """
        Return the next data chunk or raise a StopIteration exception if all chunks have been retrieved.
        :returns: DataChunk object with the data and selection of the current chunk
        :rtype: DataChunk
        """
        if self.__curr_timestep >= self.num_timesteps:
            raise StopIteration
        end_timestep =  (self.__curr_timestep + self.buffer_size)
        if end_timestep > self.num_timesteps:
            end_timestep = self.num_timesteps
        selection = np.s_[self.__curr_timestep: end_timestep, :]
        data = self.__read_block(self.__curr_timestep, end_timestep)
        # in numpy the upper end is exclusive so our next start is at the current end index
        self.__curr_timestep = end_timestep
        # print(data.shape, selection)
        return DataChunk(data=data, selection=selection)
​
    def recommended_chunk_shape(self):
        """
        Recommend the chunk shape for the data array.
        :return: NumPy-style shape tuple describing the recommended shape for the chunks of the target
                 array or None. This may or may not be the same as the shape of the chunks returned in the
                 iteration process.
        """
        return (self.buffer_size, self.num_channels)
​
    def recommended_data_shape(self):
        """
        Recommend the initial shape for the data array.
        This is useful in particular to avoid repeated resized of the target array when reading from
        this data iterator. This should typically be either the final size of the array or the known
        minimal shape of the array.
        :return: NumPy-style shape tuple indicating the recommended initial shape for the target array.
                 This may or may not be the final full shape of the array, i.e., the array is allowed
                 to grow. This should not be None.
        """
        return (self.num_timesteps, self.num_channels)
​
    @property
    def dtype(self):
        """
        Define the data type of the array
        :return: NumPy style dtype or otherwise compliant dtype string
        """
        return np.dtype('float32')
​
    @property
    def maxshape(self):
        """
        Property describing the maximum shape of the data array that is being iterated over
        :return: NumPy-style shape tuple indicating the maxiumum dimensions up to which the dataset may be
                 resized. Axes with None are unlimited.
        """
        return (None, self.num_channels)
​
​
​
​
​
if __name__ == "__main__":
​
  from hdmf.backends.hdf5.h5_utils import H5DataIO
  from datetime import datetime
  from dateutil.tz import tzlocal
  from pynwb import NWBFile, NWBHDF5IO
  from pynwb.base import TimeSeries
  try:
    from mne.io import read_raw_edf
  except ImportError:
    # Dummy class to emulate edf reader
    class read_raw_edf:
      def __init__(self, fname):
        pass
      def get_data(self, chan, start_index, stop_index):
        return np.arange(start_index, stop_index)
​
​
  #region Define functions
  def write_test_file(filename, data):
      """
      Simple helper function to write an NWBFile with a single timeseries containing data
      :param filename: String with the name of the output file
      :param data: The data of the timeseries
      """
      # Create a test NWBfile
      start_time = datetime(2017, 4, 3, 11, tzinfo=tzlocal())
      create_date = datetime(2017, 4, 15, 12, tzinfo=tzlocal())
      nwbfile = NWBFile('demonstrate NWBFile basics',
                        'NWB123',
                        start_time,
                        file_create_date=create_date)
​
      # Create our time series
      test_ts = TimeSeries(name='synthetic_timeseries',
                           data=data,                     # <---------
                           unit='SIunit',
                           rate=1.0,
                           starting_time=0.0)
      nwbfile.add_acquisition(test_ts)
​
      # Write the data to file
      io = NWBHDF5IO(filename, 'w')
      io.write(nwbfile)
      io.close()
​
  # Open file
  fname = 'myfile.edf'
  edf = read_raw_edf(fname)
​
  # Start iterator
  iter_data = EdfDataChunkIterator(
    edf=edf,
    num_timesteps=30010,
    num_channels=100,
    buffer_size=1000)
  data = H5DataIO(
      data=iter_data,
      shuffle=True,
      compression='gzip',
      compression_opts=4)
  write_test_file(filename='edf_iter_write.nwb',data=data)
```

CodyCBakerPhD:
fix #217 

## Motivation

Users have only ever been able to specify the `buffer_mb` size when writing memmap-able electrical series data. This allows them to specify any chunk shape they want.

## How to test the behavior?
With access to any kind of example SpikeGLX data (I wanted to test on actual data here, could probably utilize some random data in CI test, though)

```python
from datetime import datetime
from uuid import uuid4

from pynwb import NWBFile, NWBHDF5IO
from pynwb.ecephys import ElectricalSeries
from hdmf.backends.hdf5.h5_utils import H5DataIO
from spikeextractors import SpikeGLXRecordingExtractor, SubRecordingExtractor
from nwb_conversion_tools.utils.spike_interface import add_devices, add_electrode_groups, add_electrodes
from nwb_conversion_tools.utils.specdatachunkiterator import SpecDataChunkIterator


spikeglx_path = "path_to_spikeglx_g0_t0.imec0.ap.bin"
nwbfile_path = "E:/test_new_iterator.nwb"

nwbfile_kwargs = dict(
    session_description="Auto-generated by NwbRecordingExtractor without description.",
    identifier=str(uuid4()),
    session_start_time=datetime(1970, 1, 1)
)
nwbfile = NWBFile(**nwbfile_kwargs)


recording = SubRecordingExtractor(
    parent_recording=SpikeGLXRecordingExtractor(file_path=spikeglx_path),
    end_frame=10*20000
)
add_devices(recording=recording, nwbfile=nwbfile)
add_electrode_groups(recording=recording, nwbfile=nwbfile)
add_electrodes(recording=recording, nwbfile=nwbfile)

channel_ids = recording.get_channel_ids()
table_ids = [list(nwbfile.electrodes.id[:]).index(id) for id in channel_ids]
electrode_table_region = nwbfile.create_electrode_table_region(
    region=table_ids,
    description="electrode_table_region"
)

chunk_shape = (1000, 4)
ephys_data = SpecDataChunkIterator(
    data=recording.get_traces(return_scaled=False).T,
    full_shape=(recording.get_num_frames(), recording.get_num_channels()),
    chunk_shape=chunk_shape,
    dtype=recording.get_dtype(return_scaled=False)
)

nwbfile.add_acquisition(
    ElectricalSeries(
        data=H5DataIO(ephys_data, chunks=chunk_shape),
        name="test_chunking",
        starting_time=float(recording.frame_to_time(0)),
        rate=float(recording.get_sampling_frequency()),
        conversion=recording.get_channel_gains()[0] * 1e-6,
        electrodes=electrode_table_region
    )
)

with NWBHDF5IO(nwbfile_path, mode="w") as io:
    io.write(nwbfile)
```

## Checklist


CodyCBakerPhD:
@bendichter Code can definitely be cleaned up and improved, but for bare bones functionality this works without error. The main question I have is how we can actually tell that the specific chunking is working as expected outside of print statements inside the DataChunkIterator? When I view the file in HDFView (or any file written by any previous DataChunkIterator method on the same data), it shows up as follows
image
indicating the data storage may not be how it is actually written into the file? I can guarantee these chunks are being written in the example in sizes of 1000 frames and 4 channels since there were very specific errors I debugged where it wrote some chunks but not others, but all chunks being of our intended size.

CodyCBakerPhD:
For example, `nwbfile_out.acquisition["test_chunking"].data.chunks == (3125, 12)` no matter which `DataChunkIterator` or `chunk_shape`, or `buffer_mb` I choose.

CodyCBakerPhD:
@bendichter Nevermind, figured it out. Just have to pass the same chunk shape into the H5DataIO (edited example above).

image

CodyCBakerPhD:
I'll clean up the code, add some proper CI tests, and maybe implement it inside the existing writing functions tomorrow.

nwb_conversion_tools/utils/specdatachunkiterator.py
```diff
@@ -0,0 +1,74 @@
...
+    @property
+    def dtype(self):
+        return np.dtype("int16")
+
+    @property
+    def maxshape(self):
+        return self.full_shape
+
+    def recommended_chunk_shape(self):
+        return None
```
oruebel, line 74: Note, in HDF5DATIO to ``recommended_chunk_shape`` is used to determine the chunk settings of the HDF5 dataset https://github.com/hdmf-dev/hdmf/blob/c5f1bf7e476ce6538726db5fa84441c9ddde9957/src/hdmf/backends/hdf5/h5tools.py#L1336-L1338  When returning ``None`` here this is the same as setting ``chunking=True`` in h5py, i.e., h5py will determine which chunk shape to use. That is, unless you are using this in combination with ``H5DataIO`` to explicitly set a chunking, h5py will pick a chunking for you. To ensure reasonable write performance a key piece to keep in mind is that the chunks you write should ideally line up with the chunks of the HDF5 dataset, i.e., usually you want your ``DataChunk`` to cover a set of consecutive chunks in the HDF5 file (or at least cover one full chunk). Part of the issue is that if your ``DataChunk`` and HDF5 chunks don't line up, that on write you will update each HDF5 chunk multiple times. If the HDF5 chunk is no longer in cache, this means HDF5 will have to: 1) read the chunk, 2) apply any filters (e.g., to decompress the chunk), 3) update/merge the data values with the new data, 4) apply any filters (e.g,. compression), 5) write the chunk back to disk. This means, if the chunk you write don't line up with the chunks on disk you potentially end up with a lot more data movement and I/O than you might expect. Long story short, if you allow the user to specify a the chunk_shape used to write the data you may also want to use a recommended shape that aligns well with the given chunk shape.

CodyCBakerPhD: Fantastic info, thanks! I haven't dug into the HDF5DataIO part of it yet, but it sounds like I should in this case to ensure the code is as streamlined as possible.

nwb_conversion_tools/utils/specdatachunkiterator.py
```diff
@@ -0,0 +1,74 @@
...
+        self.chunk_shape = chunk_shape
+        self.__curr_timestep = 0
+        self.__curr_channel_idx = 0
+        self.__dtype = DataChunk(data=np.empty(shape=(0, 0), dtype=dtype))
+
+    def __iter__(self):
+        """Return the iterator object"""
+        return self
+
+    def __read_block(self, start_time_idx, stop_time_idx, start_ch_idx, stop_ch_idx):
```
bendichter, line 23: "time" and "channel" are specific to ElectricalSeries, and might also apply to other large arrays, but isn't as general as it could be. This might be a good stepping stone, and maybe we should continue with this for now, but ultimately I think we should be able to handle any array with an arbitrary number of dimensions (e.g. RGB video would not work in this case, since it is of shape (time, x, y, [rgb])

nwb_conversion_tools/utils/specdatachunkiterator.py
```diff
@@ -0,0 +1,74 @@
...
+from typing import Iterable
+import numpy as np
+
+from hdmf.data_utils import AbstractDataChunkIterator, DataChunk
+
+
+class SpecDataChunkIterator(AbstractDataChunkIterator):
+    """DataChunkIterator that lets the user specify chunk shapes."""
+
+    def __init__(self, data: Iterable, full_shape: tuple, chunk_shape: tuple, dtype: str):
```
bendichter, line 11: I feel like `full_shape` can be `None` and we can try to recover it with `data.shape`, which would work in the `memmap` case.

bendichter, line 11: Is there any way to include `super().__init__` here?

nwb_conversion_tools/utils/specdatachunkiterator.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __iter__(self):
+        """Return the iterator object"""
+        return self
+
+    def __read_block(self, start_time_idx, stop_time_idx, start_ch_idx, stop_ch_idx):
+        arr = self.data[start_time_idx:stop_time_idx, start_ch_idx:stop_ch_idx]
+        return arr
+
+    def __next__(self):
```
bendichter, line 27: I feel like this function would be using a `for` loop and `yield` instead of `if`s and `return`. I suppose it's just a stylistic difference, but I think the former is more common for these types of things

nwb_conversion_tools/utils/specdatachunkiterator.py
```diff
@@ -0,0 +1,74 @@
...
+        """
```
bendichter, line 28: Would it be possible to generalize this to n-d arrays?

nwb_conversion_tools/utils/specdatachunkiterator.py
```diff
@@ -0,0 +1,74 @@
...
+        minimal shape of the array.
+        :return: NumPy-style shape tuple indicating the recommended initial shape for the target array.
+                 This may or may not be the final full shape of the array, i.e., the array is allowed
+                 to grow. This should not be None.
+        """
+        return (self.full_shape[0], self.full_shape[1])
+
+    @property
+    def dtype(self):
+        return np.dtype("int16")
```
bendichter, line 67: It seems odd to enforce this as int16. Wouldn't it be more intuitive to just carry the type from the `self.data` attribute, and allow an override as an `__init__` arg?

nwb_conversion_tools/utils/specdatachunkiterator.py
oruebel: I don't think ``super().__init__()`` will work here.``AbstractDataChunkIterator`` is an abstract class so calling ``super().__init__()`` should produce a ``TypeError: Can't instantiate abstract class ...`

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: The yield method is typically used when explicitly defining a data generator function, which would absolutely still be the case here if we extended coverage of this to data that is already in memory as an array (the current version assumes a well-functioning memmap).

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Oh yeah, this part was definitely not generalized. I temporarily hard coded it for the data I was testing on because it was being very particular within `hdmf` how the `dtype` object was being set.

CodyCBakerPhD:
I'll think on some ways to generalize to both

- N-dimensional data
- Condition dependent data extraction based on the data storage, i.e., memmap or array
- Performance testing to ensure minimal operations

nwb_conversion_tools/utils/specdatachunkiterator.py
bendichter: What do you mean?

nwb_conversion_tools/utils/specdatachunkiterator.py
bendichter: Maybe we can just have the user set `chunk_shape` and `n_chunks`, where `n_chunks` is the number of chunks read and written on each iteration. This would lock the user to reading/writing in a pattern that fits perfectly into the chunking pattern. Another option is to have the user set `buffer_mb` and automatically determine `n_chunks` from that parameter, approximating the desired `buffer_mb` with multiples of chunks, rounding down. The read/write pattern here would still follow the chunking pattern.

oruebel: You can have a look at how h5py does the autochunking here https://github.com/h5py/h5py/blob/master/h5py/_hl/filters.py#L334-L389 to see how h5py guesses chunk sizes. For you the logic should be much simpler since you'd want the chunk-shape for I/O to be a multiple of ``self.chunk_shape``. I.e., the you can probably just iteratively multiply/divide the dimensions of ``self.chunk_shape`` by a factor of 2 until you reach the desired size. Something like this might work:

```
def recommended_chunk_shape(self):
    def get_chunksize(chunks, typesize):
        # Helper function to determine the size in bytes of a chunk
        return np.product(chunks)*typesize    
    # Start with recommending self.chunks as the chunk size
    chunks = [c for c in self.chunks] 
    typesize = self.dtype.itemsize
    target_max_size = h5py.filters.CHUNK_MAX
    target_min_size = h5py.filters.CHUNK_MIN

    # I/O chunks are too large so we need to shrink them
    if get_chunksize(chunks, typesize) > target_max_size: 
         iter_idx = 0
         while (get_chunksize(chunks, typesize) > target_max_size) and  np.product(chunks) > 2:
              chunks[iter_idx%ndims] = np.ceil(chunks[iter_idx%ndims] / 2.0)
              iter_idx += 1
    # I/O chunks are too small so we should grow them
    elif get_chunksize(chunks, typesize) < target_min_size:
         iter_idx = 0
         while (get_chunksize(chunks, typesize) < target_min_size):
              chunks[iter_idx%ndims] = np.ceil(chunks[iter_idx%ndims] * 2.0)
              iter_idx += 1
  
    return tuple(int(x) for x in chunks)
```

Now, a user can always overwrite this behavior by wrapping your iterator in a ``H5DataIO`` object and setting the chunking option explicitly, e.g., if they want to set a chunking that better aligns with typical read patterns rather than the write.

CodyCBakerPhD: The `buffer_mb` you're suggesting would definitely be more intuitive for the user to set, I'll see if I can get the data sizing calculation down.

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: It wouldn't, that's just how we've used it for this purpose in the past. As I understand though we can't mess too much with the expected return structure of the `__next__` function, but we can probably control how it pulls the chunk from the original data source in read_blocks. I'll follow it up the chain of abstraction to confirm

nwb_conversion_tools/utils/specdatachunkiterator.py
oruebel: I agree, this should be the ``dtype`` of the actual data. Using a fixed dtype here will likely not just cause type conversion of the data to int16 but will raise an error on I/O because the data you want to write doesn't match the specified dtype of the h5py dataset.

nwb_conversion_tools/utils/specdatachunkiterator.py
```diff
@@ -0,0 +1,74 @@
...
+class SpecDataChunkIterator(AbstractDataChunkIterator):
+    """DataChunkIterator that lets the user specify chunk shapes."""
+
+    def __init__(self, data: Iterable, full_shape: tuple, chunk_shape: tuple, dtype: str):
+        self.data = data
+        self.full_shape = full_shape
+        self.chunk_shape = chunk_shape
+        self.__curr_timestep = 0
+        self.__curr_channel_idx = 0
+        self.__dtype = DataChunk(data=np.empty(shape=(0, 0), dtype=dtype))
```
oruebel, line 17: ``self.__dtype`` does not appear to be used. Also should this not be ``self.__dtype=np.dtype``?

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Fixed now

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Fixed now

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Can confirm, attempting to use a generator in the `__next__` function in place of the manual iterating with the `raise StopIteration` results in the usual long `hdmf` traceback terminating in

```
  File "D:\Users\Raven\anaconda3\envs\base2\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 839, in __exhaust_dcis
    if self.__write_chunk__(dset, data):

  File "D:\Users\Raven\anaconda3\envs\base2\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 1368, in __write_chunk__
    if isinstance(chunk_i.selection, tuple):

AttributeError: 'generator' object has no attribute 'selection'
```

oruebel: From the error it seems that your generator may not have returned a ``DataChunk``. In order to write the data HDMF needs to know both: 1) the data values to write and 2) the location to write to. That is what the``DataChunk``class does, it stores a chunk of data and a numpy selection to describe where the chunk belongs in the target array.

CodyCBakerPhD: This is the code for the attempt at using a data generator
```
    for curr_timestep in range(0, full_shape[0], chunk_shape[0]):
        for curr_channel_idx in range(0, full_shape[1], chunk_shape[1]):
            selection = np.s_[
                curr_timestep:(curr_timestep+chunk_shape[0]),
                curr_channel_idx:(curr_channel_idx+chunk_shape[1])
            ]
            data = data_in[selection]
            yield DataChunk(data=data, selection=selection)
```
The idea being that at each iteration it `yields` that particular `DataChunk`, but the bounds of the selector are much easier to specify within the `for` loops.

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: One thing I'm finding with this implementation of the chunk shape inference is overcorrection along the axes if one dimension is at a different magnitude than the other.

E.g., if writing very few long time series, like `full_shape=(2e9, 400)`, the chunk_shape will be pushed to `(x, 1)` where `x` is the number of time frames that can fit into the chunk, whereas we might instead want it contained to `(x/n, n)` where n is some minimal number of series to always chunk together.

On the flip side, if it's a short-ish time series (and the iterations complete with high probability within 3-4 steps), I find that it under-chunks the second dimension by only dividing them by at most 1/4.

I guess in the end it really comes down to compiling all the reasons we want full control of the chunk shape and trying to figure out what is most optimal in the end. As I understand, the compression works on a chunk-to-chunk basis, so different shapes can produce different extents of that. And if we're loading data in for visualizations like in the widgets, that's also sensitive to how the data you're reading relates to the shapes of the stored chunks?

I think I remember hearing that the `ros3` streaming was also done on a chunk to chunk basis? If so that's something that would definitely benefit from a cost-benefit using average download speeds.

nwb_conversion_tools/utils/specdatachunkiterator.py
oruebel: I don't think you can have a yield here. Keep in mind, you are inside an iterator that itself iterates over the chunks, i.e., ``__next__`` is already being called inside a loop that stops iterating once you raise ``StopIteration``. I.e.,  in ``HDF5IO`` you have a loop of the form:

```
for chunk in mySpecDataChunkIterator:
    myHDF5Dataset[chunk.selection] = chunk.data
```

It looks a bit more complicated in ``HDF5IO`` because it has to manage a queue of possible multiple DataChunkIterators that need to be exhausted (see
https://github.com/hdmf-dev/hdmf/blob/c5f1bf7e476ce6538726db5fa84441c9ddde9957/src/hdmf/backends/hdf5/h5tools.py#L840-L848), but from the perspective of your ``SpecDataChunkIterator`` that is effectively what is happening. As such, the outer loop is outside of your direct control, since it is someone else (here ``HDF5IO``) that iterators over all elements (i.e., here ``DataChunk`` that you can produce). Placing a loop over all chunks inside ``__next__``, hence, means you now would have a nested loop iterating over all chunks inside a loop that tries to iterate over all chunks. Generally, using ``yield`` inside ``__next__`` does usually not make sense, because you now have an iterator that itself produces an iterator. This discussion seems related to this and may help: https://stackoverflow.com/questions/37929956/why-does-a-yield-from-inside-next-return-generator-object

Because of this, your loop counters (here ``curr_timestep`` and ``curr_channel``) need to be attributes on the  ``SpecDataChunkIterator`` class so you can keep track of the state of your loop, and then you need to update them inside your ``__next__``, just like you have it in the code right now 👍 

BTW, this also means if you don't reset ``self.__curr_timestep`` and ``self.__curr_channel_idx`` in ``__iter__`` or before you raise``StopIterator``, then the iterator is a on-time iterator only, i.e., if you want to reuse the same iterator again in another loop it will always raise StopIteration immediately. In this context here, I don't think this is an issue because the iterator is intended to be exhausted only once during write and not used again, I just figured I mention it just in case.

oruebel: The loop the way you have it in your example I think shouldwork if you placed it inside the ``__iter__`` function instead and leave ``__next__`` empty. Something like:

```
class SpecDataChunkIterator(AbstractDataChunkIterator):

   def __iter__(self):
        for curr_timestep in range(0, full_shape[0], chunk_shape[0]):
            for curr_channel_idx in range(0, full_shape[1], chunk_shape[1]):
                selection = np.s_[
                   curr_timestep:(curr_timestep+chunk_shape[0]),
                   curr_channel_idx:(curr_channel_idx+chunk_shape[1])
                ]
               data = data_in[selection]
               yield DataChunk(data=data, selection=selection)
```

CodyCBakerPhD: Great info, thanks for the details - I wasn't able to get it working by exclusively putting it in the `__iter__` (new issue was the call to next returned the `SpecDataChunkIterator` itself), but I was eventually able to get it to work by setting up the data generator as a function outside the class, making an internal attribute of the class equal to this, and having the `__next__` return the `next()` of that attribute.

Workflow can improve, I'm sure, but it is at least working. Makes the iteration bounds a lot easier to manage.

oruebel: Honestly, I think the original approach you used, where you updated of the loop variables inside ``__next__`` and keeping everything inside ``__next__`` seems like a cleaner design to me.

nwb_conversion_tools/utils/recordingextractordatachunkiterator.py
```diff
@@ -0,0 +1,35 @@
...
+        self.recording_extractor = recording_extractor
+        super().__init__(chunk_shape=chunk_shape, buffer_mb=buffer_mb)
+
+    def _get_data(self, selection: Tuple[slice]) -> Iterable:
+        return self.recording_extractor.get_traces(
+            channel_ids=list(range(selection[1], selection[1] + self.chunk_shape[1])),
+            start_frame=selection[0],
+            end_frame=selection[0] + self.chunk_shape[0]
+        )
+
```
Saksham20, line 20:
```suggestion
    def _get_data(self, selection: Tuple[slice]) -> Iterable:
        channel_ids = self.recording_extractor.get_channel_ids()[slice[1]]
        frame_ids = range(self.recording_get_num_frames())[slice[0]]
        return self.recording_extractor.get_traces(
            channel_ids=channel_ids ,
            start_frame=frame_ids[0],
            end_frame=frame_ids[1]
        )
```
@CodyCBakerPhD `selection[1]` and `selection[0]` would a a slice object; I'm not sure if a list command on a slice can be called. And the channel_ids values could be arbitrary and not necessary a range(num_channels)

Saksham20 (changes requested):

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+from itertools import product
+
+
+class GenericDataChunkIterator(AbstractDataChunkIterator):
+    """DataChunkIterator that lets the user specify chunk shapes."""
+    
+    def _get_chunksize_mb(self, chunk_shape: Iterable, typesize: int) -> float:
+        return np.product(chunk_shape) * typesize / 1e6
+
+    def _set_chunk_shapes(self, chunk_shape: Iterable, typesize: int, buffer_mb: float) -> tuple:
```
bendichter, line 16:
```suggestion
    def _set_chunk_shape(self, chunk_shape: Iterable, typesize: int, buffer_mb: float) -> tuple:
```

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+        n_dim = len(chunk_shape)
+        iter_idx = 0
+        while (self._get_chunksize_mb(chunk_shape, typesize) > buffer_mb) and np.product(chunk_shape) > 2:
+            dim = iter_idx % n_dim
+            chunk_shape[dim] = np.ceil(chunk_shape[dim] / 2.0)
+            iter_idx += 1
+        return tuple(int(x) for x in chunk_shape)
```
bendichter, line 23: First, could you add a docstring explaining our approach for setting chunk shape. Second, I think it might be possible to solve this in closed form.

bendichter:
There are two things that I think might be conflated here, and I think the room of the problem is our use of the word "chunk" for both. The first is the "chunking" that HDF5 uses to organize datasets into separate pieces on disk. They use the word "chunk" here, so we should too. The second issue is how we are reading the data from the source file in pieces. I'll call this "read-chunk." As @oruebel pointed out, there is an important relationship between chunk and read-chunk. Specifically, if the read-chunk is a different shape than the chunk, it might result in a large amount of extra I/O steps, as data is repeatedly compressed and decompressed. In order to resolve this, it is beneficial to have your chunks be perfectly tile-able within your read-chunks. 

HDF5 recommends < 1 MB chunks: https://www.hdfgroup.org/2017/05/hdf5-data-compression-demystified-2-performance-tuning/#:~:text=Therefore%2C%20if%20data%20in%20the,settings%20will%20not%20be%20penalized.

I think with read-chunks we should be able to get away with much larger than that, like several GB.

Would it be possible to set these independently in the init?

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(self, chunk_shape: tuple = (), buffer_mb: float = 20.):
+        """
+        Initialize the SpecDataChunkIterator object with specified chunk parameters.
+
+        data : Iterable
+            The data to be chunked. Recommended to be a np.memmap at this point in time.
+        chunk_shape : tuple, optional
```
bendichter, line 31: should this be removed?

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+            chunk_shape[dim] = np.ceil(chunk_shape[dim] / 2.0)
+            iter_idx += 1
+        return tuple(int(x) for x in chunk_shape)
+
+    def __init__(self, chunk_shape: tuple = (), buffer_mb: float = 20.):
+        """
+        Initialize the SpecDataChunkIterator object with specified chunk parameters.
+
+        data : Iterable
+            The data to be chunked. Recommended to be a np.memmap at this point in time.
```
bendichter, line 30: should this be removed?

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Should now be generalized, but those other specific child classes still need to be made and tested.

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Done!

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+        return DataChunk(data=data, selection=selection)
+
+    @property
+    def dtype(self) -> np.dtype:
+        return self._dtype
+    
+    @abstractmethod
+    @dtype.setter
+    def dtype(self, value):
+        """Retrieve the dtype of the data using absolute minimal I/O."""
```
bendichter, line 82: A "setter" does not "retrieve"

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Basic logic for this should be done, but needs testing.

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+    def _set_chunk_shapes(self, chunk_shape: Iterable, typesize: int, buffer_mb: float) -> tuple:
+        n_dim = len(chunk_shape)
+        iter_idx = 0
+        while (self._get_chunksize_mb(chunk_shape, typesize) > buffer_mb) and np.product(chunk_shape) > 2:
+            dim = iter_idx % n_dim
+            chunk_shape[dim] = np.ceil(chunk_shape[dim] / 2.0)
+            iter_idx += 1
+        return tuple(int(x) for x in chunk_shape)
+
+    def __init__(self, chunk_shape: tuple = (), buffer_mb: float = 20.):
```
bendichter, line 25:
```suggestion
    def __init__(self, chunk_shape: tuple = None, buffer_mb: float = 20.):
```

I generally use `None` for default values. I suppose since a tuple is non-mutable it's not necessarily a problem, but it is a bit of an unusual choice. Is there a particular reason you are using the empty tuple as a default here?

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+            Defaults to 20 MB.
+        """
+        assert buffer_mb > 0 and buffer_mb < psutil.virtual_memory().available / 1e6, \
+            f"Not enough memory in system handle buffer_mb of {buffer_mb}!"
+        self.data = self._get_data()
+        self.full_shape = self._get_shape()
+        self.__dtype = self._get_dtype()
+        if chunk_shape == ():
+            self.chunk_shape = self._set_chunk_shape(
+                chunk_shape=list(self.full_shape),
```
bendichter, line 44: why does `chunk_shape=full_shape`?

nwb_conversion_tools/utils/genericdatachunkiterator.py
CodyCBakerPhD: Since it's optional you're right, `None` would be better.

CodyCBakerPhD: Fixed!

nwb_conversion_tools/utils/genericdatachunkiterator.py
CodyCBakerPhD: More appropriately named to "initial_shape" since that is the role it plays in the iterative approach. Still working on analytic solution, will update docstrings when it's working better.

nwb_conversion_tools/utils/genericdatachunkiterator.py
CodyCBakerPhD: The property decorators were making an unnecessarily complicated workflow in the `__init__`, simplified now and should also make it clearer to child classes what those functions should do (_i.e._, retrieve the `dtype` and `maxshape` from the data objects they are operating on).

nwb_conversion_tools/utils/genericdatachunkiterator.py
CodyCBakerPhD: Fixed!

nwb_conversion_tools/utils/recordingextractordatachunkiterator.py
CodyCBakerPhD: For the issue of extracting data from the slice object, yes that has now been fixed.

For the issue of `channel_ids`, that's a good catch and is fixed now.

For the issue of frame indices, all the `get_traces()` methods I see utilize direct slicing of frame values starting from `0` to `num_frames`. In `spikeextractors`, I think any `RecordingExtractor` object is guaranteed to have the frame indices in this fashion - what can and should be changed in the case of mismatch is the `timestamp` associated with each frame (and used in any `frame_to_time` conversion), which is not needed for trace extraction.

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -0,0 +1,93 @@
...
+        """Retrieve the dtype of the data using absolute minimal I/O."""
+        raise NotImplementedError("The setter for the internal data type has not been built for this DataChunkIterator!")
+
+    @property
+    def maxshape(self) -> tuple:
+        return self._maxshape
+    
+    @abstractmethod
+    @maxshape.setter
+    def maxshape(self):
```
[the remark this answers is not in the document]
CodyCBakerPhD: Doing away with property decorators since the 'setters' aren't intended to be used in `self.attribute` = value fashion.

nwb_conversion_tools/utils/recordingextractordatachunkiterator.py
```diff
@@ -0,0 +1,35 @@
...
+    @dtype.setter
+    def dtype(self, value):
+        self._dtype = self.recording_extractor.get_dtype(return_scaled=False)
+
+    @property
+    def maxshape(self) -> tuple:
+        return self._maxshape
+
+    @maxshape.setter
+    def maxshape(self):
```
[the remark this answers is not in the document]
CodyCBakerPhD: Doing away with property decorators since the 'setters' aren't intended to be used in `self.attribute` = value fashion.

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Agreed - a lot changed in going to N-dimensions, though and the `__next__` is now back to doing what it more or less originally did but over generalized loops. The particular iteration of that loop still comes from an internal generator attribute (`self.chunk_idx_generator`), but it's also defined in a single line at the `__init__` stage so not that messy.

nwb_conversion_tools/utils/genericdatachunkiterator.py
CodyCBakerPhD: Yep, fixed.

CodyCBakerPhD:
Progress report:

1. The automated chunk and buffer shape inference is now non-iterative (closed-form) and documented. It works simply by finding the largest multiple (_k_) of data size ratios relative to the smallest dimension (_v_) that is less than the specified threshold, clamped of course by the actual size of the data. This provides tighter estimation to the provided threshold than does the exponential weighting of the [h5py auto-chunker](https://github.com/h5py/h5py/blob/266ffae7a09ea984a24e0445d05d67b960c14d4d/h5py/_hl/filters.py#L338-L389).

2. Implementing with RecordingExtractors, specifically, the outline provided by Saksham does well and simply overrides the relevant commands for accessing the slice of the data directly from the RecordingExtractor object, which bypasses any requirement on the type of data source, _i.e._, it will work on `memmap`s `or array`s in precisely the same way.

3. Further testing the implementation in `write_recording` (keeping the old versions as an extra optional argument), in conjunction with speed testing, reveals these current speeds (iteration option no longer applies here)

- Basic: ~40 MB/shttps://assets.grammarly.com/emoji/v1/1f937.svg
- No compression: ~58 MB/s

which is up significantly from the old `array` iteration (very slow), but down a bit from the vanilla `DataChunkIterator` from `hdmf` (with equivalent buffer values). I'm investigating now where those slow downs might be: for compression, I can reason it being slower because it is likely, in fact, applying much more compression because our chunk shapes are in practice much larger than those auto-set by h5py (see #1 above). But based on the 'Basic' write speed, that also means our bigger chunk shapes (sitting right at about 1 MB in size) might just be slower to write?

CodyCBakerPhD:
Main to-do's before calling it done are

1. Comparative testing speed on a full file of actual data
2. Full regiment of tests ensuring each part of iterator is behaving as expected

CodyCBakerPhD:
OK, performance profiling complete. Here is a data table of the results.

- Speed is in units MB/s, the value next to the compression type (`None` or `gzip`) is the compression ratio.
- `v1` is the `hdmf` `DataChunkIterator`, `v2` is the new one from this PR.
- The `buffer_gb` had to be carefully tuned to ranges on the stubbed data that were informative for both, since `v1` maxes out very quickly since it can't buffer over channels. In practice, the `v2` could be pushed well beyond these comparison limits regardless of the data shape.
- Though the `buffer_gb` for `v2` can, for standard data shapes and experiment types well exceed it, it's worth noting and remembering that the chunking/buffering method will scale similarly for an _extremely_ long recording with relatively few channels. Notwithstanding manual override of chunk and buffer shapes, which the user can control with `v2` but not `v1`.

NewDataChunk_FullComparison

Full code used to test locally on a 36 GB SpikeGLX file is below

```
import numpy as np
import pandas as pd

from nwb_conversion_tools.utils.writerecordingprofiler import WriteRecordingProfiler
from spikeextractors import SpikeGLXRecordingExtractor

recording = SpikeGLXRecordingExtractor(
    file_path=".../LR_210527_g0_t0.imec0.ap.bin"
)

mb_threshold = 2000

max_v1_buffer_gb = mb_threshold / 1e3 / recording.get_num_channels()

results = dict()
run_number = 0
total_runs = 2*5*5+2*5
for compression in [None, "gzip"]:
    for chunk_mb in np.linspace(start=0.1, stop=1, num=5):
        for buffer_gb in np.linspace(start=max_v1_buffer_gb/4, stop=3 * max_v1_buffer_gb, num=5):
            print(f"{run_number} / {total_runs}")
            write_kwargs = dict(
                iterator_type="v2", iterator_opts=dict(chunk_mb=chunk_mb, buffer_gb=buffer_gb), compression=compression,
            )
            profiler = WriteRecordingProfiler(recording=recording, mb_threshold=mb_threshold, write_kwargs=write_kwargs)
            results.update({
                run_number: dict(
                    args=dict(mb_threshold=mb_threshold, write_kwargs=write_kwargs), results=profiler.profile()
                )
            })
            run_number += 1
for compression in [None, "gzip"]:
    for buffer_gb in np.linspace(start=max_v1_buffer_gb / 10, stop=max_v1_buffer_gb, num=5):
        print(f"{run_number} / {total_runs}")
        write_kwargs = dict(
            iterator_type="v1", iterator_opts=dict(buffer_gb=buffer_gb), compression=compression,
        )
        profiler = WriteRecordingProfiler(recording=recording, mb_threshold=mb_threshold, write_kwargs=write_kwargs)
        results.update({
            run_number: dict(
                args=dict(mb_threshold=mb_threshold, write_kwargs=write_kwargs), results=profiler.profile()
            )
        })
        run_number += 1
        
flatten_results = list()
for run in results:
    flatten_results.append(
        dict(
            iterator_type=results[run]["args"]["write_kwargs"]["iterator_type"],
            speed=results[run]["results"]["speed"]["speed"],
            compression_type=results[run]["args"]["write_kwargs"]["compression"],
            compression_ratio=results[run]["results"]["compression"]["compression_ratio"],
            buffer_gb=results[run]["args"]["write_kwargs"]["iterator_opts"]["buffer_gb"],
            chunk_mb=results[run]["args"]["write_kwargs"]["iterator_opts"].get(
                "chunk_mb", 16*1024*(2**np.log10(2000*1e6/(1024.*1024))) / 1e6
            ),
        )
    )
results_df = pd.DataFrame(flatten_results)
```

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: Done

nwb_conversion_tools/utils/specdatachunkiterator.py
CodyCBakerPhD: A lot has changed from the early pushes of this; the internal structure of the GenericDataChunkIterator now more closely mimics the original Abstract design; chunks are extracted and set clearly in the `__next__`, shaping arguments set and pulled from respective private functions (including the ones used by the `hdmf`/`h5py` writing, like `recommended_chunk_shape`). It still uses an internal attribute that serves as the generating for the iterating permutations in N-dimensions, but I kinda like how that gets set at `__init__`.

bendichter:
me decided to throw error if both size and shape args are given

bendichter (approved):

PR #221, opened 2021-07-24
create abstract data chunk iterator for various downstream data types

Saksham20:
I was thinking of some structural changes to the class `SpecDataChunkIterator`:
- making `SpecDataChunkIterator` as an abstract class
- the child classes would be specific for the underlying object containing the data (could be a Iterable/extractor object etc)
- keeping all the other logic that @CodyCBakerPhD has implemented


CodyCBakerPhD:
@Saksham20  Thanks - I was prototyping with what is essentially the child class of this generic one. Looks good for the most part, but I do think it can actually be even further abstracted especially as I start extending it into N-dimensions.

Thanks for this!

PR #222, opened 2021-07-25
channel ids for dtype bug fix

Saksham20:
There are two small fixes that I'm proposing in this PR for the `add_electrodes()` method of spike_interface:
-  Bug fix: to find the dtype of the electrical series the channel_ids should be one from the list of `extractor.get_channel_ids()` and not default to 0. [Here ](https://github.com/catalystneuro/nwb-conversion-tools/pull/222/commits/93b8a23cc39215898c26e192f58fc0a84874dc1b)
- Enhancement: having the functionality to enter a custom `starting_time` for the `ElectricalSeries`. [Here](https://github.com/catalystneuro/nwb-conversion-tools/pull/222/commits/5d4391c7328a8d32bf31d553410c2ec226f99ac5)

### DIFF
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -656,7 +656,7 @@ def add_electrical_series(
             eseries_kwargs.update(conversion=1e-6)
             eseries_kwargs.update(channel_conversion=channel_conversion)
 
-    trace_dtype = recording.get_traces(channel_ids=[0], end_frame=1).dtype
+    trace_dtype = recording.get_traces(channel_ids=channel_ids[:1], end_frame=1).dtype
     estimated_memory = trace_dtype.itemsize * recording.get_num_channels() * recording.get_num_frames()
     if not iterate and psutil.virtual_memory().available <= estimated_memory:
         warn("iteration was disabled, but not enough memory to load traces! Forcing iterate=True.")
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -697,12 +697,9 @@ def add_electrical_series(
     eseries_kwargs.update(data=H5DataIO(ephys_data, compression=compression, compression_opts=compression_opts))
     if not use_times:
         eseries_kwargs.update(
+            starting_time=float(recording.frame_to_time(0)),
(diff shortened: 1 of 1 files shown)


nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -656,7 +656,7 @@ def add_electrical_series(
             eseries_kwargs.update(conversion=1e-6)
             eseries_kwargs.update(channel_conversion=channel_conversion)
 
-    trace_dtype = recording.get_traces(channel_ids=[0], end_frame=1).dtype
+    trace_dtype = recording.get_traces(channel_ids=channel_ids[:1], end_frame=1).dtype
```
CodyCBakerPhD, line 659:
```suggestion
    trace_dtype = recording.get_traces(channel_ids=channel_ids[0], end_frame=1).dtype
```

You're right, there's no guarantee that the actual index `0` is in the `channel_ids` list. Should just query the zeroth index of the known `channel_ids` list instead.

CodyCBakerPhD (commented):
> * Enhancement: having the functionality to enter a custom `starting_time` for the `ElectricalSeries`. [Here](https://github.com/catalystneuro/nwb-conversion-tools/pull/222/commits/5d4391c7328a8d32bf31d553410c2ec226f99ac5)

I think for this I'd rather have the user change the internal values of the RecordingExtractor time conversion so we don't have a decoupling between the data in the RecordingExtractor object and what gets written to the NWBFile. Does that sound possible?

Saksham20:
> > * Enhancement: having the functionality to enter a custom `starting_time` for the `ElectricalSeries`. [Here](https://github.com/catalystneuro/nwb-conversion-tools/pull/222/commits/5d4391c7328a8d32bf31d553410c2ec226f99ac5)
> 
> I think for this I'd rather have the user change the internal values of the RecordingExtractor time conversion so we don't have a decoupling between the data in the RecordingExtractor object and what gets written to the NWBFile. Does that sound possible?

yes, that works too. In facts that is a better way indeed: using `recording.set_times()` to set custom times (in my case all nans) works just as well.

Saksham20:
@CodyCBakerPhD  have reverted the commit for the optional `starting_times`, should be good to merge

CodyCBakerPhD (approved):

PR #225, opened 2021-08-06
Add gin tests

CodyCBakerPhD:
## Motivation

The GIN repo has been used on `neo` and `spikeextractors` to guarantee successful read (and some write) of the data formats supported there. Since our interfaces, particularly their metadata interaction functions, make use of data format specificities it's a good idea for us to ensure we can properly access those metadata fields from the source files.

## How to test the behavior?

I've included GIN in our main dev workflow (not production) and as such do not affect normal installation requirements (should just be for our own testing benefit). The dev CI will now take a bit longer to run, but well worth the cost.

@bendichter An alternative might be to add an entirely separate workflow just for the GIN, which would put all PRs at a 4/4 check for format, production release (all lazy imports), dev release (all interfaces with all imports loaded), and GIN. Do you have a preference?

## Checklist

### DIFF
--- a/.github/workflows/ci-development.yml
+++ b/.github/workflows/ci-development.yml
@@ -21,6 +21,9 @@ jobs:
           NWB_CONVERSION_INSTALL_MODE: development
       run: |
         pip install .
+        # needed for correct operation of git/git-annex/DataLad
+        git config --global user.email "CI@example.com"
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
With the soft introduction on intan PR #228, I'll rebase from that to include fixes for the other interfaces

PR #230, opened 2021-08-08
Refactored write ephys modules

alejoe91:
## Motivation

Refactor ephys writer to allow integration of SI<=0.13, SI>=0.90, and NEO object to be written

## How to test the behavior?
WIP: we need to add unit tests for the different data types

## Checklist


alejoe91:
@Saksham20 @bendichter @CodyCBakerPhD let me know what you think about this refactoring.

I think it's good to have different class writers for the different objects, as this would allow us to easily extend it in the future.

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -0,0 +1,161 @@
...
+            self.metadata["Ecephys"] = dict()
+
+        if "Device" not in self.metadata["Ecephys"]:
+            self.metadata["Ecephys"]["Device"] = [defaults]
+
+        for dev in self.metadata["Ecephys"]["Device"]:
+            if dev.get("name", defaults["name"]) not in self.nwbfile.devices:
+                self.nwbfile.create_device(**dict(defaults, **dev))
+
+    def add_electrodes(self):
```
bendichter, line 78: what is the difference between `write_*` and `add_*` functions?

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,77 @@
...
+
+def list_get(li: list, idx: int, default):
+    """Safe index retrieval from list."""
+    try:
+        return li[idx]
+    except IndexError:
+        return default
+
+
+def set_dynamic_table_property(
```
bendichter, line 18: this function is copied in basenwbephyswriter.py

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,77 @@
...
+    else:
+        if property_name in dynamic_table:
+            # TODO
+            raise NotImplementedError
+        else:
+            dynamic_table.add_column(name=property_name, description=description, data=values, index=index, table=table)
+
+
+def check_module(nwbfile, name: str, description: str = None):
+    """
```
bendichter, line 58: this function is copied in basenwbephyswriter.py

nwb_conversion_tools/utils/ephys_writer.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    if not isinstance(objects_to_write, list):
+        objects_to_write = [objects_to_write]
+
+    for object_to_write in objects_to_write:
+        writer_class = None
+        try:
+            if isinstance(object_to_write, SI090NwbEphysWriter.supported_types):
+                writer_class = SI090NwbEphysWriter
+        except AssertionError:
```
bendichter, line 40: what's going on here?

nwb_conversion_tools/utils/basenwbephyswriter.py
alejoe91: The `write_*` functions are in charge of organizing the writing of big portions of the NWB file (e.g. `write_recording`). The add functions are helper functions that add small bits of info to existing NWB files instances.

nwb_conversion_tools/utils/ephys_writer.py
alejoe91: @bendichter we select the correct `writer` class based on the type of the input. The `try` is for uninstalled dependencies

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -78,8 +78,16 @@ def add_to_nwb(self):
...
+    def get_num_segments(self):
+        return self.object_to_write.get_num_segments()
+
+    def _get_num_frames(self, segment_index=0):
+        if self.recording is not None:
+            return self.recording.get_num_frames(segment_index=segment_index)
+
+    def _get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_scaled=True, segment_index=0):
+        return self.recording.get_traces(channel_ids=None, start_frame=None, end_frame=None,
+                                         return_scaled=True, segment_index=segment_index).T
```
alejoe91, line 90: @Saksham20 I think it's better to transpose the SI013 and avoid transposing [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/ephys_writer/nwb_conversion_tools/utils/basenwbephyswriter.py#L579) to make it more efficient

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -110,6 +110,22 @@ def reshape_dynamictable(dt, prop_dict, defaults):
             dt.add_column(name, **des_args)
 
 
+def default_return(value):
+    def return_dec(func):
+        def wrapper(**args):
+            try:
+                func(**args)
```
alejoe91, line 117: shouldn't this be:
```
ret = func(**args)
if ret is None:
   ...
```

Otherwise you are running the function twice..

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -68,12 +68,11 @@ def _get_num_frames(self, segment_index=0):
 
     @default_return([])
     def _get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_scaled=True, segment_index=0):
-        if self.recording is not None:
-            return self.recording.get_traces(channel_ids=channel_ids,
-                                             start_frame=start_frame,
-                                             end_frame=end_frame,
-                                             return_scaled=return_scaled,
-                                             segment_index=segment_index)
+        return self.recording.get_traces(channel_ids=channel_ids,
```
alejoe91, line 71: @Saksham20 I don't thin you have to decorate all functions..the `_get_traces` will never return None and so the `_get_times`, `_get_unit_spike_train` etc

The decorator IMO would be useful for 3-4 functions max, so I wonder whether we need it after all...

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -126,10 +126,9 @@ def _get_num_frames(self, segment_index=0):
 
     @default_return([])
     def _get_recording_times(self, segment_index=0):
-        if self.recording is not None:
-            if self.recording._times is None:
-                return np.arange(0, self._get_num_frames()*self._get_sampling_frequency(), self._get_sampling_frequency())
-            return self.recording._times
+        if self.recording._times is None:
```
alejoe91, line 129: even with the decorator, if `self.recording is None` this will fail because it would try to get `self.recording._times`

nwb_conversion_tools/utils/basesinwbephyswriter.py
```diff
@@ -33,8 +33,7 @@ def _get_unit_sampling_frequency(self):
 
     @default_return([])
     def _get_unit_ids(self):
-        if self.sorting is not None:
-            return self.sorting.get_unit_ids()
+        return self.sorting.get_unit_ids()
```
alejoe91, line 36: even with the decorator, if `self.sorting is None` this will fail because it would try to get `self.sorting.get_unit_ids()`

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -121,9 +120,8 @@ def _get_unit_feature_values(self, prop):
 
     @default_return([])
     def _get_unit_spike_train_ids(self, unit_id, start_frame=None, end_frame=None, segment_index=None):
-        if self.sorting is not None:
-            return self.sorting.get_unit_spike_train(unit_id, start_frame=start_frame,
-                                                     end_frame=end_frame, segment_index=segment_index)
+        return self.sorting.get_unit_spike_train(unit_id, start_frame=start_frame,
```
alejoe91, line 123: same here

nwb_conversion_tools/utils/basesinwbephyswriter.py
Saksham20: the decorator also checks if there is an AttributeError with the 'NoneType' in its comments. Basically, if we try to look up an attribute for a None object. It will still return the default return value then. [This Line](https://github.com/catalystneuro/nwb-conversion-tools/pull/230/files/b6823f957472ffe478bd2d64036a8010149d68db#diff-1b7da2ff140e7bc2daa681d1fdcf7458864e4080decec994675cc95d261df5acR123)

nwb_conversion_tools/utils/si090nwbephyswriter.py
Saksham20: the decorator also checks if we're calling some method/attribute on a NoneType (if self.recording/sorting etc are None). Then it also returns the default value also.

nwb_conversion_tools/utils/si013nwbephyswriter.py
Saksham20: the decorator also checks if we're calling some method/attribute on a NoneType (if self.recording/sorting etc are None). Then it also returns the default value also.

nwb_conversion_tools/utils/basesinwbephyswriter.py
Saksham20: the decorator also checks if we're calling some method/attribute on a NoneType (if self.recording/sorting etc are None). Then it also returns the default value also.

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -128,11 +128,19 @@ def _get_unit_spike_train_times(self, unit_id, segment_index=0):
         return self._get_unit_spike_train_ids(unit_id, segment_index)/self._get_unit_sampling_frequency()
 
     def _get_unit_property_names(self):
-        return list(self.sorting._properties.keys())
+        if 'max_channel' not in self.sorting._properties:
+            prop_names = list(self.sorting._properties.keys()).extend('max_channel')
+        else:
+            prop_names = list(self.sorting._properties.keys())
```
alejoe91, line 134: `self.sorting.get_property_keys()` instead of `self.sorting._properties`

Saksham20: yeah, done

nwb_conversion_tools/utils/si090nwbephyswriter.py
alejoe91: I see. But if say `self.sorting` is None, then in my opinion the `_get_unit_spike_train` SHOULD raise an exception, as calling it would be a mistake. No?

CodyCBakerPhD:
> In the latest commits [here](https://github.com/catalystneuro/nwb-conversion-tools/pull/230/files/e11ee247eeea843e549bf1059b56813be85046e6..bc7e423a8935d8c1235dc79d2cac94fa408c55dc), I have added the option of writing a stub of recording/sorting to the `BaseSINwbEphysWriter` themselves so have made corresponding changes to the `BaseRecordingExtractorInterface` also which @bendichter and @CodyCBakerPhD you had developed initially. Can you please review them

Oh, sorry, didn't notice this ping. Will start looking at it soon

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -110,6 +114,40 @@ def _get_unit_waveforms_templates(self, unit_id, mode="mean"):
...
         pass
 
+    def get_default_nwbfile_metadata(self):
+        """
+        Returns structure with defaulted metadata values required for a NWBFile.
+        """
+        metadata = dict(
+            NWBFile=dict(
+                session_description="no description",
+                session_start_time=datetime(1970, 1, 1),
```
bendichter, line 124: Please do not do this

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -110,6 +114,40 @@ def _get_unit_waveforms_templates(self, unit_id, mode="mean"):
...
+                identifier=str(uuid.uuid4()),
+            ),
+            Ecephys=dict(
+                Device=[dict(name="Device", description="Ecephys probe. Automatically generated.")],
+                Electrodes=[],
+                ElectrodeGroup=[],
```
bendichter, line 130: I'd rather not have empty fields

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -110,6 +114,40 @@ def _get_unit_waveforms_templates(self, unit_id, mode="mean"):
...
+                    description="Processed data",
+                    comments="Generated from SpikeInterface::NwbRecordingExtractor",
+                ),
+                ElectricalSeries_lfp=dict(
+                    name="ElectricalSeries_lfp",
+                    description="Processed data - LFP",
+                    comments="Generated from SpikeInterface::NwbRecordingExtractor",
+                ),
+            ),
+            units=[],
```
bendichter, line 147: I'd rather not have empty fields

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -110,6 +114,40 @@ def _get_unit_waveforms_templates(self, unit_id, mode="mean"):
...
+                ElectricalSeries_processed=dict(
+                    name="ElectricalSeries_processed",
+                    description="Processed data",
+                    comments="Generated from SpikeInterface::NwbRecordingExtractor",
+                ),
+                ElectricalSeries_lfp=dict(
+                    name="ElectricalSeries_lfp",
+                    description="Processed data - LFP",
+                    comments="Generated from SpikeInterface::NwbRecordingExtractor",
+                ),
```
bendichter, line 145: I don't think these fields should be present in the metadata unless there is processed or lfp data

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -158,11 +158,11 @@ def default_export_ops():
     return dict(
         use_times=False,
         write_as="raw",
-        es_key="ElectricalSeries",
+        es_key="ElectricalSeries_raw",
```
bendichter, line 161: I prefer "ElectricalSeries"

nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -18,46 +24,24 @@ class BaseSortingExtractorInterface(BaseDataInterface, ABC):
     @classmethod
     def get_source_schema(cls):
         """Compile input schema for the SortingExtractor."""
-        return get_schema_from_method_signature(cls.__init__)
+        return get_schema_from_method_signature(cls.SX.__init__)
```
alejoe91, line 27: why this change?

nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -18,46 +24,24 @@ class BaseSortingExtractorInterface(BaseDataInterface, ABC):
...
-            UnitProperties=dict(
-                type="array",
-                minItems=0,
-                renderForm=False,
-                items={"$ref": "#/properties/Ecephys/properties/definitions/UnitProperties"},
-            ),
-        )
-        # Schema definition for arrays
-        metadata_schema["properties"]["Ecephys"]["properties"]["definitions"] = dict(
-            UnitProperties=dict(
```
alejoe91, line 44: `UnitProperties` need to be in the schema, otherwise many tests are failing

nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -26,12 +27,14 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
     @classmethod
     def get_source_schema(cls):
         """Compile input schema for the RecordingExtractor."""
-        return get_schema_from_method_signature(cls.__init__)
+        return get_schema_from_method_signature(cls.RX.__init__)
```
bendichter, line 30: I think we should keep this as `cls.__init__`

nwb_conversion_tools/utils/basenwbephyswriter.py
Saksham20: yeah right this is a repetition. Reverted this change. Back to the old way of looking for and updating meta in each write method, here https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/6fec4f5ef7268be30bc7e8d78561b8fed70ce56b

nwb_conversion_tools/utils/basenwbephyswriter.py
Saksham20: reverted https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/6fec4f5ef7268be30bc7e8d78561b8fed70ce56b

nwb_conversion_tools/utils/common_writer_tools.py
Saksham20: Yes, its better to have the raw stored as "ElectricalSeries" only. I have kept the default to None instead. In the `write_electrical_series()` method, it defaults to "ElectricalSeries" if `es_key=None`. I think this is a better way to handle this as `es_key` will only be used if the user wishes to create some custom description for the `ElectricalSeries`. Also, if `es_key` is supplied, then its expected (else there is a warning [here line 413](https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/ccb5886aef04a3ae3de6096394417746ad6cf1be#diff-abd8cf3d8732ccf67ab34bf977a3a7991ce438cba646d559d68ce174e69c1cf6R413) )then the user will also need to update the metadata and supply custom descriptions.

nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
Saksham20: I think this is a wrong use of defining unit properties: we are currently ([example here](https://github.com/catalystneuro/nwb-conversion-tools/blob/e64d6f7fdd96ac82d7f5803f3f5a9ed294c9bfa2/nwb_conversion_tools/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py#L31)) defining unit properties along with data in the metadata which should not be the case. Subsequently, the `run_conversion()`, its writing those properties (pulling that data) into the units table. This is a wrong way to add unit properties: these should be added using the `set_unit_properties()` or `set_property()` method of the spikeextractors/spikeinterface in the `__init__`. This would automatically be pulled in the `add_units()` method then. 

Lets fix these changes in the child classes of `BaseSortingExtractorInterface` in another PR.

nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -66,56 +67,49 @@ def run_conversion(
...
-                        sorting_extractor.set_unit_property(
-                            unit_id=unit_id,
-                            property_name=metadata_column["name"],
-                            value=nwbfile.electrode_groups[
-                                self.sorting_extractor.get_unit_property(
-                                    unit_id=unit_id, property_name="electrode_group"
-                                )
-                            ],
-                        )
-        write_sorting(sorting_extractor, property_descriptions=property_descriptions, nwbfile=nwbfile)
```
Saksham20, line 121: I think this is unnecessary, since setting of any properties (`sorting_extractor.set_unit_property()`)should happen at the `__init__` . Once this is done, the `add_units()` method will automatically pull these.

nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -129,51 +121,32 @@ def run_conversion(
...
-            iterator_opts=iterator_opts,
-        )
+            self.subset_recording()
+
+        conversion_opts = default_export_ops()
+        conversion_opts.update(**kwargs)
+        conversion_opt_schema = default_export_ops_schema()
+        validate(instance=conversion_opts, schema=conversion_opt_schema)
+
+        self.writer_class.write_to_nwb(nwbfile, metadata, **conversion_opts)
```
alejoe91, line 146:
```suggestion
        self.writer_class.add_to_nwb(nwbfile, metadata, **conversion_opts)
```
?

nwb_conversion_tools/utils/basesinwbephyswriter.py
```diff
@@ -0,0 +1,87 @@
...
+            self.add_electrical_series(segment_index=segment_index)
+
+    def add_sorting(self):
+        self.add_units()
+
+    def add_waveforms(self):
+        if self.waveforms is not None:
+            self.add_units_waveforms()
+
+    def add_to_nwb(self, nwbfile: pynwb.NWBFile, metadata=None, **kwargs):
```
alejoe91, line 69: I'd rather have the `nwbfile` and `metadata` passed as arguments to the init. Note that the `nwbfile` can be `None` by default in case one only needs the functions to get proeprties etc. With this implementation, it would be impossible to call separately a `add_***` function because the `metadata` and `nwbfile` would not be instantiated.

Saksham20: in the base class [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/584faa8ee7a33187345ae2a85764648aa2b444e0/nwb_conversion_tools/utils/basenwbephyswriter.py#L31) I do instantiate it, but dont require them as args. 
I've moved this to another PR to this branch

Saksham20: @alejoe91 please check #270

Saksham20:
@CodyCBakerPhD 

the tests [for intanrecordingextractor here](https://github.com/catalystneuro/nwb-conversion-tools/blob/ephys_writer/tests/test_gin.py#L119) are failing. The traces written into the nwb file are being corrected with the `unsigned_coercion` value because of which the [test here](https://github.com/catalystneuro/nwb-conversion-tools/blob/ephys_writer/tests/test_gin.py#L189) is failing. 
The **spike_interface.py** from [master](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/utils/spike_interface.py#L640) is not using the `unsigned_coercion` value anywhere, maybe thats why the tests did not pick it up?

Could you please try running the test_gin.py of this branch on your end and check?

Saksham20 (approved):

alejoe91:
@Saksham20 can you merge with master and solve conflicts?

alejoe91:
@Saksham20 can you also try to fix tests? I think you need to add `spikeinterface` as a requirement

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -0,0 +1,46 @@
...
+        channel_idxs = [self.channel_ids.index(ch) for ch in channel_ids]
+        return (
+            self.ephys_writer._get_traces(
+                channel_ids=channel_ids,
+                start_frame=selection[0].start,
+                end_frame=selection[0].stop,
+                return_scaled=self.write_scaled,
+                segment_index=self.segment_index,
+            )
+            + self.unsigned_coercion[channel_idxs]
```
alejoe91, line 39:
```suggestion
            + self.unsigned_coercion[np.array(channel_idxs)]
```

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -0,0 +1,687 @@
...
+    """Return the number of spikes for chosen unit."""
+    ids = np.array(units_table.id[:])
+    indexes = np.where(ids == unit_id)[0]
+    if not len(indexes):
+        raise ValueError(f"{unit_id} is an invalid unit_id. Valid ids: {ids}.")
+    index = indexes[0]
+    if index == 0:
+        return units_table["spike_times_index"].data[index]
+    else:
+        return units_table["spike_times_index"].data[index] - units_table["spike_times_index"].data[index - 1]
```
alejoe91, line 687: should this go in utils?

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -0,0 +1,687 @@
...
+                    continue
+                set_dynamic_table_property(
+                    dynamic_table=self.nwbfile.units,
+                    row_ids=[int(k) for k in unit_ids],
+                    property_name=ft,
+                    values=flatten_vals,
+                    index=spikes_index,
+                )
+
+    def add_units_waveforms(self):
```
alejoe91, line 653:
```suggestion
    def add_unit_waveforms(self):
```

nwb_conversion_tools/utils/ephys_writer.py
```diff
@@ -0,0 +1,99 @@
...
+          (spikeinterface.RecordingExtractor / spikeinterface.SortingExtractor / spikeinterface.WaveformExtractor)
+        - NEO >= 0.10 objects
+          (neo.rawIO / neo.IO)
+
+    Parameters
+    ----------
+    object_to_write
+    nwb_file_path
+    metadata
+    kwargs
```
alejoe91, line 59: @Saksham20 can you add docstrings here? Especially for kwargs

nwb_conversion_tools/utils/ephys_writer.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    conversion_ops = default_export_ops()
+    conversion_ops.update(kwargs)
+    conversion_opt_schema = default_export_ops_schema()
+    validate(instance=conversion_ops, schema=conversion_opt_schema)
+
+    if nwb_file_path is not None:
+        nwb_file_path = Path(nwb_file_path)
+        if nwb_file_path.is_file() and not conversion_ops["overwrite"]:
+            raise FileExistsError(f"{nwb_file_path} is already existing!")
```
alejoe91, line 73: I think we should remove this, as it will raise an exception when attempting to wirte in append mode (https://github.com/catalystneuro/nwb-conversion-tools/pull/230/files#diff-1b9517b2749c854e5b86609db33f420532ae630b3e10cd568f37f5d3ddaf58d3R91-R92)

nwb_conversion_tools/utils/neonwbephyswriter.py
```diff
@@ -0,0 +1,94 @@
...
+        raise NotImplementedError
+
+    def add_electrical_series(self):
+        raise NotImplementedError
+
+    def add_units(self):
+        raise NotImplementedError
+
+    def add_epochs(self):
+        raise NotImplementedError
```
alejoe91, line 94: even if not implemented yet, we should update this based on the correct base methods (e.g. `write_recording` --> `add_recording`)

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @staticmethod
+    def supported_types():
+        assert HAVE_SI_090
+        return (si.BaseRecording, si.BaseSorting, si.BaseEvent, si.WaveformExtractor)
+
+    def get_num_segments(self):
+        return self.object_to_write.get_num_segments()
+
+    @default_return(None)
```
alejoe91, line 80: I would remove all decorators for now. Again, I think that if something is wrong an exception shoudl be raised, and these decorators might catch the exception and make it hard to debug. @bendichter @CodyCBakerPhD opinions?

alejoe91 (commented):
@Saksham20 looks good! I have some last minor comments

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,218 @@
...
+                    return out
+            except Exception as e:
+                if "NoneType" in str(e):
+                    return value
+                else:
+                    return e
+
+        return wrapper
+
+    return return_dec
```
bendichter, line 131: what is the purpose of this? It seems to be silencing all errors. Why would we want that?

CodyCBakerPhD:
@Saksham20 @alejoe91 Quick question, will this write a custom electrodes column of `offsets` even if `all(recording.get_channel_offsets() == 0)`? If so it should not bother doing that in this case

CodyCBakerPhD:
@alejoe91 Also, does the new version of spikeinterface extractors have a SubRecording class like before and does it copy properties from its parent? If not we need to be careful in nwb-conversion-tools, that any time we subset one of those objects, we copy relevant properties over from it's parent. I've run into this problem many times with the old version of spikeextractors

nwb_conversion_tools/utils/common_writer_tools.py
alejoe91: @Saksham20 please remove

alejoe91:
> @alejoe91 Also, does the new version of spikeinterface extractors have a SubRecording class like before and does it copy properties from its parent? If not we need to be careful in nwb-conversion-tools, that any time we subset one of those objects, we copy relevant properties over from it's parent. I've run into this problem many times with the old version of spikeextractors

Yes, we also have a functional way of doing it:
```
recording_sub = recording.frame_slice(start_frame=..., end_frame=...)
```

> @Saksham20 @alejoe91 Quick question, will this write a custom electrodes column of `offsets` even if `all(recording.get_channel_offsets() == 0)`? If so it should not bother doing that in this case

For the new API, offsets and gains are NOT automatically populated with zeros and ones, respectively. This means that for new objects those properties will not be presents and therefore not written

nwb_conversion_tools/utils/basenwbephyswriter.py
Saksham20: yes, we can move this to `common_writer_tools.py`

nwb_conversion_tools/utils/ephys_writer.py
Saksham20: https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/eaa4691d053025d78ca12738f827a6aa31677172

nwb_conversion_tools/utils/si090nwbephyswriter.py
Saksham20: the use of this decorator is to neatly deal with cases:
1. when the method returns a wrong dtype (list, vs array vs none); we can ensure the return value is of a specific dtype only. 
2. if the `self.recording`, `self.sorting` objects are none and the user calls the methods that require it, example calling `self._get_unit_spike_train_ids()` is called when the `object_to_write` is a `self.recording` (so `self.sorting` is None). Yes, calling spiketrainids is invalid in this case, but I think we should not throw an error but just return None.

nwb_conversion_tools/utils/common_writer_tools.py
Saksham20: > what is the purpose of this? It seems to be silencing all errors. Why would we want that?

I have changed it to raise errors instead of returning them. The only error that we need to deal with is when calling a method on a `NoneType` object (which can happen: https://github.com/catalystneuro/nwb-conversion-tools/pull/230#discussion_r744249753

Saksham20: https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/d9ea4597f5a70ac143c96c35b922b4c62f8b2a4f

Saksham20:
> For the new API, offsets and gains are NOT automatically populated with zeros and ones, respectively. This means that for new objects those properties will not be presents and therefore not written

But these are set as default values in the writer (so they will be written as zeros and ones. 
https://github.com/catalystneuro/nwb-conversion-tools/blob/d9ea4597f5a70ac143c96c35b922b4c62f8b2a4f/nwb_conversion_tools/utils/si090nwbephyswriter.py#L111

alejoe91:
> > For the new API, offsets and gains are NOT automatically populated with zeros and ones, respectively. This means that for new objects those properties will not be presents and therefore not written
> 
> But these are set as default values in the writer (so they will be written as zeros and ones.
> 
> https://github.com/catalystneuro/nwb-conversion-tools/blob/d9ea4597f5a70ac143c96c35b922b4c62f8b2a4f/nwb_conversion_tools/utils/si090nwbephyswriter.py#L111

I think we can safely remove these lines: https://github.com/catalystneuro/nwb-conversion-tools/blob/d9ea4597f5a70ac143c96c35b922b4c62f8b2a4f/nwb_conversion_tools/utils/si090nwbephyswriter.py#L117-L126

alejoe91:
@Saksham20 thanks for the changes. Personally I would remove the decorators...what is the advantage of having them?

nwb_conversion_tools/utils/neonwbephyswriter.py
Saksham20: https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/4429fb5dd61d1ba225bcab2cdeb552fdc075534c

Saksham20:
> @Saksham20 thanks for the changes. Personally I would remove the decorators...what is the advantage of having them?

we will have to write if else statements each time to take care of the cases when the methods may fail. Just adds lines of code so I had implemented decorators instead.

bendichter:
> > @Saksham20 thanks for the changes. Personally I would remove the decorators...what is the advantage of having them?
> 
> 
> 
> we will have to write if else statements each time to take care of the cases when the methods may fail. Just adds lines of code so I had implemented decorators instead. 

Why would we do that?

nwb_conversion_tools/utils/si090nwbephyswriter.py
alejoe91: In my opinion we don't need them. They are only for internal methods (e.g. `_get_traces()`) so they are not supposed to be called by external users.

Saksham20:
okay lets not keep decorators as they may over complicate things making it harder to maintain in the future. They are now removed https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/697ee10736716d335aed816346bae5ef652e7c06

alejoe91:
Thanks Saksham, looks ready to merge on my side :) @bendichter @CodyCBakerPhD any "last words"?

CodyCBakerPhD:
LGTM, though internally prepared to deal with expected debugging as the new writer gets used more often. Easier to make those changes once it's into master, at least.

Saksham20:
> LGTM, though internally prepared to deal with expected debugging as the new writer gets used more often. Easier to make those changes once it's into master, at least.

one of the things that needs change is that all the recordingextractordatainterfaces need to call the new si.recordingextractor instead of se.recordingextractor. [Example here](https://github.com/catalystneuro/nwb-conversion-tools/blob/4d282286ec9bb65159dcab6c6e90412e6156369e/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py#L23). That I will create a new PR for once this is merged.

bendichter:
Can we have a look at a few example NWB files created by this new pipeline?

bendichter:
From the [example files](https://catalystneuro.slack.com/files/U01FDLMFN59/F02M9A08XDH/nwbtest_files.zip), `units/spike_times_index` is of type `uint8`. Does that change if we have more spikes? Otherwise we will certainly go above the max of that data type, 255.

alejoe91:
> From the [example files](https://catalystneuro.slack.com/files/U01FDLMFN59/F02M9A08XDH/nwbtest_files.zip), `units/spike_times_index` is of type `uint8`. Does that change if we have more spikes? Otherwise we will certainly go above the max of that data type, 255.

@Saksham20 can you look into that? I agree it should be int64

Saksham20:
> > From the [example files](https://catalystneuro.slack.com/files/U01FDLMFN59/F02M9A08XDH/nwbtest_files.zip), `units/spike_times_index` is of type `uint8`. Does that change if we have more spikes? Otherwise we will certainly go above the max of that data type, 255.
> 
> @Saksham20 can you look into that? I agree it should be int64

Yes for more number of spikes, it changes the dtype of `spike_times_index` to uint32. Check this [nwb file](https://files.slack.com/files-pri/T01EYN14615-F02ML1YJP2Q/download/test_ephys_units.nwb).

I created it like so:
```python
from spikeinterface.core.testing_tools import generate_sorting
from nwb_conversion_tools.utils import export_ecephys_to_nwb

sx = generate_sorting(durations=[10000])
nwbloc = 'test_ephys_units.nwb'
export_ecephys_to_nwb(object_to_write=sx, nwbfile=nwbloc)
```

Saksham20:
@alejoe91 the sample files look good on my end also. I think we are ready to go!

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,209 @@
...
+ArrayType = Union[list, np.ndarray]
+DynamicTableSupportedDtypes = {list: [], np.ndarray: np.array(np.nan), str: "", Real: np.nan}
+
+
+def list_get(li: list, idx: int, default):
+    """Safe index retrieval from list."""
+    try:
+        return li[idx]
+    except IndexError:
+        return default
```
bendichter, line 18: Is this used anywhere?

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,209 @@
...
+        write_scaled=False,
+        compression="gzip",
+        compression_opts=4,
+        skip_unit_properties=[],
+        skip_unit_features=[],
+        skip_electrode_properties=[],
+        unit_property_descriptions=dict(),
+        write_electrical_series=True,
+        overwrite=True,
+    )
```
bendichter, line 169: why is this a function?

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,209 @@
...
+        compression=dict(type="string", enum=["gzip", "lzf"]),
+        compression_opts=dict(type="number", minimun=0, maximum=9),
+        skip_unit_properties=dict(type="array", items=dict(type="string")),
+        skip_unit_features=dict(type="array", items=dict(type="string")),
+        skip_electrode_properties=dict(type="array", items=dict(type="string")),
+        unit_property_descriptions=dict(type="object"),
+        write_electrical_series=dict(type="boolean"),
+        overwrite=dict(type="boolean"),
+    )
+    return schema
```
bendichter, line 195: why is this a function?

nwb_conversion_tools/utils/ephys_writer.py
```diff
@@ -0,0 +1,133 @@
+from distutils.version import StrictVersion
+import pynwb
+from pathlib import Path
+from jsonschema import validate
+
+# import ephys writers
+from .si013nwbephyswriter import SI013NwbEphysWriter
+from .si090nwbephyswriter import SI090NwbEphysWriter
+from .neonwbephyswriter import NEONwbEphysWriter
+from nwb_conversion_tools.nwbconverter import NWBConverter
```
bendichter, line 10: use relative path

nwb_conversion_tools/utils/ephys_writer.py
```diff
@@ -0,0 +1,133 @@
...
+        - spikeinterface version >= 0.90 objects
+          (spikeinterface.RecordingExtractor / spikeinterface.SortingExtractor / spikeinterface.WaveformExtractor)
+        - NEO >= 0.10 objects
+          (neo.rawIO / neo.IO)
+
+    Parameters
+    ----------
+    object_to_write: object
+        spike interface/extractors object
+    nwb_file_path: str
```
bendichter, line 60:
```suggestion


    Parameters
    ----------
    object_to_write: object
        spike interface/extractors object
            Supported objects are:
            - spikeinterface version <= 0.13 extractors
              (spikeextractors.RecordingExtractor / spikeextractors.SortingExtractor)
            - spikeinterface version >= 0.90 objects
              (spikeinterface.RecordingExtractor / spikeinterface.SortingExtractor / spikeinterface.WaveformExtractor)
            - NEO >= 0.10 objects
              (neo.rawIO / neo.IO)
    nwb_file_path: str
```

nwb_conversion_tools/utils/ephys_writer.py
```diff
@@ -0,0 +1,133 @@
...
+        iterator_type ("v2"): v2 is using the custom datachunkiterator for chunking, v1 is using the built in
+            datachunk iterator from hdf5
+        chunk_shape (tuple): Manually defined shape of the chunks. Defaults to None.
+        write_scaled (False): If True, writes the scaled traces (return_scaled=True)
+        compression ("gzip"): Type of compression to use. Valid types are "gzip" and "lzf".
+            Set to None to disable all compression.
+        compression_opts (int): Only applies to compression="gzip". Controls the level of the GZIP.
+        skip_unit_properties (list): names of unit properties to skip writing to nwbfile,
+        skip_unit_features (list): names of unit features (old spikeextractors) to skip,
+        skip_electrode_properties= (list): extractors' channel properties to skip,
```
bendichter, line 86:
```suggestion
        skip_electrode_properties= (list): extractor's channel properties to skip,
```

nwb_conversion_tools/utils/neonwbephyswriter.py
```diff
@@ -0,0 +1,110 @@
...
+    def add_sorting(self):
+        raise NotImplementedError
+
+    def add_waveforms(self):
+        raise NotImplementedError
+
+    def add_epochs(self):
+        raise NotImplementedError
+
+    @abstractmethod
```
bendichter, line 59: why is this abstract?

nwb_conversion_tools/utils/neonwbephyswriter.py
```diff
@@ -0,0 +1,110 @@
...
+    def _make_recording_stub(self):
+        pass
+
+    @abstractmethod
+    def _make_sorting_stub(self):
+        pass
+
+    def _get_sampling_frequency(self):
+        pass
```
bendichter, line 68: should these be `NotImplementedError` instead of `pass`?

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,302 @@
...
+                        tags=epoch_name,
+                    )
+
+    def add_recording(self, segment_index=0):
+        super().add_recording(segment_index=segment_index)
+        if self._conversion_ops["write_electrical_series"]:
+            self.add_epochs()
+
+
+def create_si013_example(seed):
```
bendichter, line 221: Why is this here? Should it be moved to tests?

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -0,0 +1,191 @@
...
+        return self._check_valid_property(prop_values)
+
+    def _get_unit_waveforms_templates(self, unit_id, mode="average"):
+        return self.waveforms.get_template(unit_id, mode=mode)
+
+    def add_epochs(self):
+        return
+
+
+def create_si090_example():
```
bendichter, line 182: should this be here?

requirements-dev.txt
```diff
@@ -11,6 +11,7 @@ psutil==5.8.0
 lxml==4.6.3
 opencv-python==4.5.1.48
 spikeextractors==0.9.8
+spikeinterface@git+https://github.com/SpikeInterface/spikeinterface.git
```
bendichter, line 14: Would it be possible to list a version on pypi?

requirements-rtd.txt
```diff
@@ -10,6 +10,7 @@ jsonschema==3.2.0
 psutil==5.8.0
 lxml==4.6.3
 spikeextractors==0.9.7
+spikeinterface@git+https://github.com/SpikeInterface/spikeinterface.git
```
bendichter, line 13: Would it be possible to list a version on pypi?

requirements.txt
```diff
@@ -9,8 +9,10 @@ PyYAML==5.4
 jsonschema==3.2.0
 psutil==5.8.0
 lxml==4.6.3
-spikeextractors==0.9.7
+spikeextractors
+spikeinterface@git+https://github.com/SpikeInterface/spikeinterface.git
```
bendichter, line 13: Would it be possible to list a version on pypi?

tests/test_si090.py
```diff
@@ -0,0 +1,262 @@
...
+        self.RX, self.SX = create_si090_example()
+        self.RX2, self.SX2 = create_si090_example()
+        self.RX3, self.SX3 = create_si090_example()
+        self.test_dir = tempfile.mkdtemp()
+        self.path1 = self.test_dir + "/test_electrodes1.nwb"
+        self.path2 = self.test_dir + "/test_electrodes2.nwb"
+        self.path3 = self.test_dir + "/test_electrodes3.nwb"
+        self.nwbfile1 = NWBFile("sess desc1", "file id1", datetime.now())
+        self.nwbfile2 = NWBFile("sess desc2", "file id2", datetime.now())
+        self.nwbfile3 = NWBFile("sess desc3", "file id3", datetime.now())
```
bendichter, line 149: It looks like these are unused

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,302 @@
...
+
+    def _get_channel_property_values(self, prop):
+        if prop == "location":
+            return self.recording.get_channel_locations()
+        elif prop == "gain":
+            return self.recording.get_channel_gains()
+        elif prop == "offset":
+            return self.recording.get_channel_offsets()
+        elif prop == "group":
+            if any(["group_name" in self.recording._properties.get(id, "") for id in self.recording.get_channel_ids()]):
```
bendichter, line 127: Could we avoid accessing protected members of classes? `recording._properties` has a leading underscore, which indicates that it should not be accessed directly.

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,209 @@
...
+    name: str
+    description: str | None (optional)
+
+    Returns
+    -------
+    pynwb.module
+    """
+    assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
+    if name in nwbfile.modules:
+        return nwbfile.modules[name]
```
bendichter, line 130:
```suggestion
    if name in nwbfile.processing:
        return nwbfile.processing[name]
```

nwb_conversion_tools/utils/ephys_writer.py
```diff
@@ -0,0 +1,133 @@
...
+    if writer_class is None:
+        raise Exception(f"Could not write object of type {type(object_to_write)}")
+    return writer_class
+
+
+def export_ecephys_to_nwb(
+    object_to_write,
+    nwb_file_path=None,
+    nwbfile=None,
+    metadata=None,
```
bendichter, line 40: not in docstring

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,302 @@
...
+    def _make_sorting_stub(self):
+        max_min_spike_time = max(
+            [min(x) for y in self.sorting.get_unit_ids() for x in [self.sorting.get_unit_spike_train(y)] if any(x)]
+        )
+        self.sorting = se.SubSortingExtractor(self.sorting, start_frame=0, end_frame=1.1 * max_min_spike_time)
+
+    @staticmethod
+    def supported_types():
+        assert HAVE_SI013
+        return (se.RecordingExtractor, se.SortingExtractor)
```
bendichter, line 60:
```suggestion
        return se.RecordingExtractor, se.SortingExtractor
```

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,302 @@
...
+        features3=features3,
+        unit_prop=80,
+        channel_prop=(0, 0),
+        ttls=ttls,
+        epochs_info=((0, 10), (10, 20)),
+        geom=geom,
+        times=times,
+    )
+
+    return (RX, RX2, RX3, SX, SX2, SX3, example_info)
```
bendichter, line 302:
```suggestion
    return RX, RX2, RX3, SX, SX2, SX3, example_info
```

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -0,0 +1,191 @@
...
+    def _make_sorting_stub(self):
+        max_min_spike_time = max(
+            [min(x) for y in self.sorting.get_unit_ids() for x in [self.sorting.get_unit_spike_train(y)] if any(x)]
+        )
+        self.sorting = si.FrameSliceSorting(start_frame=0, end_frame=1.1 * max_min_spike_time)
+
+    @staticmethod
+    def supported_types():
+        assert HAVE_SI_090
+        return (si.BaseRecording, si.BaseSorting, si.BaseEvent, si.WaveformExtractor)
```
bendichter, line 74:
```suggestion
        return si.BaseRecording, si.BaseSorting, si.BaseEvent, si.WaveformExtractor
```

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,302 @@
...
+        property_names = set()
+        for chan_id in self._get_channel_ids():
+            for i in self.recording.get_channel_property_names(chan_id):
+                property_names.add(i)
+        return list(property_names)
+
+    def _fill_missing_property_values(self, ids, prop, get_prop_func):
+        self.dt_column_defaults = {list: [], str: "", Real: np.nan, np.ndarray: np.array([np.nan])}
+        # find the size of ndarray dtype:
+        for id in ids:
```
bendichter, line 83: This shadows the built-in name "id"

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,302 @@
...
+                if isinstance(id_data, np.ndarray):
+                    self.dt_column_defaults.update({np.ndarray: np.nan * np.ones(shape=[1, id_data.shape[1:]])})
+                    break
+                else:
+                    break
+            except:
+                continue
+        # find the channel property dtype:
+        found_property_types = Real
+        for id in ids:
```
bendichter, line 95: This shadows the built-in name "id"

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,302 @@
...
+            try:
+                id_data = get_prop_func(id, prop)
+                if isinstance(id_data, np.ndarray):
+                    self.dt_column_defaults.update({np.ndarray: np.nan * np.ones(shape=[1, id_data.shape[1:]])})
+                    break
+                else:
+                    break
+            except:
+                continue
```
bendichter, line 92: Why do we have these except clauses? Is there a better way to write this?

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -0,0 +1,191 @@
+from distutils.version import StrictVersion
+from typing import Union
```
bendichter, line 2:
```suggestion
```

nwb_conversion_tools/utils/si090nwbephyswriter.py
```diff
@@ -0,0 +1,191 @@
...
+
+import numpy as np
+import pynwb
```
bendichter, line 5:
```suggestion
```

nwb_conversion_tools/utils/neonwbephyswriter.py
Saksham20: yes correct `make_recording_stub()` and `make_sorting_stub()` should be moved to the base class: `BaseNwbEphysWriter`. NeoEphysWriter will them implement this.

nwb_conversion_tools/utils/common_writer_tools.py
Saksham20: so that the same dict can be retrieved each time. It would be possible to change it if it is kept as a constant in the .py file no?

nwb_conversion_tools/utils/neonwbephyswriter.py
Saksham20: I decided to have all the `get` methods return a `None` for now so `pass`. All the `add` methods which manipulate a variable, is a functionality that will be implemented in another PR.

nwb_conversion_tools/utils/si013nwbephyswriter.py
Saksham20: thanks for catching that, yes we can directly use `recording.get_channel_property_names(id)` for this.

nwb_conversion_tools/utils/common_writer_tools.py
Saksham20: right, this has no usages. removed

nwb_conversion_tools/utils/si090nwbephyswriter.py
Saksham20: moved the methods https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/083dafa2a3eae16ac099905a1508e28f3c97bf52

nwb_conversion_tools/utils/si013nwbephyswriter.py
Saksham20: https://github.com/catalystneuro/nwb-conversion-tools/pull/230#discussion_r751235770

nwb_conversion_tools/utils/neonwbephyswriter.py
Saksham20: https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/d7cb6236c547ac89bee3be6d759121b314c464cb

nwb_conversion_tools/utils/si013nwbephyswriter.py
Saksham20: changed as discussed https://github.com/catalystneuro/nwb-conversion-tools/pull/230/commits/c26593605361bc92cafb5305b9096f51e6145d5e

nwb_conversion_tools/utils/common_writer_tools.py
```diff
@@ -0,0 +1,201 @@
...
+        write_scaled=False,
+        compression="gzip",
+        compression_opts=4,
+        skip_unit_properties=[],
+        skip_unit_features=[],
+        skip_electrode_properties=[],
+        unit_property_descriptions=dict(),
+        write_electrical_series=True,
+        overwrite=True,
+    )
```
bendichter, line 161: this should be a global dict. I don't see any reason why it should be a function.

```suggestion
default_export_ops =  dict(
    use_times=False,
    write_as="raw",
    es_key=None,
    buffer_gb=None,
    buffer_shape=None,
    chunk_mb=None,
    iterator_type="v2",
    chunk_shape=None,
    write_scaled=False,
    compression="gzip",
    compression_opts=4,
    skip_unit_properties=[],
    skip_unit_features=[],
    skip_electrode_properties=[],
    unit_property_descriptions=dict(),
    write_electrical_series=True,
    overwrite=True,
)
```

bendichter: Also, the null values should be omitted. I don't see why we should be passing empty lists and dictionaries.

nwb_conversion_tools/utils/si013nwbephyswriter.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def _fill_missing_property_values(self, channel_ids, prop, get_prop_func, get_prop_names_func):
+        self.dt_column_defaults = {list: [], str: "", Real: np.nan, np.ndarray: np.array([np.nan])}
+        # find the size of ndarray dtype:
+        for chan_id in channel_ids:
+            if prop not in get_prop_names_func(chan_id):
+                continue
+            id_data = get_prop_func(chan_id, prop)
+            if isinstance(id_data, np.ndarray):
+                self.dt_column_defaults.update({np.ndarray: np.nan * np.ones(shape=[1, id_data.shape[1:]])})
```
bendichter, line 89:
```suggestion
                self.dt_column_defaults.update({np.ndarray: np.nan * np.ones(shape=[1, *id_data.shape[1:]])})
```

nwb_conversion_tools/utils/common_writer_tools.py
Saksham20: if its a global variable that I import in a script then if its updated, then there is no way to get the unaltered one again in that script right?

```python
from nwb_conversion_tools.utils.common_writer_tools import default_export_ops, default_export_ops_schema

conversion_ops = default_export_ops
 # will have to instead call: conversion_ops = dict(default_export_ops) each time
conversion_ops.update(kwargs)
```
I dont know, from a design point of view, is it better to call dict() each time to create a new instance of the global variable or have the global variable be the output of a function

PR #234, opened 2021-08-16
Units provenance

CodyCBakerPhD:
## Motivation

A common request that has come up from many different labs and during hackathons, is a better way for us to provide provenance to a 'history' of spike sorting culminating in the final Units Table.

For example, let's say we start with the basic spike sorted output returned by a run of `ironclust` with some particular set of parameters. We then auto-curate it based on quality metrics and features using specified thresholds, and then manually curate the results in Phy. We then write these final spike sorted units to the NWBFile.

It would be very difficult for someone working with the same data to come close to replicating the results without knowing much of the detail preceding the final steps, but our current write functions only support adding directly to the primary `nwbfile.units` location, which is intended to be the highest form of processed units used in the experiment.

These changes to our write functions allow 

1. a description to be set on the table (including the primary table); algorithm parameters, manual or automated curation details could also be included there, etc.
2. the Units Table to be written to the ecephys processing module if it is intended to serve the role of provenance for the final units table
3. the name of the table to be changed (this will throw an error, however, if not writing to processing since the `nwbfile.units` table must be named 'units')

@bendichter @luiztauffer @Saksham20 Happy to discuss this as a team in our next meeting

## How to test the behavior?

The usual expansion tests have been added for the new arguments.

```
!pytest test_si.py
```

## Checklist


nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1002,61 +1004,62 @@ def write_units(
...
+                    shapes.append(1)
+                elif isinstance(prop_value, (list, np.ndarray)):
+                    if np.array(prop_value).ndim == 1:
+                        shapes.append(len(prop_value))
+                    else:
+                        shapes.append(np.array(prop_value).shape)
+                elif isinstance(prop_value, dict):
+                    print(f"Skipping property '{pr}' because dictionaries are not supported.")
+                    skip_properties.append(pr)
+                    break
```
bendichter, line 1026: This appears to me to be a completely different type of change that could be separated from the write location change

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1183,10 +1200,18 @@ def write_sorting(
     metadata: dict
         Information for constructing the nwb file (optional).
         Only used if no nwbfile exists at the save_path, and no nwbfile was directly passed.
+    write_as: str (optional, defaults to 'units')
+        How to save the units table in the nwb file. Options:
+        - 'units' will save it to the official NWBFile.Units position; recommended only for the final form of the data.
+        - 'processing' will save it to the processing module to serve as a historical provenance for the official table.
```
bendichter, line 1206: Oh, I see, you put the docstring here. Could you copy this over?

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1150,7 +1164,10 @@ def write_sorting(
     skip_properties: Optional[List[str]] = None,
     skip_features: Optional[List[str]] = None,
     use_times: bool = True,
-    metadata: dict = None,
+    metadata: Optional[dict] = None,
+    write_as: str = "units",
+    units_name: str = "units",
```
bendichter, line 1169: Can you add a check to make sure that if `write_as=="units"` then `units_name` must also be `"units"`

tests/test_si.py
```diff
@@ -354,6 +354,53 @@ def test_write_sorting(self):
...
+        with NWBHDF5IO(path=path, mode="r") as io:
+            nwbfile = io.read()
+            description_out = nwbfile.units.description
+        self.assertEqual(
+            description_out,
+            units_description,
+            "Intended units table description does not match what was written! "
+            f"(Out: {description_out}, should be: {units_description})",
+        )
+
```
bendichter, line 403: the tests look good!

bendichter:
@CodyCBakerPhD thanks for working on this. I agree this would be a great feature as it is something NWB supports without any trouble and it is a common need for NWB users. Your implementation here looks good to me overall, I've just made a few comments.

bendichter (changes requested):
see above

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: No changes other than the units table writing were included in this PR - this appears to be an artifact of the removal of [line #1005 from master](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/utils/spike_interface.py#L1005), which re-nests the entire code block below it, which GitHub has trouble parsing in the changelog for whatever reason.

It's easier to see this if you put these code blocks from [master](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/utils/spike_interface.py#L1005-L1059) and [units_provenance](https://github.com/catalystneuro/nwb-conversion-tools/blob/units_provenance/nwb_conversion_tools/utils/spike_interface.py#L1009-L1062) side by side for comparison.

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: Docstrings added!

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: Added!

bendichter (approved):

PR #237, opened 2021-08-26
Improve file vs folder annotations

CodyCBakerPhD:
## Motivation

@luiztauffer If you could check this works OK with your stuff

PR #213 introduced the concept of automating file vs. folder path detection in source schema assignment from our own locally controlled method signatures. The current implementation uses `typing.TypeVar` to built specific annotation types for this purpose, read more at

- https://stackoverflow.com/questions/58755948/what-is-the-difference-between-typevar-and-newtype
- https://www.python.org/dev/peps/pep-0484/#newtype-helper-function
- https://mypy.readthedocs.io/en/stable/more_types.html#newtypes

I had tried Ben's method suggested in a meeting of
```
PathType = Union[str, Path]
class FilePathType(PathType):
    pass
```
but couldn't quite get the inheritance behavior to mesh with the typing and signatures from annotations, even once I corrected for the general aliasing. Basically, for that to work the user would actually has to cast their `PathType` as an actual `FilePathType` (or we would have to do it each time ourselves in the interfaces; which doesn't really save us much code, which is the purpose of this PR).

The one slight downside to this is the check for annotation mapping for each argument requires equality
```
if arg == FilePathType:
      input_schema["properties"].update({param_name: dict(format="file")})
```
rather than the 'nicer' `isinstance` that accounts for inheritance.

## How to test the behavior?
Technically no tests have been changed, but the individual interfaces were debugged to reflect the automated schema changes and all tests should be passing for these changes.
```
!pytest
```

## Checklist

### DIFF
--- a/nwb_conversion_tools/__init__.py
+++ b/nwb_conversion_tools/__init__.py
@@ -1,3 +1,2 @@
 from .nwbconverter import NWBConverter
-
 from .datainterfaces import *
--- a/nwb_conversion_tools/datainterfaces/ecephys/axona/axonadatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/axona/axonadatainterface.py
@@ -3,30 +3,21 @@ import os
 import dateutil
 import numpy as np
 from pathlib import Path
-from typing import Union, Optional
+from typing import Union
 
 import spikeextractors as se
 from pynwb import NWBFile
 from pynwb.behavior import Position, SpatialSeries
-from pynwb.ecephys import ElectricalSeries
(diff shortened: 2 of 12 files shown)


bendichter (approved):
good to merge after resolving PR conflicts

CodyCBakerPhD:
@luiztauffer Do check this if you get a minute, to ensure the forms look like they figure out file vs. folder format correctly

luiztauffer:
@CodyCBakerPhD I checked and it doesn't seem to be working yet, see #244

PR #240, opened 2021-08-27
pip install wheel

luiztauffer:
The auto-release was missing at least this pip install @CodyCBakerPhD


luiztauffer:
@CodyCBakerPhD it looks like that's the case... however, shouldn't this specific dependency be updated? Or was that the last commit before the new SI refactor?

PR #243, opened 2021-08-31
adding nsx_to_load as arg for Blackrock files. Fixing tests

luiztauffer:
## Motivation

Latest changes broke BlackrockRecordingExtractorInterface.
@Saksham20 I reverted back your changes, because locally for me they broke the conversion, due to older spikeextractor not being able to deal with multiple streams. I believe your changes had the newer SI/NEO in mind?
I left them commented out for easy reference, cloud you please check them out?

## How to test the behavior?
```
from movshon_lab_to_nwb import MovshonBlackrockNWBConverter
from pathlib import Path

# Source data
base_path = Path('/home/luiz/storage/taufferconsulting/client_ben/project_movshon/data_blackrock_0/')
file_recording_raw = str(base_path / 'XX_LE_textures_20191128_002.ns6')
file_recording_processed = str(base_path / 'XX_LE_textures_20191128_002.ns3')
file_sorting = str(base_path / 'XX_LE_textures_20191128_002.nev')

source_data = dict(
    BlackrockRaw=dict(filename=file_recording_raw),
    BlackrockProcessed=dict(filename=file_recording_processed),
    BlackrockSorting=dict(
        filename=file_sorting,
        nsx_to_load=6
    )
)

# Initialize converter
converter = MovshonBlackrockNWBConverter(source_data=source_data)

# Get metadata from source data and modify any values you want
metadata['NWBFile']['session_description'] = 'my example conversion'

# Get conversion options and modify any values you want
conversion_options = converter.get_conversion_options()
conversion_options['BlackrockRaw'] = dict(stub_test=True)
conversion_options['BlackrockProcessed'] = dict(stub_test=True)

# Run conversion
output_file = 'out_example_blackrock.nwb'

converter.run_conversion(
    metadata=metadata, 
    nwbfile_path=output_file, 
    save_to_file=True,
    overwrite=True,
)
```

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
@@ -31,16 +31,10 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
         source_schema["properties"]["filename"]["description"] = "Path to Blackrock file."
         return source_schema
 
-    def __init__(self, filename: FilePathType, nsx_override: OptionalFilePathType = None):
-        nsx_to_load = int(str(filename).split(".")[-1][-1])
-        super().__init__(filename=filename, nsx_to_load=nsx_to_load)
-        # super().__init__(filename=filename, nsx_override=nsx_override)
-        # if self.source_data["nsx_override"] is not None:
-        #     # if 'nsx_override' is specified as a path, then the 'filename' argument is ignored
-        #     # This filename be used to extract the version of nsx: ns3/4/5/6 from the filepath
-        #     self.data_filename = self.source_data["nsx_override"]
-        # else:
-        #     self.data_filename = self.source_data["filename"]
+    def __init__(self, filename: FilePathType,
+                 nsx_override: OptionalFilePathType = None,
+                 nsx_to_load: OptionalFilePathType = None):
+        super().__init__(filename=filename, nsx_override=nsx_override, nsx_to_load=nsx_to_load)
 
     def get_metadata_schema(self):
         """Compile metadata schema for the RecordingExtractor."""
@@ -110,7 +104,7 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
         stub_test: bool, optional (default False)
             If True, will truncate the data to run the conversion faster and take up less memory.
         """
-        if self.source_data["filename"].split(".")[-1][-1] == "6":
+        if 6 in self.RX.neo_reader.nsx_to_load:
             write_as = "raw"
         elif write_as not in ["processed", "lfp"]:
             write_as = "processed"
--- a/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
@@ -135,7 +135,7 @@ class BlackrockSortingExtractorInterface(BaseSortingExtractorInterface):
     def get_source_schema(cls):
         """Compile input schema for the RecordingExtractor."""
         metadata_schema = get_schema_from_method_signature(
-            class_method=cls.SX.__init__, exclude=["block_index", "seg_index", "nsx_to_load"]
+            class_method=cls.__init__, exclude=["block_index", "seg_index"]
         )
         metadata_schema["additionalProperties"] = True
         metadata_schema["properties"]["file_path"]["description"] = "Path to Blackrock file."
--- a/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
@@ -107,7 +107,7 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
         stub_test: bool, optional (default False)
             If True, will truncate the data to run the conversion faster and take up less memory.
         """
-        if 6 in self.recording_extractor.neo_reader.nsx_to_load:
+        if max(self.recording_extractor.neo_reader.nsx_to_load) >= 5:
             write_as = "raw"
         elif write_as not in ["processed", "lfp"]:
             write_as = "processed"
--- a/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
@@ -28,16 +28,16 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
         source_schema = get_schema_from_method_signature(
             class_method=cls.__init__, exclude=["block_index", "seg_index"]
         )
-        source_schema["properties"]["file_path"]["description"] = "Path to Blackrock file."
+        source_schema["properties"]["filename"]["description"] = "Path to Blackrock file."
         return source_schema
 
     def __init__(
         self,
-        file_path: FilePathType,
+        filename: FilePathType,
         nsx_override: OptionalFilePathType = None,
         nsx_to_load: Optional[int] = None,
     ):
-        super().__init__(file_path=file_path, nsx_override=nsx_override, nsx_to_load=nsx_to_load)
+        super().__init__(filename=filename, nsx_override=nsx_override, nsx_to_load=nsx_to_load)
 
     def get_metadata_schema(self):
         """Compile metadata schema for the RecordingExtractor."""
@@ -53,7 +53,7 @@ class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
         metadata = super().get_metadata()
 
         # Open file and extract headers
-        nsx_file = NsxFile(datafile=self.source_data["file_path"])
+        nsx_file = NsxFile(datafile=self.source_data["filename"])
         session_start_time = nsx_file.basic_header["TimeOrigin"]
         session_start_time_tzaware = pytz.timezone("EST").localize(session_start_time)
         comment = nsx_file.basic_header["Comment"]
@@ -138,10 +138,10 @@ class BlackrockSortingExtractorInterface(BaseSortingExtractorInterface):
             class_method=cls.__init__, exclude=["block_index", "seg_index"]
         )
         metadata_schema["additionalProperties"] = True
-        metadata_schema["properties"]["file_path"]["description"] = "Path to Blackrock file."
+        metadata_schema["properties"]["filename"]["description"] = "Path to Blackrock file."
         return metadata_schema
 
     def __init__(
-        self, file_path: FilePathType, nsx_to_load: Optional[int] = None, nev_override: OptionalFilePathType = None
+        self, filename: FilePathType, nsx_to_load: Optional[int] = None, nev_override: OptionalFilePathType = None
     ):
-        super().__init__(file_path=file_path, nsx_to_load=nsx_to_load, nev_override=nev_override)
+        super().__init__(filename=filename, nsx_to_load=nsx_to_load, nev_override=nev_override)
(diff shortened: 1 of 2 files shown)


Saksham20:
@luiztauffer can you paste the error trace here?
These changes were done for the old spikeextractors version, so they should work. Did you update to the latest spikeextractors ? THere were corresponding changes even in spikeextractors also to handle this. See here https://github.com/catalystneuro/nwb-conversion-tools/pull/214

luiztauffer:
@Saksham20 
I'm using spikeextractors v 0.9.7
here's the error:
```python
<ipython-input-3-16cdbf18b1df> in <module>
     15 
     16 # Initialize converter
---> 17 converter = MovshonBlackrockNWBConverter(source_data=source_data)
     18 
     19 # Get metadata from source data and modify any values you want

/media/luiz/storage/Github/nwb-conversion-tools/nwb_conversion_tools/nwbconverter.py in __init__(self, source_data)
     67         self.data_interface_objects = {
     68             name: data_interface(**source_data[name])
---> 69             for name, data_interface in self.data_interface_classes.items()
     70             if name in source_data
     71         }

/media/luiz/storage/Github/nwb-conversion-tools/nwb_conversion_tools/nwbconverter.py in <dictcomp>(.0)
     68             name: data_interface(**source_data[name])
     69             for name, data_interface in self.data_interface_classes.items()
---> 70             if name in source_data
     71         }
     72 

/media/luiz/storage/Github/nwb-conversion-tools/nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py in __init__(self, filename, nsx_override)
     33 
     34     def __init__(self, filename: FilePathType, nsx_override: OptionalFilePathType = None):
---> 35         super().__init__(filename=filename, nsx_override=nsx_override)
     36         if self.source_data["nsx_override"] is not None:
     37             # if 'nsx_override' is specified as a path, then the 'filename' argument is ignored

/media/luiz/storage/Github/nwb-conversion-tools/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py in __init__(self, **source_data)
     34     def __init__(self, **source_data):
     35         super().__init__(**source_data)
---> 36         self.recording_extractor = self.RX(**source_data)
     37         self.subset_channels = None
     38         self.source_data = source_data

~/anaconda3/envs/env_movshon/lib/python3.7/site-packages/spikeextractors/extractors/neoextractors/blackrockextractor.py in __init__(self, filename, nsx_to_load, block_index, seg_index, **kwargs)
     35                  seg_index: Optional[int] = None, **kwargs):
     36         super().__init__(filename=filename, nsx_to_load=nsx_to_load, 
---> 37                          block_index=block_index, seg_index=seg_index, **kwargs)
     38 
     39 

~/anaconda3/envs/env_movshon/lib/python3.7/site-packages/spikeextractors/extractors/neoextractors/neobaseextractor.py in __init__(self, block_index, seg_index, **kargs)
     76             raise ValueError('Strange neo version. Please upgrade your neo package: pip install --upgrade neo')
     77 
---> 78         assert num_streams == 1, 'This file have several signal streams spikeextractors support only one streams' \
     79                                  'Maybe you can use option to select only one stream'
     80 

AssertionError: This file have several signal streams spikeextractors support only one streamsMaybe you can use option to select only one stream
```

Saksham20:
@luiztauffer , spikeextractors was unable to support multiple streams of data even before my change. Are you able to run it after reverting my changes? 
The change that I had done was only if you would want to override the default nsx file that neo locates from the filename argument.

luiztauffer:
@Saksham20 yes, I was able to avoid that error after reverting your changes

Saksham20:
@luiztauffer 
can you try:
```python
source_data = dict(
    BlackrockRaw=dict(filename=file_recording_raw),
    BlackrockProcessed=dict(filename=file_recording_processed),
    BlackrockSorting=dict(
        filename=file_sorting,
        nsx_override=file_recording_raw
    )
)
```

luiztauffer:
@Saksham20 I overlooked the difference in args names, thanks for pointing that out. But it did not solve the problem
any specific reason why you changed the previous use of `nsx_to_load` for `nsx_override`?

Saksham20:
@JuliaSprenger I'm involving you in this conversation as I think this issue is something on the neo end: when reading a blackrock file using `BlackRockIO`, what is the difference between the arguments `nsx_override `and `nsx_to_load` in the case when there is a specific nsx version to load ?
 if this command https://github.com/catalystneuro/nwb-conversion-tools/pull/243#issuecomment-914269964
is run with `nsx_overide = 'nsx_path.ns6'` vs `nsx_to_load = 6` for the same filename?

JuliaSprenger:
Hi @Saksham20!
The `nsx_overide` permits to include an unrelated nsx file into the current set of session files. The nsx file does not need to be located in the same folder as the other session files. In contrast to that `nsx_to_load` permits you a fine grained choice of which nsx files of a given set of session files is going to be used as you can specify a number here to select a single file or a list of number corresponding to a set of nsx extensions to be considered or the maximum or all nsx detected. Here the excerpt from the docstring
```
        nsx_override (string):
            File name of the .nsX files (without extension). If None,
            filename is used.
            Default: None.
        nsx_to_load (int, list, 'max', 'all' (=None)) default None:
            IDs of nsX file from which to load data, e.g., if set to
            5 only data from the ns5 file are loaded.
            If 'all', then all nsX will be loaded.
            Contrary to previous version of the IO  (<0.7), nsx_to_load
            must be set at the init before parse_header().
```

In the case where you want to load a complete session and select one or multiple nsx to use from the set of session files I would therefore recommend to use `nsx_to_load` as you don't want to include an 'external' file. However, I remember from an earlier discussion that @Saksham20 wants to load an isolated nsx file without any accompanying files. In that case `filename=''` and `nsx_to_load='<isolated_nsx_filename>` would be the way to go. @Saksham20 do I remember this correctly?

Saksham20:
>  I remember from an earlier discussion that @Saksham20 wants to load an isolated nsx file without any accompanying files. In that case `filename=''` and `nsx_to_load='<isolated_nsx_filename>` would be the way to go. @Saksham20 do I remember this correctly?

Yes previously, I wanted to have neo only load a specific nsx file and not load any nev file so I had the `filename` arg as empty and only specified the `nsx_override` argument. But for a given session files, wouldn't setting `nsx_to_load=6` (looks for a .ns6 file only) and `nsx_override=path/to/*.ns6` work the same?

JuliaSprenger:
Not exactly, since `nsx_override` is used to override the stem of the nsx filename (the extension will be ignored). In the doc this is ndicated by 'nsx_override - File name of the .nsX files **(without extension)**'. So if you provide another stem for potential nsx files, then also all `stem.ns[1,2,3,4,5,6]` files will be loaded. @Saksham20 Does this help?

Saksham20:
> Not exactly, since `nsx_override` is used to override the stem of the nsx filename (the extension will be ignored). In the doc this is ndicated by 'nsx_override - File name of the .nsX files **(without extension)**'. So if you provide another stem for potential nsx files, then also all `stem.ns[1,2,3,4,5,6]` files will be loaded. @Saksham20 Does this help?

Okay, that makes sense now, so each argument (`nsx_to_load` and `nsx_override`) have their own advantages. 
@luiztauffer I've added the `nsx_to_load` as an additional argument. Can you try with the latest commit?
@JuliaSprenger So when calling `BlackrockRawIO._get_analog_signal_chunk()` with `segment index = 0` (this is what spikeextractors uses to retrieve the traces), which nsx number is loaded when there are multiple nsx version available in the folder?

luiztauffer:
@Saksham20 the way it was before, `ndx_to_load` was automatically extracted from the file name passed by the user, what's the advantage of having it the way you did now?
It does make sense on SortingExtractor (to specify n6 as source of traces for example), but for RecordingExtractor it needs to be the same as the file passed by the user

JuliaSprenger:
> @JuliaSprenger So when calling `BlackrockRawIO._get_analog_signal_chunk()` with `segment index = 0` (this is what spikeextractors uses to retrieve the traces), which nsx number is loaded when there are multiple nsx version available in the folder?

The different AnalogSignals are handled as different streams. When calling `_get_analogsignal_chunk(block_index, seg_index, i_start, i_stop,  stream_index, channel_indexes)` you select which stream to consider via the `stream_index`. For the BlackrockIO the order of the streams is the same as in `nsx_to_load` (assuming all streams are non-empty). Note that the stream concept is relatively new, so this holds for neo 0.10.0 but not for older versions.

Also another side note: You might miss part of the data if you only consider `segment_index = 0`, as a recording can be split into multiple segments, e.g. in case of broken data packets or a recording break.

Saksham20:
@JuliaSprenger got it. 
Taking an example if the file structure is like this: 
```
parent_folder/
  abcd.ns6
  abcd.ns3
  abcd.nev
```
then if I call `BlackrockRawIO(filename='parent_folder/abcd.ns6)` then shouldn't the number of streams (the number of nsx files found in the folder) be ==2?

@luiztauffer what are you doing different which is not causing the error after you revert my changes?

The below causes the error right?
```python
# Source data
# essentially there are two files so num_streams should be 2 no? And recordingextractor supports only 1 stream so there should be an error
base_path = Path('/home/luiz/storage/taufferconsulting/client_ben/project_movshon/data_blackrock_0/')
file_recording_raw = str(base_path / 'XX_LE_textures_20191128_002.ns6')
file_recording_processed = str(base_path / 'XX_LE_textures_20191128_002.ns3')
file_sorting = str(base_path / 'XX_LE_textures_20191128_002.nev')

source_data = dict(
    BlackrockRaw=dict(filename=file_recording_raw), # equivalent to the BlackrockRawIO(filename=file_recording_raw) so streams count should be ==2
    BlackrockProcessed=dict(filename=file_recording_processed),
    BlackrockSorting=dict(
        filename=file_sorting,
        nsx_to_load=6
    )
)
```

JuliaSprenger:
> @JuliaSprenger got it.
> Taking an example if the file structure is like this:
> 
> ```
> parent_folder/
>   abcd.ns6
>   abcd.ns3
>   abcd.nev
> ```
> 
> then if I call `BlackrockRawIO(filename='parent_folder/abcd.ns6)` then shouldn't the number of streams (the number of nsx files found in the folder) be ==2?

Yes, it should be `io.nsx_to_load == [3,6]`. Is it different in your case?

Saksham20 (approved):

luiztauffer:
@Saksham20 the change proposed on this PR was a simple thing, as you can see in the first commit:
```python
nsx_to_load = int(str(filename).split('.')[-1][-1])
super().__init__(filename=filename, nsx_to_load=nsx_to_load)
```
For the Recording data interface, letting `nsx_to_load` as a source_data arg is redundant in the best case and ambiguous in the worst, since the user already passes the path to the specific file they want to convert. The original code just extracted the file extension and passed that automatically to NEO as `nsx_to_load`. 
Is there any reason why you want to change this behavior?

Saksham20:
> For the Recording data interface, letting `nsx_to_load` as a source_data arg is redundant in the best case and ambiguous in the worst, since the user already passes the path to the specific file they want to convert. The original code just extracted the file extension and passed that automatically to NEO as `nsx_to_load`.
> Is there any reason why you want to change this behavior

>@Saksham20 the change proposed on this PR was a simple thing, as you can see in the first commit:
```python
nsx_to_load = int(str(filename).split('.')[-1][-1])
super().__init__(filename=filename, nsx_to_load=nsx_to_load)
```

all these arguments are input into the blackrockIO in neo. nsx_to_load and nsx_overide serve different functions so I thought of exposing them at the datainterface level also would increase the functionality. In this PR i've added nsx_to_load back with the nsx_override also. 

The reason I had needed `nsx_overide` in the past was because, I had a bunch of datasets for which there were no nev files but only nsx (so I could only read the raw analog traces). For this to work I had put the `filepath=""` and `nsx_override="path-to-nsx.ns6"`

All the other commits in this PR were for correcting the tests.

Saksham20:
@luiztauffer I agree that we can automatically extract the `nsx_to_load` from the file ext input in the `filename` argument. 

Note, there is a case where a user may enter an empty str as filename and enter only `nsx_override` as arg. : this is the case when there is no nev file present with the nsx file. 
This feature is now a separate PR: #263

PR #246, opened 2021-09-07
Adding SpikeGadgets tests

CodyCBakerPhD:
## Motivation

Adding SpikeGadgets tests for the first time on the spikeextractor/interface end. This process revealed an issue in the roundtrip reader through NwbRecordingExtractor when dealing with scrambled channel id order (fixed included on master branch of spikeextractors, hence why the requirements point to this).

The most important thing to be aware of with the SpikeGadgets format is how, in this early version data at least, there are no factors (`gains`) internally stored for conversion from the raw `int16` data type. As such, the user is encouraged to set these manually, which I've built into the interface here as an optional argument that performs the relevant modification to the `recording_extractor` object.

This setting is also included in the new testing matrix for the interface, which has been refactored to better allow that sort of matrix over the same `RecordingInterface`, which greatly reduces the changelog for that file.

## How to test the behavior?
Tests added to the `test_gin` family.

```
!pytest
```

## Checklist


requirements-dev.txt
```diff
@@ -10,7 +10,7 @@ jsonschema==3.2.0
 psutil==5.8.0
 lxml==4.6.3
 opencv-python==4.5.1.48
-spikeextractors==0.9.7
+spikeextractors @ git+https://github.com/spikeinterface/spikeextractors.git@master
```
bendichter, line 13: Is this necessary? It looks like the last change to SpikeGadgets was in March and the latest release was in July. Is there some feature this needs that I am missing?

CodyCBakerPhD: Mentioned why it's necessary in PR description - requires a debug that was quickly merged to master the other day: [se PR #662](https://github.com/SpikeInterface/spikeextractors/pull/662)

bendichter: @alejoe91 would it be possible to push a release of spikeextractors?

CodyCBakerPhD: Done, waiting for tests to pass now

CodyCBakerPhD:
@bendichter Version pin to spikeextractors has now been updated

bendichter (approved):

PR #248, opened 2021-09-08
Add phy gin

CodyCBakerPhD:
## Motivation

Adding SortingInterface GIN test framework, starting with the PhySortingInterface tests. There is some pickling error occurring from spikeextractors occurring during the equality check when I run locally, seeing if it occurs on Linux as well.

## How to test the behavior?
```
pytest
```

## Checklist

### DIFF
--- a/tests/test_gin.py
+++ b/tests/test_gin.py
@@ -3,13 +3,14 @@ import tempfile
 import unittest
 from pathlib import Path
 
-from spikeextractors import NwbRecordingExtractor
-from spikeextractors.testing import check_recordings_equal
+from spikeextractors import NwbRecordingExtractor, NwbSortingExtractor
+from spikeextractors.testing import check_recordings_equal, check_sortings_equal
 from nwb_conversion_tools import (
     NWBConverter,
     IntanRecordingInterface,
     NeuralynxRecordingInterface,
     NeuroscopeRecordingInterface,
+    PhySortingInterface,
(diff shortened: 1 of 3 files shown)


tests/test_gin.py
```diff
@@ -28,7 +29,7 @@
     HAVE_PARAMETERIZED = False
```
bendichter, line 29: I think `parameterized` should be listed in `requirements-dev.txt`, so we should be able to expect that it is downloaded and shouldn't need this import catch

bendichter:
@CodyCBakerPhD can you take a look at these changes?

tests/test_gin.py
CodyCBakerPhD: It is [https://github.com/catalystneuro/nwb-conversion-tools/blob/add_phy_gin/requirements-dev.txt#L19](https://github.com/catalystneuro/nwb-conversion-tools/blob/add_phy_gin/requirements-dev.txt#L19) but it's still needed when running pytest locally (not on CI) in case the developer has not installed it

CodyCBakerPhD:
@bendichter Those do look nicer; they will produce merge conflicts for the other 2 PRs though but I can handle that if this one gets merged first

bendichter:
@CodyCBakerPhD can you fix the merge conflict? Then this is good to merge

CodyCBakerPhD:
@bendichter Already on it, will merge when done. Thanks!

PR #252, opened 2021-09-10
Fix sorting interface

CodyCBakerPhD:
## Motivation

Long overdue update of the BaseSortingInterface to match our API design in terms of internal properties vs. heavy metadata.

Also updates the schema to more properly reflect the forms. Basically has an identical setup as the custom columns of the Electrodes table but for Units tables instead.

Attn: @luiztauffer 

Edit: Also now includes SpikeInterface `.pkl` tests

## How to test the behavior?

Doesn't seem as if there are any explicit tests for metadata schema validity that span the interfaces (the GIN tests and maybe a couple others do this implicitly). Attempting to add them causes a very large number of errors, and fixing them would be beyond the scope of this PR.

So I'll add and debug metadata schema tests in a subsequent PR branching off this one.

## Checklist


nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -27,16 +26,38 @@ def __init__(self, **source_data):
...
+        metadata_schema["properties"]["Ecephys"]["properties"]["definitions"] = dict(
+            UnitProperties=dict(
+                type="object",
+                additionalProperties=False,
+                required=["name"],
+                properties=dict(
+                    name=dict(type="string", description="name of this units column"),
+                    description=dict(type="string", description="description of this units column"),
+                ),
+            ),
```
bendichter, line 52: shouldn't this be an array?

CodyCBakerPhD: That would be a question for @luiztauffer 

I basically copied the way the electrode custom columns are specified by the schema seen [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/fix_sorting_interface/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py#L46-L66)

CodyCBakerPhD: As I understand though, this level of "object" specifies that what is inside the 'UnitProperties' array are dictionaries with keys 'name' and 'description', with the code above this specifying that the 'UnitProperties' field itself is an array

nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -27,16 +26,38 @@ def __init__(self, **source_data):
...
+
+        # Initiate Ecephys metadata
+        metadata_schema["properties"]["Ecephys"] = get_base_schema(tag="Ecephys")
+        metadata_schema["properties"]["Ecephys"]["required"] = []
+        metadata_schema["properties"]["Ecephys"]["properties"] = dict(
+            UnitProperties=dict(
+                type="array",
+                minItems=0,
+                renderForm=False,
+                items={"$ref": "#/properties/Ecephys/properties/definitions/UnitProperties"},
```
bendichter, line 39: I think it's a bit odd to have definitions deep in the schema as opposed to just at the root level

CodyCBakerPhD: Yet another question for @luiztauffer, I don't know why this is the way it is for the forms. Any ideas or simpler ways to do it @luiztauffer?

bendichter: @luiztauffer why are we using definitions here

luiztauffer: @bendichter @CodyCBakerPhD 
tbh I don’t remember why we made this choice, it was a while back. 
Since these schemas might be overwritten by subsequent DataInterface schemas in a multi-interface Converter, maybe it makes sense to keep this level of separation by modules, to avoid eventual overwriting of same definitions but with different default values, for example. But I can’t really think on an example where this would happen right now.
In principle these should work just the same, when it comes to validation:
- {"$ref": "#/properties/Ecephys/properties/definitions/UnitProperties"}
- {"$ref": "#/properties/definitions/UnitProperties"}
- {"$ref": "#/definitions/UnitProperties"}

But I can’t guarantee changing this won’t break the converters or the GUI Forms at this moment.
We can open an Issue for that, if you think these definitions should move to the root.

bendichter (approved):

PR #255, opened 2021-09-13
Icephys and Neo interfaces

luiztauffer:
## Motivation

**WIP**

Creates Icephys module for data interfaces, using Neo as base.

- `BaseIcephysNeoInterface`, which accepts multiple input files, and forms a list of neo readers accordingly
- `AbfNeoDataInterface`
- a more general ` add_devices` function in `conversion_tools.py`, which creates devices based on metadata from any neurodata type (Ecephys, Icephys, Ophys, Behavior). Previously it only read it from Ecephys metadata.
- new module `utils/neo.py` with utils functions similar to `utils/spike_interface.py`, but for neo+icephys

## How to test the behavior?
```python
from nwb_conversion_tools import AbfNeoDataInterface, NWBConverter
from pathlib import Path

class GuagweiConverter(NWBConverter):
    data_interface_classes = dict(AbfNeoDataInterface=AbfNeoDataInterface)

# Source data
directory = Path("Oct_2021_DANDI/20210708002")
file_list = list()
for x in directory.iterdir():
    if x.is_file() and x.suffix == '.abf':
        file_list.append(str(x.resolve()))
    if x.is_file() and x.name == 'metadata.json':
        metadata_file_path = str(x.resolve())

source_data = dict(
    AbfNeoDataInterface=dict(
        files_paths=file_list,
        metadata_file_path=metadata_file_path
    )
)

# Initialize converter
converter = GuagweiConverter(source_data=source_data)

# Get metadata from source data and modify any values you want
metadata['NWBFile']['session_description'] = 'my icephys experiment'

# Run conversion
output_file = 'out_example.nwb'

converter.run_conversion(
    metadata=metadata, 
    nwbfile_path=output_file, 
    save_to_file=True,
    overwrite=True,
)
```

## Checklist


nwb_conversion_tools/datainterfaces/icephys/abf/abf_neo_datainterface.py
```diff
@@ -0,0 +1,45 @@
...
+        metadata = super().get_metadata()
+
+        # Extract start_time info
+        startDate = str(self.reader._axon_info["uFileStartDate"])
+        startTime = round(self.reader._axon_info["uFileStartTimeMS"] / 1000)
+        startDate = datetime.strptime(startDate, "%Y%M%d")
+        startTime = timedelta(seconds=startTime)
+        abfDateTime = startDate + startTime
+        # Time Zone
+        session_start_time_tzaware = pytz.timezone("UTC").localize(abfDateTime)
```
bendichter, line 36: Are these times always in UTC?

luiztauffer: I'll look further as this info might be available somewhere in the abf file, but tz info is not available from the neo reader

luiztauffer: via metadata, but user would have to:
1. get metadata, which would return the right timestamp, but the default EST
2. update the session_start_time with the correct TZ
3. pass this metadata to run_conversion

nwb_conversion_tools/utils/neo.py
```diff
@@ -0,0 +1,570 @@
...
+    Get number of segments from Neo reader
+
+    Args:
+        neo_reader ([type]): Neo reader
+        block (int, optional): Block id. Defaults to 0.
+
+    Returns:
+        int: number of electrodes
+    """
+    return neo_reader.header["nb_segment"][block]
```
samuelgarcia, line 80: You should take in account the case with multiple streams.

nwb_conversion_tools/utils/neo.py
```diff
@@ -0,0 +1,570 @@
...
+    """
+    Get number of electrodes from Neo reader
+
+    Args:
+        neo_reader ([type]): Neo reader
+
+    Returns:
+        int: number of electrodes
+    """
+    return len(neo_reader.header["signal_channels"])
```
samuelgarcia, line 65: You should take in account the case with multiple streams.

nwb_conversion_tools/utils/neo.py
```diff
@@ -0,0 +1,570 @@
...
+    Args:
+        neo_reader ([type]): [description]
+        block (int, optional): [description]. Defaults to 0.
+        segment (int, optional): [description]. Defaults to 0.
+        cmd_channel (int, optional): ABF command channel (0 to 7). Defaults to 0.
+
+    Returns:
+        list: [description]
+    """
+    traces, titles, units = neo_reader.read_raw_protocol()
```
samuelgarcia, line 97: This is not generic to neo but only for AxonIO case

nwb_conversion_tools/utils/neo.py
```diff
@@ -0,0 +1,570 @@
...
+    if unit in ["pA", "pV"]:
+        gain = 10 ** -12
+    elif unit in ["nA", "nV"]:
+        gain = 10 ** -9
+    elif unit in ["uA", "uV"]:
+        gain = 10 ** -6
+    elif unit in ["mA", "mV"]:
+        gain = 10 ** -3
+    elif unit in ["A", "V"]:
+        gain = 10 ** 0
```
samuelgarcia, line 121: You should handle the case outside this list. It happen for some files.

nwb_conversion_tools/utils/neo.py
luiztauffer: for now AxonIO is the only Icephys we're dealing with, we can expand this for other cases when they appear

nwb_conversion_tools/utils/neo.py
luiztauffer: Im now passing a warning message, anything else you suggest, @samuelgarcia?

luiztauffer:
@CodyCBakerPhD this is now closed

PR #256, opened 2021-09-13
optional nsx header fields

Saksham20:
## Motivation

Some nsx files do not have the keys that `BlackrockRecordingExtractorInterface` looks for in constructing the metadata. This PR makes it optional so that it does not throw an error when not found. 

@luiztauffer you had ported this code in its first commit so can you give a quick glance. 

## Checklist


nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
```diff
@@ -51,18 +51,15 @@ def get_metadata_schema(self):
...
-        comment = nsx_file.basic_header["Comment"]
-
-        # Updates basic metadata from files
-        metadata["NWBFile"] = dict(
-            session_start_time=session_start_time_tzaware.strftime("%Y-%m-%dT%H:%M:%S"),
-            session_description=comment,
-        )
+        if "TimeOrigin" in nsx_file.basic_header:
+            session_start_time = nsx_file.basic_header["TimeOrigin"]
+            session_start_time_tzaware = pytz.timezone("EST").localize(session_start_time)
```
bendichter, line 59: why are we assuming EST?

luiztauffer: as I remember it, there is no tz info in blackrock files, we need to assume something

luiztauffer:
@Saksham20 thanks for those fixes!

luiztauffer (approved):

bendichter:
Would there be a way of inputting the timezone information?

luiztauffer:
> Would there be a way of inputting the timezone information?

via metadata, but user would have to:
1. get metadata, which would return the right timestamp, but the default EST
2. update the session_start_time with the correct TZ
3. pass this metadata to run_conversion

bendichter:
yeah I suppose that's not that bad

bendichter:
Could we return a non-localized time?

luiztauffer:
> Could we return a non-localized time?

yes, but pynwb will raise a warning at conversion time, I believe

bendichter:
Yes I think it's OK to have a warning. That will automatically localize to the user's timezone, which I think it the best thing to guess by default, and will also throw a warning, indicating that the user should add this information themselves if they want to override this default. When we show a demo using the platform, we can show how to apply a timezone localization.

nwb_conversion_tools/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
```diff
@@ -51,18 +51,15 @@ def get_metadata_schema(self):
...
+            metadata["NWBFile"].update(session_start_time=session_start_time_tzaware.strftime("%Y-%m-%dT%H:%M:%S"))
```
bendichter, line 60:
```suggestion
            metadata["NWBFile"].update(session_start_time=session_start_time.strftime("%Y-%m-%dT%H:%M:%S"))
```

luiztauffer:
@Saksham20 thanks! let me run some local tests and I'll merge it soon

PR #257, opened 2021-09-13
Fix remaining annotations

CodyCBakerPhD:
## Motivation

We have recently enforced local control of method signature for the automated setting of source schemas for `ecephys` interfaces - this primarily extends it to the `ophys` interfaces as well removes a few random references to unspecified PathTypes that were previously missed.

All references to PathType across the repo should now be properly specified as `FilePathType`, `OptionalFilePathType`, or `FolderPathType` and import these directly from the utils rather than redefining them every time.

## How to test the behavior?
No changes to actual tests, but they are quite useful in checking that nothing gets broken with the update.

```
pytest
```

## Checklist


bendichter (approved):
looks great!

CodyCBakerPhD:
Cool, I'll merge once the full tests finish making sure nothing is broken.

PR #258, opened 2021-09-13
Standardize file folder input args

CodyCBakerPhD:
## Motivation

The new annotation method that gives us full control over path types also now allows us to control localized naming conventions rather than having to rely on the underlying package scheme. We therefore have an opportunity to standardize how all our interfaces operate here with respect to input argument naming schemes, which is a nice convenience IMO and came up briefly in our last team meeting.

This PR standardizes all `FilePathType`s to correspond to an input argument called `file_path`. It mostly affects the previous interfaces reliant on `neo` as the extractor, since these instead usually follow the convention of `filename`.

This would introduce a break to back-compatibility with previous versions, so if the convenience of this is agreed upon we can wait to merge until the next big version update (likely the SpikeInterface2 writers + Icephys). Something we could do now if this is approved is to add some deprecation warnings to the affected interfaces to let people know of the upcoming change.

## How to test the behavior?
`test_interfaces` ensures nothing breaks in the schema validation with the name changes; I had to update `test_gin` to properly use the adjusted naming, indicative of the break in back compatability

```
pytest
```

## Checklist


bendichter:
I agree this would be great to have, but that we should be careful about backwards compatibility so I agree we should wait a bit

CodyCBakerPhD:
Swapping branches

PR #263, opened 2021-09-14
autopick nsx_to_load from filename

Saksham20:
## Motivation

This is the feature that @luiztauffer suggested in #243 .


luiztauffer:
@Saksham20 thanks! let me run some local tests and I'll merge it soon

luiztauffer (approved):
looks good, thanks for fixing this

PR #264, opened 2021-09-15
pass one movie filepath to MovieInterface

Saksham20:
## Motivation

Providing option to overrite the name, description and other metadata for ImageSeries that stores the Movie. 

The `MovieInterface.run_conversion()` takes metadata as arg but it is not used.

### DIFF
--- a/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
@@ -13,7 +13,7 @@ from hdmf.data_utils import DataChunkIterator
 
 from ....basedatainterface import BaseDataInterface
 from ....utils.conversion_tools import check_regular_timestamps, get_module
-from ....utils.json_schema import get_schema_from_method_signature
+from ....utils.json_schema import get_schema_from_method_signature, FilePathType
 from .movie_utils import get_movie_timestamps, get_movie_fps, get_frame_shape
 
 
@@ -29,7 +29,7 @@ INSTALL_MESSAGE = "Please install opencv to use this extractor (pip install open
 class MovieInterface(BaseDataInterface):
     """Data interface for writing movies as ImageSeries."""
 
-    def __init__(self, file_paths: list):
+    def __init__(self, file_path: FilePathType):
         """
         Create the interface for writing movies as ImageSeries.
 
@@ -40,7 +40,7 @@ class MovieInterface(BaseDataInterface):
             Pass the file paths for this movies as a list in sorted, consecutive order.
         """
         assert HAVE_OPENCV, INSTALL_MESSAGE
-        super().__init__(file_paths=file_paths)
+        super().__init__(file_path=file_path)
 
     @classmethod
     def get_source_schema(cls):
@@ -52,7 +52,7 @@ class MovieInterface(BaseDataInterface):
         metadata: dict,
         stub_test: bool = False,
         external_mode: bool = True,
-        starting_times: Optional[list] = None,
+        starting_time: Optional[float] = None,
         chunk_data: bool = True,
         module_name: Optional[str] = None,
         module_description: Optional[str] = None,
@@ -72,7 +72,7 @@ class MovieInterface(BaseDataInterface):
             data sharing, the video files must be contained in the same folder as the NWBFile. If the intention of this
             NWBFile involves an upload to DANDI, the non-NWBFile types are not allowed so this flag would have to be
             set to False. The default is True.
-        starting_times : list, optional
+        starting_time : list, optional
             List of start times for each movie. If unspecified, assumes that the movies in the file_paths list are in
             sequential order and are contiguous.
         chunk_data : bool, optional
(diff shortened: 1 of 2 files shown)


bendichter:
@Saksham20 is there a reason you can't just use a list of length 1?

Saksham20:
If we need to write a list of ImageSeries, then, we would also need to pass metadata as a list of dicts to override default info for each of the movies

CodyCBakerPhD:
> ## Motivation
> * Providing option to overrite the name, description and other metadata for ImageSeries that stores the Movie.
> * Pass a single movie to write to nwb

Glad to see you're using the interface!

I'd say using `metadata` control over naming stuff is perfectly fine to add in.

However, I see no reason you can't use a list of a single movie in the situation you are describing, and a corresponding ordered list of dictionaries for the metadata. Doing it as you suggest would be a major version increase so we would want to wait a while for merging it with a couple other similar PRs.

But also, beyond that, the vectorized notion of the `file_paths` list and `starting_times` is very particularly designed around a common case where multiple sequential segments of movies are recorded in tandem, so this PR breaks that feature of the interface, which I can't approve.

Keep in mind opening an issue or discussion of the design for these kinds of features prior to a PR can save time spent coding stuff that isn't desired.

Saksham20:
OKay,  we could potentially have a list of movies written in sequentially but in the metadata, have the `ImageSeries` as a list of dicts under the aquisition dict

Saksham20:
@CodyCBakerPhD what I suggest as an alternative is something like the latest commit https://github.com/catalystneuro/nwb-conversion-tools/pull/264/commits/fab21b556ea2773b266a6dbbb72b060301e820ff

CodyCBakerPhD:
@Saksham20 Looking better; this now opens up an interesting consideration for what key the metadata should be specified under. I see you're by default in `get_metadata()` setting it to 'acquisition' but you are correctly considering the conversion option argument `module_name` in the ultimate retrieval.

I suppose the intended use case would then be `metadata["acquisition"]` by default, then if the user wants to specify a particular `module_name` in the `conversion_options`, they would also be required to (a) manually remove the `metadata["acquisition"]` field for that movie(s) and (b) manually add that metadata with custom naming to `metadata[module_name]`?

I'll think on that workflow for a bit to see if there isn't something better...

Saksham20:
>I suppose the intended use case would then be metadata["acquisition"] by default, then if the user wants to specify a particular module_name in the conversion_options, they would also be required to (a) manually remove the metadata["acquisition"] field for that movie(s) and (b) manually add that metadata with custom naming to metadata[module_name]?

@CodyCBakerPhD , yes one way a user could specify the new module name is to create a new module name (as dict key in metadata after retrieving from `converter.get_metadata()`) and put alll the image_series kwargs dict under that like so:
```python
metadata=datainterface.get_metadata()
metadata[module_name]=dict(ImageSeries=metadata['acquisition']['ImageSeries'])
_=metadata['acquisition'].pop('ImageSeries')
#.. then go about changing the 'ImageSeries' kwargs as needed. 
datainterface.run_conversion(metadata=metadata,module_name=module_name,module_description=description)
```
Or the other way is that the `MovieInterface` can have 2 additional args: `module_name`,`module_description` that the user wishes the `ImageSeries` goes under. Then we can create the metadata from `interface.get_metadata()` accordingly and update the schema to conform to this like so:
```python
    def get_metadata(self):
        metadata = super().get_metadata()
        default_image_series = dict(
            ImageSeries=[
                dict(name=f"Video: {Path(file_path).stem}", description="Video recorded by camera.", unit="Frames")
                for file_path in self.source_data["file_paths"]
            ]
        )
        if module_name is not None:
           if 'processing' in metadata:
              metadata['processing'].update({module_name:default_image_series})
           else:
              metadata['processing']={module_name: module_name:default_image_series
        else:
           metadata['acquisition']={module_name: module_name:default_image_series
        return metadata
```
. Of course the 'description' will need to be saved as an internal attribute and pulled when creating the custom processing module. 
This second method is better as it would require lesser work on the user side.

CodyCBakerPhD:
@Saksham20 I'd be a fan of keeping metadata naming schemes as independent of custom module names as possible; how about the following?

Movie metadata always goes under

`metadata["Behavior"]["Movies"] = [dict(name="Movie1", description="This is the description for Movie1.")]`

This reflects the other modalities as well, such as extracellular electrophysiology going under

`metadata["Ecephys"]["Device"]`, `metadata["Ecephys"]["Electrodes"]`, and so on.

How does that sound?

CodyCBakerPhD:
> > @Saksham20 I'd be a fan of keeping metadata naming schemes as independent of custom module names as possible; how about the following?
> > Movie metadata always goes under
> > `metadata["Behavior"]["Movies"] = [dict(name="Movie1", description="This is the description for Movie1.")]`
> > This reflects the other modalities as well, such as extracellular electrophysiology going under
> > `metadata["Ecephys"]["Device"]`, `metadata["Ecephys"]["Electrodes"]`, and so on.
> > How does that sound?
> 
> You propose we create a `NWBContainer` with the name 'Movies' under processing? OR just have this format in metadata but then store it under acquisition?

I'm saying keep the naming scheme in the metadata separate from underlying functionality in the conversion. _I.e._, the metadata information always goes into the same exact place of the metadata schema no matter which combination of conversion options we happen to specify for the interface.

CodyCBakerPhD:
The closest analog we have to this sort of thing is by considering the `ecephys` metadata again with `BaseRecordingExtractorInterface.run_conversion(...)`.

The ElectricalSeries in that case can be written to (a) acquisition, (b) LFP submodule of processing, (c) FilteredEphys submodule of processing.

We _tend_ to adhere to the following metadata schema for these cases: (a) `ElectricalSeries_raw`, (b) `ElectricalSeries_lfp`, (c) `ElectricalSeries_processed`, but the name of each metadata key is not actually mandated to correspond to the `write_as` location (which can be overridden if necessary) in the conversion options.

Saksham20:
> @Saksham20 I'd be a fan of keeping metadata naming schemes as independent of custom module names as possible; how about the following?
> 
> Movie metadata always goes under
> 
> `metadata["Behavior"]["Movies"] = [dict(name="Movie1", description="This is the description for Movie1.")]`

@CodyCBakerPhD yes I think, for now, this is a reasonable strategy to fix the metadata dict keys under which we store this and only have the user control the module the `ImageSeries` is written using the `module_name`/`module_description` arg to `run_conversion()`. Check this https://github.com/catalystneuro/nwb-conversion-tools/pull/264/commits/7c4f84896a160f61f8efa40bb3fe209ec1707bc6

We could potentially alter this if something better comes out of the broader discussion on the metadata schema. But fo

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -42,6 +47,25 @@ def __init__(self, file_paths: list):
         assert HAVE_OPENCV, INSTALL_MESSAGE
         super().__init__(file_paths=file_paths)
 
+    def get_metadata_schema(self):
+        metadata_schema = super().get_metadata_schema()
+        metadata_schema["properties"]["behavior"] = get_base_schema(tag="behavior")
+        metadata_schema["properties"]["behavior"]["required"] = ["Movies"]
+        metadata_schema["properties"]["behavior"]["properties"] = dict(
+            Movies=dict(type="array", items=get_schema_from_hdmf_class(ImageSeries))
```
CodyCBakerPhD, line 55:
```suggestion
        metadata_schema["properties"]["Behavior"] = get_base_schema(tag="Behavior")
        metadata_schema["properties"]["Behavior"]["required"] = ["Movies"]
        metadata_schema["properties"]["Behavior"]["properties"] = dict(
            Movies=dict(type="array", items=get_schema_from_hdmf_class(ImageSeries))
```
To be consistent with the Ecephys metadata, starting each key with a capital letter

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -42,6 +47,25 @@ def __init__(self, file_paths: list):
...
+        )
+        return metadata_schema
+
+    def get_metadata(self):
+        metadata = super().get_metadata()
+        metadata["behavior"] = dict(
```
CodyCBakerPhD, line 61:
```suggestion
        metadata["Behavior"] = dict(
```

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -98,15 +122,19 @@ def run_conversion(
         else:
             starting_times = [0.0]
 
+        if "behavior" in metadata and "Movies" in metadata["behavior"]:
+            image_series_kwargs_list = metadata["behavior"]["behavior"]
```
CodyCBakerPhD, line 126:
```suggestion
        if "Behavior" in metadata and "Movies" in metadata["Behavior"]:
            image_series_kwargs_list = metadata["Behavior"]["behavior"]
```

CodyCBakerPhD, line 126:
```suggestion
            image_series_kwargs_list = metadata["Behavior"]["Movies"]
```

PR #265, opened 2021-09-16
movie data chunk iterator

Saksham20:
## Motivation

Being consistent with the new datachunking methods, it would be more organized to have one for  a movie also.

If we agree on this, then I can make corresponding changes to the #264 also.


CodyCBakerPhD:
Just redownloaded some of the movie files I originally build this off of, will start testing it later today.

Also, if you could come up with some performance comparisons, I'd just be curious how much faster or slower this is compared to the old iterator.

CodyCBakerPhD:
@Saksham20 Can you please implement the DataChunkIterator into the MovieInterface so we can test it out?

Saksham20:
@CodyCBakerPhD I've implemented and tested it out. It works on my end with the new datachunkiterator.

Could you test it out on your end ?

CodyCBakerPhD:
@Saksham20 Yeah, something isn't working right on the files I'm testing with.

Console spams the top error a few hundred times, stalls for 15 minutes, then terminates in this error leading to the new iterator for this type

```
[NULL @ 000002845aafcfc0] non-existing PPS 0 referenced
[NULL @ 000002845aafcfc0] non-existing PPS 0 refTraceback (most recent call last):

  File "C:\Users\Raven\AppData\Local\Temp/ipykernel_3452/136753418.py", line 1, in <module>
    runfile('C:/Users/Raven/Downloads/test_new_movie.py', wdir='C:/Users/Raven/Downloads')

  File "d:\anaconda3\envs\nwbct\lib\site-packages\debugpy\_vendored\pydevd\_pydev_bundle\pydev_umd.py", line 167, in runfile
    execfile(filename, namespace)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\debugpy\_vendored\pydevd\_pydev_imps\_pydev_execfile.py", line 25, in execfile
    exec(compile(contents + "\n", file, 'exec'), glob, loc)

  File "C:/Users/Raven/Downloads/test_new_movie.py", line 20, in <module>
    converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, conversion_options=conversion_options)

  File "d:\github\nwb-conversion-tools\nwb_conversion_tools\nwbconverter.py", line 182, in run_conversion
    io.write(nwbfile)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\utils.py", line 583, in func_call
    return func(args[0], **pargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 407, in write
    call_docval_func(super().write, kwargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\utils.py", line 424, in call_docval_func
    return func(*fargs, **fkwargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\utils.py", line 583, in func_call
    return func(args[0], **pargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\io.py", line 50, in write
    self.write_builder(f_builder, **kwargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\utils.py", line 583, in func_call
    return func(args[0], **pargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 806, in write_builder
    self.write_group(self.__file, gbldr, **kwargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\utils.py", line 583, in func_call
    return func(args[0], **pargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 995, in write_group
    self.write_group(group, sub_builder, **kwargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\utils.py", line 583, in func_call
    return func(args[0], **pargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 1000, in write_group
    self.write_dataset(group, sub_builder, **kwargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\utils.py", line 583, in func_call
    return func(args[0], **pargs)

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 1294, in write_dataset
    self.__exhaust_dcis()

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 847, in __exhaust_dcis
    if self.__write_chunk__(dset, data):

  File "d:\anaconda3\envs\nwbct\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 1373, in __write_chunk__
    chunk_i = next(data)

  File "d:\github\nwb-conversion-tools\nwb_conversion_tools\utils\genericdatachunkiterator.py", line 128, in __next__
    data = self._get_data(selection=selection)

  File "d:\github\nwb-conversion-tools\nwb_conversion_tools\datainterfaces\behavior\movie\movie_utils.py", line 139, in _get_data
    frames_return.append(frame[selection[1:]])

TypeError: 'NoneType' object is not subscriptable
```

Any ideas?

Saksham20:
@CodyCBakerPhD  could you try saving with `stub_test=True` and check?

Saksham20:
> @Saksham20 I don't know whats going on with this class. `with self`? @CodyCBakerPhD , can you help?

I'm invoking the context manager `__enter__` and `__exit__` within the `__init__` to open and close the movie file

CodyCBakerPhD:
> I'm invoking the context manager `__enter__` and `__exit__` within the `__init__` to open and close the movie file

All we need the context for is file management, not data access. The class should be very sparse and only define what's necessary for opening/closing the file. For all the other data access functions you have attached to the context class, they would then exist separately.
```
with VideoCaptureContext(movie_file, stub=stub) as vc:
                # Get the next frames from vc, which we know in the context is open
# now, exiting the with containment, we know the file is safely closed
```
I also would not set this as an attribute of the `Iterator`, nor would I set the context as a variable in the interface code.

Basically, just like how we use
```
with NWBHDF5IO(filename="my_nwb_file.nwb", mode="r") as io:
    nwbfile = io.read()
```
but for movie files accessed through `open-cv`.

Saksham20:
> All we need the context for is file management, not data access. The class should be very sparse and only define what's necessary for opening/closing the file. For all the other data access functions you have attached to the context class, they would then exist separately.

Yes ,I understand, the basic implementation would give it this functionality. But if we intend to have the data access functions work with the opened file (so not calling `cv2.VideoCapture(path)` all the time), then might has well have them as methods of the class no?


> I also would not set this as an attribute of the `Iterator`, nor would I set the context as a variable in the interface code.

My rationale of putting the context class (which given the current functionality we can rename to something else) we would have to open and close the file each time we want to retrieve information about the frames etc in the `MovieDatachunkIterator`. And since we would often want to read all that information together (frame rate, no of frames, frame dtype, frame shape), retrieving them in one shot and then closing the file is better. Only the other heavy data access like getting the actual frames, timestamps would need the file to be re-opened each time we want to access. 

Do you see any reasons for this class based implementation for being suboptimal to having independent functions? 
I can see some pros and cons:
**Pros**: IMO they're just neatly packed in the class with other iterator like functionality added.
**Cons**: not as simple as previous implementation so harder to maintain/debug in the future?

> Basically, just like how we use
> 
> ```
> with NWBHDF5IO(filename="my_nwb_file.nwb", mode="r") as io:
>     nwbfile = io.read()
> ```
> 
> but for movie files accessed through `open-cv`.

It still works just like this:
```
with VideoCaptureContext("movie/path.avi", stub=True) as vc:
   #retrieve stuff
```

CodyCBakerPhD:
> But if we intend to have the data access functions work with the opened file (so not calling `cv2.VideoCapture(path)` all the time), then might has well have them as methods of the class no?

The original util functions [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/ad6074ba389faf08717d30318338fc5db7100e3f/nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py) are designed to work independently of the interface and iteration. I guess they should more appropriately be called `movie_tools` instead of utils since they do rely on `open-cv` as well. But they are supposed to only require the `file_path` to some movie that the user wants to know more about since not all of these kinds of properties are reported through standard file exploration methods.

By forcing them to be called within the context of the new class we require a lot more knowledge required by the user regarding how to use these, which might be deterring. I'd really prefer a simpler one-shot mechanism for retrieval requiring only a string/path input.


> And since we would often want to read all that information together (frame rate, no of frames, frame dtype, frame shape), retrieving them in one shot and then closing the file is better.

The actual open/close of file context, normally so but also here, is _not expensive_ in the slightest. It's entirely lazy, especially with `open-cv`. So that part of the computation is not the concern.

What's not lazy is the calculation of certain fields that scan the entire file, like the timestamp retrieval. And the user, as per above design intention, may not want all these fields all the times so I'd rather have each function do the bare minimum to retrieve just that piece of info that is requested.


What is improperly done at the current time is manual open/close of the file, for example https://github.com/catalystneuro/nwb-conversion-tools/blob/ad6074ba389faf08717d30318338fc5db7100e3f/nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py#L24-L30

this should instead be

```
    with MyVideoContext(movie_file_path=movie_file) as cap:
        timestamps = [cap.get(cv2.CAP_PROP_POS_MSEC)]
        success, frame = cap.read()
        while success:
            timestamps.append(cap.get(cv2.CAP_PROP_POS_MSEC))
            success, frame = cap.read()
```

and one reason for this (I'm sure there are others) is if `cap.get` or `cap.read` ever error out, the original code would leave the file open whereas in a context, after raising an error, the file will automatically close when exiting the containment.


> It still works just like this:
> ```
> with VideoCaptureContext("movie/path.avi", stub=True) as vc:
>   #retrieve stuff
> ```

It may be able to work that way (but I would address Ben's comments about internal structure of those access functions), but the way it's integrated in the interface and iterator right now is dangerous.

Take for instance

```
video_capture_ob = VideoCaptureContext(str(file), stub=stub_test)

...

dtype = video_capture_ob.get_movie_frame_dtype()
```

The intention may be for the those things to be set as properties of the class (not really a context at that point, though), but that means they are mutable prior to retrieval call, they aren't simply being directly pulled from the source at time of request, so they sit around as an attribute and can be changed anywhere in those `...` lines. And if the property name is one common to other typically used classes in python, it's always possible that something called under the hood during a separate data operation might change it prior to when we want to call and use it ourselves.

Analogously, in the iterator

```
self.video_capture_ob = VideoCaptureContext(movie_file, stub=stub)

...

with self.video_capture_ob as vc:
      frame = vc.get_movie_frame(frame_no)
```

is extremely dangerous, because the point of a proper context call like

> ```
> with VideoCaptureContext("movie/path.avi", stub=True) as vc:
>   #retrieve stuff
> ```

is that you are making a _new instance_ of the I/O class each time, not reusing the same one. This is especially important when some internal of that class is a generator for iteration; if you make an outside instance of the class then pass the same one repeatedly in multiple different context calls, they won't be rebuilding those generators each time. Even if you think the internals are safely designed for that use it still leaves the door open to allow those types of errors to occur, so it's better to just avoid it as a matter of practice.


In general, though, the entire approach to making a brand new class just to do a few things like this, is making a mountain out of a molehill; all that was requested was a simple context manager to the file access so we don't have to make a new cap every time and then manually release it.

I would recommend reverting to commit `2b7da43f694aff9db708cbcdcd9bd7aaf993a857` and making only the smallest changes to wrap file access everywhere needed in that state of the code. No need to re-invent the wheel on all the `movie_tools` access parts.

CodyCBakerPhD:
@Saksham20 Sorry this is so delayed - I think it would help if you factored out the context and util changes into a separate PR that we can deal with first, and then this PR can focus solely on the Iteration part and should be much easier to follow

CodyCBakerPhD:
moved to `restore_movie_data_chunk_iterator`; closing this so we can remove `master`.

PR #268, opened 2021-09-27
generalized get_source_schema() across all interfaces

CodyCBakerPhD:
## Motivation

I noticed that, as of PR #257, all of our BaseDataInterfaces use direct schema generation from class initialization arguments. This, therefore, generalizes it to always apply in place of a base_schema for the abstract DataInterface.

## How to test the behavior?
No behavior change

## Checklist

### DIFF
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -21,8 +21,6 @@ def test_converter():
         nwbfile_path = str(test_dir / "extension_test.nwb")
 
         class NdxEventsInterface(BaseDataInterface):
-            def __init__(self):
-                super().__init__()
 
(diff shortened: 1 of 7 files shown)


bendichter (approved):

tests/test_converter.py
```diff
@@ -21,6 +21,9 @@ def test_converter():
         nwbfile_path = str(test_dir / "extension_test.nwb")
 
         class NdxEventsInterface(BaseDataInterface):
+            def __init__(self):
+                super().__init__()
```
bendichter, line 25: Why do you need to define this here?

CodyCBakerPhD: Thought of a better way to fix it now

PR #269, opened 2021-09-27
get_metadata_schema as class method, with tests

### LINKED ISSUE #261: Metadata schema validation tests
Our testing infrastructure currently validates `source_schema` and `conversion_option` schema, but does not validate the metadata_schema. Note this is different from validating the actual instantiated metadata with its corresponding schema, which is done automatically during the run_conversion step of an NWBConverter.

Not a problem per se, but it does limit coverage. Would be great if we knew for sure that every single `get_metadata_schema()` method passes basic validation.

Tests added to test_interfaces.py similar to the `source_schema` and `conversion_option_schema` ones. Note however that one difference between this schema and the others is that its calling function is not a class method - perhaps making `get_metadata_schema()` a class method would make more sense, not just to help with this but also to be more like the other methods?

## Checklist

CodyCBakerPhD:
fix #261 

## Motivation

Previously, the `get_metadata_schema` methods were instance-specific compared to the other `get_*_schema` methods. This shows what it would be like if they were class methods

Listing as draft until we can discuss some of the specifics on the ophys default value setting. @Saksham20 @luiztauffer would know more about that, or possible @alejoe91...

## How to test the behavior?
Added (and debugged) new tests to ensure all set metadata schemas are proper for interfaces in our `interface_list`

```
pytest
```

## Checklist

### DIFF
--- a/nwb_conversion_tools/basedatainterface.py
+++ b/nwb_conversion_tools/basedatainterface.py
@@ -2,7 +2,7 @@
 from abc import abstractmethod, ABC
 import warnings
 
-from .utils.json_schema import get_base_schema, get_schema_from_method_signature, fill_defaults
+from .utils.json_schema import get_base_schema, get_schema_from_method_signature
 
 
 class BaseDataInterface(ABC):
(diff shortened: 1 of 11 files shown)


CodyCBakerPhD:
Closing since we want to maintain optional capability of automatically adapting to data in the source files.

PR #270, opened 2021-10-04
add option to set nwbfile before calling add_methods

Saksham20:
### Problem:
The ephys writers may want to be used outside of:
1.  Export method [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/135d325b3e674c1474e1dba21c6760fc1def0f2a/nwb_conversion_tools/utils/ephys_writer.py#L36)
2. Using the sorting/recording datainterfaces. 
Both these methods instantiate an nwbfile and require all of sorting/recording extractor information to be written to the nwbfile (so they directly use the `add_to_nwb()` method. 

But there could be cases where we want have a preinstantiated nwbfile and use `add_*()` methods directly to write only specific details into the nwbfile. 

### Motivation: 
The sorging/recording datainterfaces may require to retrieve some info from the sorting/recording [like here](https://github.com/catalystneuro/nwb-conversion-tools/blob/dbb0290346d1e9cf8094cbfa7d72b9e892f0ce4a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py#L82). The `nwbfile`, `metadata` arguments are only required at `run_conversion()` so we should be able to instantiate the ephyswriter class without args: `nwbfile`,`metadata` also. [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/dbb0290346d1e9cf8094cbfa7d72b9e892f0ce4a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py#L37)

### DIFF
--- a/nwb_conversion_tools/utils/basesinwbephyswriter.py
+++ b/nwb_conversion_tools/utils/basesinwbephyswriter.py
@@ -68,7 +68,7 @@ class BaseSINwbEphysWriter(BaseNwbEphysWriter, ABC):
 
     def add_to_nwb(self, nwbfile: pynwb.NWBFile, metadata=None, **kwargs):
         if self.nwbfile is None:
-            self.set_nwbfile(nwbfile, metadata,**kwargs)
+            self.set_nwbfile(nwbfile, metadata, **kwargs)
         if self.waveforms is not None:
             self.add_recording()
             self.add_sorting()
--- a/nwb_conversion_tools/utils/basesinwbephyswriter.py
+++ b/nwb_conversion_tools/utils/basesinwbephyswriter.py
@@ -67,10 +67,8 @@ class BaseSINwbEphysWriter(BaseNwbEphysWriter, ABC):
             self.add_units_waveforms()
 
     def add_to_nwb(self, nwbfile: pynwb.NWBFile, metadata=None, **kwargs):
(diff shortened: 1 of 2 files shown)


nwb_conversion_tools/utils/basesinwbephyswriter.py
```diff
@@ -67,10 +67,8 @@ def add_waveforms(self):
             self.add_units_waveforms()
 
     def add_to_nwb(self, nwbfile: pynwb.NWBFile, metadata=None, **kwargs):
-        assert nwbfile is not None and isinstance(nwbfile, pynwb.NWBFile), "Instantiate an NWBFile and pass as argument"
-        self.metadata = metadata if metadata is not None else dict()
-        self.nwbfile = nwbfile
-        self._conversion_ops = kwargs
+        if self.nwbfile is None:
```
alejoe91, line 70: I think you can avoid passing `nwbfile` here. We can use the `set_nwbfile` before calling this function!

CodyCBakerPhD:
@Saksham20 @alejoe91 This wasn't merged into the previous head before that branch was merged to master; how do you want to proceed on this item?

Saksham20:
lets keeps this as a low priority for now.

PR #272, opened 2021-10-04
Update neuroscope

CodyCBakerPhD:
## Motivation

The Neuroscope interface (especially its metadata) hadn't been fully used since older versions of nwb-conversion-tools. This updates a  couple of things that came up in the new `add_electrodes` functionality.

For cleanliness, I also separated the Neuroscope utility functions from the primary interfaces.

The biggest actual code changes are to the `get_shank_channels` function which previously used a messy try/except approach to automatically pull group structure from keys of the .xml which may or may not exist across different sessions, with the full `anatomicalDescription` always being the backup. Now it always pulls the full info from `anatomicalDescription` and adds a custom column for if it was used in `spikeDetection` based on the optional existence of that field.

Also added docstrings for the interfaces (this takes up about 75 of the added lines to the changelog).

## How to test the behavior?

Eventually, I really should add explicit metadata tests for formats like this. Right now it only tests if run_conversion() actually succeeds but does not look at the structure of the file explicitly beyond the actual data. Probably best to do that in a more general PR, though.

## Checklist

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscope_utils.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscope_utils.py
@@ -1,0 +1,91 @@
+"""Authors: Cody Baker and Ben Dichter."""
+from pathlib import Path
+
+import numpy as np
+from lxml import etree as et
+from spikeextractors import RecordingExtractor
+
+
+def get_xml_file_path(data_file_path: str):
+    """
+    Infer the xml_file_path from the data_file_path (.dat or .eeg).
+
+    Assumes the two are in the same folder and follow the session_id naming convention.
+    """
+    session_path = Path(data_file_path).parent
+    return str(session_path / f"{session_path.stem}.xml")
+
+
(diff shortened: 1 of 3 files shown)


bendichter (approved):

CodyCBakerPhD:
Update: per our meeting, I adjusted the code to re-include the subsetting of channels whenever they are flagged as being in `spikeDetection` as well as the `anatomicalDescription`, in an effort to ensure that by default we only write authentic neural data to the ElectricalSeries.

That said, if the `spikeDetection` values were not set for whatever reason, I think the onus should be upon the user (our conversion scripts mostly) to manually inspect and ensure that channels are proper in this manner, which could easily be included as a step in the `__init__` of the offending datasets NWBConverter.

Anyway, I went ahead and double checked our previous uploads (BTW, DANDI ros3 streaming is amazing, I was able to check the internals of each dataset in under 10 minutes) and it appears as if only Yuta and Grosmark may include some extra channels not described by the probe setup (`i.e.`, likely not shank). All others looked proper in this regard.

PR #274, opened 2021-10-08
GIN as cache

CodyCBakerPhD:
## Motivation

@bendichter An even better solution to #266 is to cache the datasets on GitHub. This is subject to different limits than the artifacts, but offers much better performance. Some of these limits are (a) 5 GB total cap overall caches, with rolling deletion on the oldest and (b) 7 day duration on caches that aren't accessed.

Similar to the artifacts, the idea is instead of manually re-downloading the GIN datasets each time we want to do a full test, do (1) an initial download from GIN (takes about 42 minutes) and save as a cache (takes only a few minutes, see https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts), (2) then simply restore the cached directory and run tests.

With this approach, every day at midnight, the update-dataset workflow will check the current hash of the GIN dataset, and if we haven't downloaded and cached it here it will then do so (takes about an hour until they reduce some of the larger files).

This also simplifies the test_gin script to no longer need datalad through that imported python API (all gets offloaded to the update-dataset workflow), which reduces the amount of code needed by a significant margin.

This also allows the possibility of running full GIN tests on other parts of the OS matrix to ensure our awareness of OS specific issues. The one caveat being windows, which by cache issues [#301](https://github.com/actions/cache/issues/301) and [#379](https://github.com/actions/cache/issues/379) has problems loading the directory where linux and macos do not. Ditto for artifacts, the zip compression library side of things causes things to not work. So until they fix those issues the full tests will really just be for linux and mac.

This type of action can also enable caching of virtual environments, but I've had a bit of trouble getting that working right away so I'll do it in a later PR as a low priority.

## How to test the behavior?

No changes to behavior, other than addition of new actions and overall improved speed.

## Checklist

- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/nwb-conversion-tools/pulls) for the same change? (improvement over #266)

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscope_utils.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscope_utils.py
@@ -1,0 +1,55 @@
+"""Authors: Cody Baker and Ben Dichter."""
+from pathlib import Path
+
+from lxml import etree as et
+
+
+def get_xml_file_path(data_file_path: str):
+    """
+    Infer the xml_file_path from the data_file_path (.dat or .eeg).
+
+    Assumes the two are in the same folder and follow the session_id naming convention.
+    """
+    session_path = Path(data_file_path).parent
+    return str(session_path / f"{session_path.stem}.xml")
+
+
+def get_xml(xml_file_path: str):
+    """Auxiliary function for retrieving root of xml."""
+    return et.parse(xml_file_path).getroot()
+
+
+def safe_find(root: et._Element, key: str, findall: bool = False):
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
@bendichter Actually, this is apparently what they've used on neo for some time now, which is how I heard about it, lol.

samuelgarcia:
This looks cool.
In neo and spikeinterface we already have something similar for caching using the git HEAD as a hash key for cache.
What we don't have is the cron at midnigh, this is a super good idea, I will try to add this in neo and si.
Or maybe Cody if you have time you could push the update-testing-data in SI and neo.

bendichter:
Oh, I see. Very cool. Well it looks good to me!

Issue #277, opened 2021-10-18
increase test coverage


bendichter:
right now, test coverage is reported as 66%. Let's try to bump that up a bit.

bendichter:
It's tricky because brpylib.py doesn't have a license associated with it. Ideally we would ask Blackrock to host it themselves

CodyCBakerPhD:
- [ ] Thorough tests of append mode: https://app.codecov.io/gh/catalystneuro/nwb-conversion-tools/blob/main/nwb_conversion_tools/nwbconverter.py
- [ ] Basic assertion tests for core classes

CodyCBakerPhD:
NeuroConv now has 90% coverage! Thanks @h-mayorquin and @weiglszonja for all your work adding tests!

PR #282, opened 2021-10-19
efficient blackrock headerfile parsing

Saksham20:
## Motivation

this PR helps improve the codecov. #277 
Blackrock (.nsx) files contain some relevant info in the headers that are currently used by the `BlackrockRecordingExtractorInterface` to populate some `NWBFile` metadata like `session_description` and `session_start_time`.
 
Until now, we were using the full fledged `brpylib.py` by blackrock to extract this info.  However, extracting these fields are only a tiny subset of the whole logic in the file. This PR is just the bare minimum code required from brpylib.py for this purpose.

### DIFF
--- a/.github/workflows/full-tests-linux.yml
+++ b/.github/workflows/full-tests-linux.yml
@@ -1,13 +1,11 @@
-name: CI development
+name: Full tests on Linux
 on: pull_request
 jobs:
   run:
-    name: Full tests on (${{ matrix.os }})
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
I mean, I'm a big fan of cleaning this up if you're saying we copied that entire library only to use a handful of functions. I wasn't a part of this project though, so I don't really know the background of why this was done this way - something about licensing I think I heard?

PR #285, opened 2021-10-21
Fix codecov coverage

CodyCBakerPhD:
## Motivation

My apologies, the changes earlier today in PR #284 actually did not entirely fix the problem, which had roots in both calculation and upload. PR #284 fixed upload paths, but introduced a new issue in the calculation of hits/misses over repo.

You can confirm this fix is what we want by viewing the codecov tree through that interface: https://codecov.io/gh/catalystneuro/nwb-conversion-tools/tree/c5b33dd53ba41dceb8b925a131e80cec46206ae1

Now, the mac-os part of the matrix didn't like the explicit paths of PR #284 during codecov calculation, and since the generation of the `.xml` is really only required for linux anyway, I separated them into different workflows with only the linux one doing the codecov stuff. This could also improve badge displays on the main README by having a separate display for the full tests on each system, to advertise support for each OS - since the badge can only point to a workflow produced by a file, not individual jobs that it runs (unless there's a way to do that I don't know about).

On a separate note, analyzing the codecov tree reminds me of a question: do we want to include the /tests themselves under the codecov analysis? Or set them to ignore? @bendichter What do you think?

## Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -15,4 +15,5 @@ coverage:
     patch: off
 
 ignore:
-  - "/tests"
+  - "./tests/"
+  - "./setup.py"


bendichter:
Thanks, @CodyCBakerPhD! We should ignore coverage of /tests

PR #291, opened 2021-11-03
dict deep update fix

Saksham20:
## Motivation

_dict1_: to be updated
_dict2_: used to update dict1
While using `dict_deep_update()`: for a specific key, if the value is a list of dicts, in that case the base _dict1_ value's list is updated such that the element (each is also a dict) of the list that matches the 'name' key of the _dict2_ is completely replaced.

## How to test the behavior?
```python
dict1 = dict(mainname1 [dict(name='name0',description='desc0',location='location0'),dict(name='name1',description='desc1',location='location1')])
dict2 = dict(mainname1=[dict(name='name0',description='desc0 changed',device='device0 changed')])
from nwb_conversion_tools.utils.json_schema import dict_deep_update
d_out = dict_deep_update(dict1,dict2)
# old behavior: 
d_out 
>> {'mainname1': [{'name': 'name0', 'description': 'desc0 changed', 'device': 'device0 changed'}, {'name': 'name1', 'description': 'desc1', 'location': 'location1'}]}
# new behavior: 
d_out 
>> {'mainname1': [{'name': 'name0', 'description': 'desc0 changed', 'location': 'location0', 'device': 'device0 changed'}, {'name': 'name1', 'description': 'desc1', 'location': 'location1'}]}
```
Note how the in the new case, the first dict is updated so the '**location**' key remains. While in the first case (current) its simply replaced so the '**location**' key, value is gone
## Checklist

- [] Have you checked our [Contributing](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/docs/contribute.rst) document?
- [] Have you ensured the PR description clearly describes the problem and solutions?
- [] Have you checked to ensure that there aren't other open or previously closed [Pull Requests]

### DIFF
--- a/nwb_conversion_tools/utils/json_schema.py
+++ b/nwb_conversion_tools/utils/json_schema.py
@@ -33,7 +33,7 @@ def append_replace_dict_in_list(d, ls, k):
         # Index where the value dictionary[k] exists in the list of dicts
         ind = np.where([d[k] == i[k] for i in ls])[0]
         if len(ind) > 0:
-            ls[ind[0]] = d
+            ls[ind[0]].update(d)
         else:
             ls.append(d)
     else:


nwb_conversion_tools/utils/json_schema.py
```diff
@@ -33,7 +33,7 @@ def append_replace_dict_in_list(d, ls, k):
         # Index where the value dictionary[k] exists in the list of dicts
         ind = np.where([d[k] == i[k] for i in ls])[0]
         if len(ind) > 0:
-            ls[ind[0]] = d
+            ls[ind[0]].update(d)
```
bendichter, line 36: should this be a recursive call to `deep_dict_update`?

CodyCBakerPhD: For maximum safety, I would think yes - for cases where that dictionary itself has keys that contain multiple hierarchies

CodyCBakerPhD:
@Saksham20 Sorry, not sure how I missed this PR in my email watchers.

Thanks for catching this. I'd recommend a few things in addition to Ben's suggestion...

1) Similar to how we treat list appending behavior, I'd suggest adding this functionality as an optional argument with the original behavior as the default to maintain back-compatibility with previous uses across our ecosystem, since this is such a core function and doubtless we've been relying on the current behavior in unseen ways.
2) I'd suggest adding extensive tests for this so we can ensure the desired behavior always results; you have a good start for this with the reproducible code posted in the PR description
3) I'd also suggest adding some good documentation describing the effect this new feature has on the dictionaries, probably with easily understood examples (_i.e._, simplified versions of the test functionality and/or reproducible code would be fine)

nwb_conversion_tools/utils/json_schema.py
Saksham20: agreed, I have made changes in the latest commits

nwb_conversion_tools/utils/json_schema.py
```diff
@@ -31,33 +31,50 @@ def append_replace_dict_in_list(d, ls, k):
...
+                ls[idx] = dict_deep_update(ls[idx], d)
         else:
             ls.append(d)
     else:
         ls.append(d)
     return ls
 
 
-def dict_deep_update(d: dict, u: dict, append_list: bool = True, remove_repeats: bool = True) -> dict:
+def dict_get_dtype(d: collections.abc.Mapping, key: str, default_val=None):
```
Saksham20, line 45: if `dict2[key]` is a `dict()` (say) while `dict1[key]` is integer, then this returns an empty dict with the get method. This is useful downstream functions that update dicts and lists.

nwb_conversion_tools/utils/json_schema.py
```diff
@@ -31,33 +31,50 @@ def append_replace_dict_in_list(d, ls, k):
...
+    else:
+        return default_val
+
+
+def dict_deep_update(
+    d: collections.abc.Mapping,
+    u: collections.abc.Mapping,
+    append_list: bool = True,
+    remove_repeats: bool = True,
+    copy: bool = False,
```
Saksham20, line 58: I think its good to have this as an optional arg, since dicts are mutable, its updating may not be desired.

nwb_conversion_tools/utils/json_schema.py
```diff
@@ -31,33 +31,50 @@ def append_replace_dict_in_list(d, ls, k):
...
-                    # add dict only if not repeated
-                    # if not exist_dict_in_list(vv, d.get(k, [])):
-                    # d[k] = d.get(k, []) + [vv]
-            else:
-                d[k] = d.get(k, []) + v
-                # Remove repeated items if they exist
-                if remove_repeats and len(d[k]) > 0:
-                    d[k] = list(set(d[k]))
+            d[k] = dict_get_dtype(d, k, [])
+            for vv in v:
```
Saksham20, line 69: it loops over the elements of list and does not need all the elements to be of the same dtype. A more general case from before, where we assume all elements of the list would be dictionaries

tests/test_json_schema_utils.py
```diff
@@ -8,11 +8,35 @@
...
+    a = sort_dict(a)
+    b = sort_dict(b)
+    assert json.dumps(a, indent=2) == json.dumps(b, indent=2)
 
 
 def compare_dicts_2(a: dict, b: dict):
-    assert json.dumps(a, sort_keys=True) == json.dumps(b, sort_keys=True)
+    a = sort_dict(a)
+    b = sort_dict(b)
+    assert json.dumps(a) == json.dumps(b)
```
Saksham20, line 19: If a dict has a list or other dicts as keys, then `json.dumps(d, sort_keys=True)` does not sort recursively.

CodyCBakerPhD:
@Saksham20 Looking good - I'll think carefully through the tests to see if I can spot any holes in the behavior. Otherwise last thing to do would be to my point #3 above, thorough documentation. The dict_deep_update function now has quite a few arguments ranging quite a few behaviors, would be nice to have both a proper docstring for it so our auto-API can write it to readthedocs - and quite honestly, it would also be nice to have a thorough mini-tutorial written up in markdown showing explicitly how to use them to achieve certain tasks.

CodyCBakerPhD:
@Saksham20 Can you write the docstring for the function? I'd consider that the final thing needed before merge here

nwb_conversion_tools/utils/json_schema.py
```diff
@@ -23,41 +23,64 @@ def exist_dict_in_list(d, ls):
     return any([d == i for i in ls])
 
 
-def append_replace_dict_in_list(d, ls, k):
+def append_replace_dict_in_list(d, ls, k, list_dict_deep_update: bool = True):
```
bendichter, line 26: @Saksham20 If you add an argument like this, you must use the docstring to explain what you are doing

nwb_conversion_tools/utils/json_schema.py
```diff
@@ -23,41 +23,64 @@ def exist_dict_in_list(d, ls):
...
+            for idx in indxs:
+                if list_dict_deep_update:
+                    ls[idx] = dict_deep_update(ls[idx], d)
+                else:
+                    ls[idx] = d
         else:
             ls.append(d)
     else:
         ls.append(d)
     return ls
```
bendichter, line 45: Can you refactor this function to be less indented?

tests/test_json_schema_utils.py
```diff
@@ -8,11 +8,35 @@
...
+
+
+def sort_list(b: list):
+    b.sort(key=str)
+    for bb in b:
+        if isinstance(bb, list):
+            b = sort_list(b)
+        elif isinstance(bb, dict):
+            b = sort_dict(bb)
+    return b
```
bendichter, line 39:
```suggestion
def sort_item(item):
    if isinstance(item, list):
        return sort_list(item)
    elif isinstance(item, dict):
        return sort_dict(item)
    else:
        return item

def sort_list(list_):
    return [sort_item(x) for x in sorted(list_, key=str)]

def sort_dict(dict_):
    return {k: sort_item(dict_[k]) for k in sorted(dict_)}
```

tests/test_json_schema_utils.py
```diff
@@ -8,11 +8,35 @@
 
 
 def compare_dicts(a: dict, b: dict):
-    assert json.dumps(a, sort_keys=True, indent=2) == json.dumps(b, sort_keys=True, indent=2)
+    a = sort_dict(a)
+    b = sort_dict(b)
```
bendichter, line 12:
```suggestion
    a = sort_item(a)
    b = sort_item(b)
```

tests/test_json_schema_utils.py
```diff
@@ -8,11 +8,35 @@
...
+    assert json.dumps(a, indent=2) == json.dumps(b, indent=2)
 
 
 def compare_dicts_2(a: dict, b: dict):
-    assert json.dumps(a, sort_keys=True) == json.dumps(b, sort_keys=True)
+    a = sort_dict(a)
+    b = sort_dict(b)
```
bendichter, line 18:
```suggestion
    a = sort_item(a)
    b = sort_item(b)
```

tests/test_json_schema_utils.py
```diff
@@ -44,22 +68,46 @@ def __init__(self, a: int, b: float, c: Union[Path, str], d: bool, e: str = "hi"
...
+    a4 = dict(a3, ls1=[a1, b1])
+    b4 = dict(b3, ls1=[c1])
+    # compare key is common in both:
+    result4 = dict_deep_update(a4, b4, copy=True, compare_key="a")
+    correct_result = dict(result3_1, ls1=[dict_deep_update(a1, c1, copy=True), b1])
+    compare_dicts(result4, correct_result)
+    # compare key missing:
+    result4 = dict_deep_update(a4, b4, copy=True, compare_key="b")
+    correct_result = dict(result3_1, ls1=[a1, c1, b1])
+    compare_dicts(result4, correct_result)
```
bendichter, line 110: Can you turn these into individual tests and name those tests to indicate specifically what you are testing? If it's too complicate to explain in the name of a function just number them and use docstring to explain.

CodyCBakerPhD:
@Saksham20 Looking great - I made very minor format modifications for `numpy` style docstrings, such as correcting D205/D411 [pydocstyle](http://www.pydocstyle.org/en/stable/error_codes.html) codes. You can catch these yourself if you add docstring linting to your IDE.

Is this good to merge now or are there any last-minute changes you'd like to make?

CodyCBakerPhD (approved):

Saksham20:
> @Saksham20 Looking great - I made very minor format modifications for `numpy` style docstrings, such as correcting D205/D411 [pydocstyle](http://www.pydocstyle.org/en/stable/error_codes.html) codes. You can catch these yourself if you add docstring linting to your IDE.
> 
> If there are any last-minute changes you'd like to make, go ahead - otherwise merge when you're ready.

Its ready from my side. If @bendichter has any comments?

PR #293, opened 2021-11-06
Fix stub conversion factor

CodyCBakerPhD:
## Motivation

Further debug for writing proper conversion factors to an ElectricalSeries run under `stub_test=True` mode. In general, any use of a `SubRecording` or `SubSorting` does not copy properties, including gains, from the parent class.

## How to test the behavior?
I added new tests specifically for this behavior, at least with the Neuroscope interface. I think most others set their gains automatically rather than manually. I can extend it to other interfaces if you want, though.

## Checklist


CodyCBakerPhD:
Unlisting as it's unclear how or if this is necessary in the new writers. Used it to fix the latest Buzsaki conversion, though.

PR #296, opened 2021-11-18
GenericDataChunkIterator improvements

CodyCBakerPhD:
## Motivation

Moderate performance increase and massive code simplification based on suggestions by @bendichter. Pending other suggestions, this will also be the form of the class I'll submit upstream to `hdmf`. It's come a long way from the first versions and now that it's pared down to basics, I can't think of anything left that would increase performance from this side of things.

Performance test on a 957.5s SpikeGLX example file (384 channels) using default chunk/buffer sizes (1MB/1GB, respectively)...

- New method: ~592s
- Previous method (master branch):  ~624s

This raises the speed efficiency to ~161% of relative recording length for Neuropixels, meaning we can write a minute of such data in around ~37s.

Details from top to bottom;

1. Updated default shape setters to return their values rather than set the attribute, to be more consistent in `__init__` with the attribute setting from argument specification. Also changes their names to reflect this. Changes are intended mostly for readability.
2. Updated docstring to emphasize the distinction of chunk (HDF5 handled) vs. buffer (what this class operates with, ideally spans multiple chunks).
3.  Moved assertions to also apply to whatever gets generated from default shape methods; but mathematically, they are guaranteed to pass. Just holding my own methods equally accountable as a user's manual specification.
4. Massive code simplification to selection mapping. Chunks are no longer extracted by us, and the buffer is expressly passed.
5. Also managed to solve the remaining 'to-do' optimization that's been hanging around related to application of the `min` function to safely constrain the upper bound of the buffer slices. I've always known that was a slow down, and knew of a potential way around it using `itertools.chain`, but couldn't figure out the multinomial form of it when applied after the `itertools.product` under the previous method. Instead, refactoring the generator to cover actual lower/upper bounds instead of the permutation indices actually allows the `itertools.chain` to be added prior to `itertools.product`, which saves this complexity. This new method is much more efficient and compacts all details of the mapping process into a single generator created at `__init__`, and that's all that is ever used to perform selections.

## How to test the behavior?
```
pytest
```

Updated `test_gin` to test entire equality of data array rather than just the snippet from inside `check_recordings_equal`. Mostly for verification that the `GenericDataChunkIterator` handles endcaps correctly.

## Checklist

### DIFF
--- a/tests/test_gin.py
+++ b/tests/test_gin.py
@@ -29,7 +29,7 @@ except ImportError:
 #   ecephys: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data
 #   ophys: TODO
 #   icephys: TODO
-LOCAL_PATH = Path("/home/jovyan")  # Must be set to "." for CI - temporarily override for local testing
+LOCAL_PATH = Path(".")  # Must be set to "." for CI - temporarily override for local testing
 DATA_PATH = LOCAL_PATH / "ephy_testing_data"
 HAVE_DATA = DATA_PATH.exists()
 
--- a/tests/test_gin.py
+++ b/tests/test_gin.py
@@ -29,7 +29,7 @@ except ImportError:
 #   ecephys: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data
 #   ophys: TODO
 #   icephys: TODO
-LOCAL_PATH = Path(".")  # Must be set to "." for CI - temporarily override for local testing
+LOCAL_PATH = Path("/home/jovyan")  # Must be set to "." for CI - temporarily override for local testing
 DATA_PATH = LOCAL_PATH / "ephy_testing_data"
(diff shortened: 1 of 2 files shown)


tests/test_gin.py
```diff
@@ -91,7 +92,7 @@ class TestNwbConversions(unittest.TestCase):
 
         @parameterized.expand(parameterized_recording_list)
         def test_convert_recording_extractor_to_nwb(self, recording_interface, interface_kwargs):
-            nwbfile_path = self.savedir / f"{recording_interface.__name__}.nwb"
+            nwbfile_path = str(self.savedir / f"{recording_interface.__name__}.nwb")
```
CodyCBakerPhD, line 95: Worth mentioning; this change doesn't actually effect the CI, but the string casting seems important for running these tests on a local device (may be system specific) with slightly different versions of upstream packages than the CI installs.

bendichter (approved):

PR #298, opened 2021-11-23
relax pip dependencies

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist


.github/workflows/ci-production.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       env:
           NWB_CONVERSION_INSTALL_MODE: production
       run: |
+        pip install -r requirements-dev.txt
```
CodyCBakerPhD, line 25:
```suggestion
      run: |
        pip install -r requirements.txt
```
The ci-production tests are the ones that test the 'laziness' of each of the interfaces that have additional requirements; the `requirements-dev.txt` installs every possible dependency for every possible interface.

setup.py
```diff
@@ -40,7 +30,24 @@
     package_data={"": ["template_metafile.yml"]},
     include_package_data=True,
     python_requires=">=3.7",
-    install_requires=install_requires,
+    install_requires=[
+        "pynwb",
+        "tqdm",
```
CodyCBakerPhD, line 35: IMO this is perhaps a bit _too_ loose of relaxation since there are, for many of these, absolute minimal versions required to do even the most basic things in `nwb-conversion-tools`. An example is schema formation from `hdmf` docvals, which requires a function that wasn't added until a particular version of `hdmf`.

I also prefer to have these extracted from a `requirements-*.txt` file so that we always know internally that requirements files are the ones that need to be updated any time a version bump or new dependency is needed; historically it's easy to forget to also update the setup every time you need to add something.

CodyCBakerPhD:
@bendichter Looks good, made some minor comments and opened a branched PR for some file naming changes; let me know what you think about loose+lazy vs. strict+full install

CodyCBakerPhD:
Also keep in mind it's on the eventual to-do list to remove modality-specific things like spikeextractors/roiextractors to make the minimal install even lazier since people that want to use the repo for `ophys` have no need to ever install `spikeinterface`.

I also just confirmed the sister PR to this one successfully installs on 3.7 as well as 3.9.

setup.py
bendichter: I'd be fine with having these as `requirements-min.txt`. And yes, we should specify minimum requirements for ones where that is needed.

CodyCBakerPhD:
Merging once tests finish, then. Thanks for helping with this

CodyCBakerPhD (approved):

PR #300, opened 2021-11-25
re-adding UnitProperties to metadata schema in SortingExtractorInterface

Saksham20:
## Motivation

The PR merge #230 did not take care of the ability to pass 'UnitProperty' metadata to any `sortingextractorinterface`. It was expecting any custom properties to exist in `self.metadata["units"]=dict(prop1='desc', prop2='desc2')`. The old way of doing this was `self.metadata["Ecephys"]["UnitProperties"]= dict(prop1='desc', prop2='desc2')`.
This PR adds this back. 

## How to test the behavior?
The below, should run with a specific `sortingextractorinterface`
```python
from jsonschema import validate

metadata = sortingextractorinterface.get_metadata()
metadata["Ecephys"] = dict(UnitProperties=[dict(name='prop_name1', description='description1'), 
                                           dict(name='prop_name1', description='description1')])
validate(instance= metadata, schema=sortingextractorinterface.get_metadata_schema()) # this should run fine
sortingextractorinterface.run_conversion(nwbfile, metadata=metadata)
```

## Checklist

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
@@ -28,6 +28,35 @@ class BaseSortingExtractorInterface(BaseDataInterface, ABC):
         self.sorting_extractor = self.SX(**source_data)
         self.writer_class = map_si_object_to_writer(self.sorting_extractor)(self.sorting_extractor)
 
+    def get_metadata_schema(self):
+        """Compile metadata schema for the RecordingExtractor."""
+        metadata_schema = super().get_metadata_schema()
+
+        # Initiate Ecephys metadata
+        metadata_schema["properties"]["Ecephys"] = get_base_schema(tag="Ecephys")
+        metadata_schema["properties"]["Ecephys"]["required"] = []
+        metadata_schema["properties"]["Ecephys"]["properties"] = dict(
+            UnitProperties=dict(
+                type="array",
+                minItems=0,
+                renderForm=False,
+                items={"$ref": "#/properties/Ecephys/properties/definitions/UnitProperties"},
+            ),
+        )
+        # Schema definition for arrays
+        metadata_schema["properties"]["Ecephys"]["properties"]["definitions"] = dict(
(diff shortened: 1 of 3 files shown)


Saksham20:
@CodyCBakerPhD thanks for pointing this out, glad that this was picked up so early after the PR merge!

The reason why this was not captured in tests: the sortingextractorinterfaces that we are testing out in test_gin.py are `BlackrockSortingExtractorInterface` and `PhySortingInterface` [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/15d9625ef0e284fe8f746e3469466a8751c21479/tests/test_gin.py#L112) which do not explicitly set any custom `UnitProperties`. 
Its only `CellExplorerSortingInterface` that does this [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/e64d6f7fdd96ac82d7f5803f3f5a9ed294c9bfa2/nwb_conversion_tools/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py#L26) but this is not in tests (the file does not exist in GIN either, maybe we can upload that?). So the absence of a schema supporting this did throw any error in the metadata validation stage that happens [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/3648b053f1f19995e745938a6e266ea3a649b1a1/nwb_conversion_tools/nwbconverter.py#L108)

CodyCBakerPhD:
@Saksham20 Thanks so much for the quick fix! This is also pertaining to some fixes that might be needed specific to CellExplorer, so I'll branch from here to test out those changes and by so doing, ensure this works as well.

You are 100% right about the GIN data - it's been on the to-do list for some time now to add Neuroscope and CellExplorer Sorting data, but of all the formats CellExplorer is really the one that heavily utilizes UnitProperties.

I think we could add some artificial tests for adding such properties to what we have now until we get some added to GIN, though, I'd be happy to add those in tomorrow.

Saksham20:
@CodyCBakerPhD I have added tests for this feature. I have created a custom `sortingextractordatainterface` for this.

CodyCBakerPhD (approved):

PR #307, opened 2021-11-30
Test python versions

### LINKED ISSUE #305: Add Python 3.8 to testing pipeline
related to #303

CodyCBakerPhD:
fix #305 

## Motivation
Currently, we hard-require Python 3.7 on all devices. This is beginning to be a bit harsh, so investigating how much needs to be done to loosen it.

As seen from the first run, linux works fine on 3.8 but begins running into installation problems in 3.9: https://github.com/catalystneuro/nwb-conversion-tools/runs/4373195033?check_suite_focus=true

Will investigate to see if this can be easily reconciled by basic version bumps on dependencies, or if Python version-specific requirements are the way to go.

## How to test the behavior?
Adjusted main linux workflow to expand a matrix over Python versions. Will adjust others as well pending some investigation into merging codecov reports, since some of our tests are now spanning multiple workflow runs.

## Checklist

### DIFF
--- a/.github/workflows/full-tests-linux.yml
+++ b/.github/workflows/full-tests-linux.yml
@@ -1,5 +1,5 @@
 name: Full tests on Linux
-on: push
+on: pull_request
 jobs:
   run:
     name: Full tests on Linux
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@bendichter I guess the question is, then, if we want to have this run the version matrix on each push of a PR, or have a single version for PRs with the multiple versions run nightly and checked daily to identify version-specific issues?

PR #308, opened 2021-12-01
Add ophys GIN

CodyCBakerPhD:
## Motivation

Adding coverage to all `ophys` interfaces, revealing some bugs that have changed over the months with various API changes.

## How to test the behavior?
Thanks to @Saksham20 fixing the GIN for this data, the caching CI now checks for updates every night along with the ecephys data.

I've then added two simple `Imaging` and `Segmentation` interfaces to the `test_gin.py` so @cechava can begin investigating a fix. Will add all the others we currently support later today.

## Checklist

- [X] Have you checked our [Contributing](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/docs/contribute.rst) document?
- [X] Have you ensured the PR description clearly describes the problem and solutions?
- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/nwb-conversion-tools/pulls) for the same change?

PR #309, opened 2021-12-02
Add ophys GIN (#308)

CodyCBakerPhD:
## Motivation

Adding coverage to all `ophys` interfaces, revealing some bugs that have changed over the months with various API changes.

## How to test the behavior?
Thanks to @Saksham20 fixing the GIN for this data, the caching CI now checks for updates every night along with the ecephys data.

I've then added two simple `Imaging` and `Segmentation` interfaces to the `test_gin.py` so @cechava can begin investigating a fix. Will add all the others we currently support later today.

## Checklist


CodyCBakerPhD:
@cechava OK, it's a bit odd the error traces on the last two commits show what we're after here for your debugging: https://github.com/catalystneuro/nwb-conversion-tools/runs/4391859605?check_suite_focus=true

Hopefully, the future runs that get triggered on this branch load the right cache key, it seems to have different behavior within and outside of an open PR. When in doubt, download the `ophys_testing` dataset locally and do manual debugging on there until the CI issues are resolved.

This looks like it ought to be a pretty straightforward debug and so should serve as a good introduction to how we do schema validation at multiple levels throughout `nwb-conversion-tools`; don't hesitate to ask any questions about anything, though.

tests/test_gin.py
```diff
@@ -138,5 +159,108 @@ class TestConverter(NWBConverter):
...
+            nwbfile_path = str(self.savedir / f"{data_interface.__name__}.nwb")
+
+            class TestConverter(NWBConverter):
+                data_interface_classes = dict(TestImaging=data_interface)
+                def get_metadata(self):
+                    metadata = super().get_metadata()
+                    # attach device to ImagingPlane lacking property
+                    # assume single device and single imaging plane
+                    if 'device' not in metadata['Ophys']['ImagingPlane'][0].keys():
+                        metadata['Ophys']['ImagingPlane'][0]['device'] = metadata['Ophys']['Device']
```
cechava: @CodyCBakerPhD the schema requires the 'device' property an `ImagingPlane` to be a string, assigning the device itself like I do here leads to an error. What's the proper way to do this here? Is there something like softlinks in matnwb that will figure out the path to the device by itself?

CodyCBakerPhD: In metadata, all values are simply strings. In this case, the string name of the device for that object, and hopefully the imaging write function (which should take metadata as an argument) will then handle the actual object referencing in the NWBFile based on that string name.

CodyCBakerPhD: Because we don't have great examples of, and little work done using `ophys` - I'd recommend seeing how it's done in `ecephys` for things we rigorously test already: for example, https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py#L70-L77

cechava: perfect. I was looking for something like this

CodyCBakerPhD:
@cechava Looks like https://github.com/catalystneuro/nwb-conversion-tools/pull/309/commits/4dcdefce6a0eb779ad7ca73a7cb3c50d0a5c8fe5 and https://github.com/catalystneuro/nwb-conversion-tools/pull/309/commits/1f7e0e7faccf92669f618f938ae1e512d13d73c4 added unresolved merge conflicts. Please always resolve such conflicts locally before commiting them as code changes

PR #310, opened 2021-12-04
Make GIN tests run explicitly

CodyCBakerPhD:
## Motivation

In situations where the GIN cache gets wiped or experienced a key mismatch, the current setup of `pytest` simply skips them since it thinks the intention is to run the lazy version of the tests. This essentially means the 'full tests' then do not test on any actual data if this happens, but this failure is a silent skip and the person reviewing the PR has to be careful to notice it.

This change to the infrastructure makes the GIN tests run and fail explicitly in their respective 'full' CI's. The 'partial' (lazy) CI's will run `pytest` the same way a user without any local GIN data should: with `pytest tests/test_internals/` (from root directory of `nwb-conversion-tools`) which therefore bypasses the `test_gin.py`, which would have explicitly failed in the case of not finding local data.

## Checklist


CodyCBakerPhD:
@alejoe91 If you could take a look at this since Ben is busy

.github/workflows/full-tests-linux.yml
```diff
@@ -22,15 +22,22 @@ jobs:
...
       with:
         path: /home/runner/work/nwb-conversion-tools/nwb-conversion-tools/ephy_testing_data
-        key: ecephys-dset-${{ steps.vars.outputs.HASH_EPHY_DATASET }}
-        restore-keys: ecephys-dset-${{ steps.vars.outputs.HASH_EPHY_DATASET }}
-    - name: Run pytest with coverage
+        key: ecephys-dset-${{ steps.ephys.outputs.HASH_EPHY_DATASET }}
+        restore-keys: ecephys-dset-${{ steps.ephys.outputs.HASH_EPHY_DATASET }}
+    - name: Set environment variables
+      run: |
+        echo "TEST_ENV=ci" >> $GITHUB_ENV
```
CodyCBakerPhD, line 36:
```suggestion
    - name: Set environment variables
      run: echo "TEST_ENV=ci" >> $GITHUB_ENV
```
No need to break over multiple lines unless there are multiple commands

.github/workflows/full-tests-linux.yml
```diff
@@ -22,15 +22,22 @@ jobs:
...
+    - name: Show environment variables
+      run: |
+        env
```
CodyCBakerPhD, line 39:
```suggestion
    - name: Show environment variables
      run: env
```

.github/workflows/full-tests-linux.yml
CodyCBakerPhD, line 36: So it looks from https://github.com/catalystneuro/nwb-conversion-tools/runs/4456719722?check_suite_focus=true that the "CI" key being used in `test_gin.py` is inherent to the workflow environment; thus the `TEST_ENV` variable here is going unused?

Do we need this line then?

tests/test_on_data/test_gin.py
```diff
@@ -0,0 +1,156 @@
...
+#   ecephys: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data
+#   ophys: TODO
+#   icephys: TODO
+
+if os.getenv("CI"):
+    LOCAL_PATH = "."
+    print("running on Github CI")
+else:
+    LOCAL_PATH = Path(".")  # Must be set to "." for CI - temporarily override for local testing
+    print("running locally")
```
CodyCBakerPhD, line 40:
```suggestion
if os.getenv("CI"):
    LOCAL_PATH = Path(".")  # Must be set to "." for CI
    print("Running GIN tests on Github CI!")
else:
    LOCAL_PATH = Path(".")  # Override this on personal device for local testing
    print("Running GIN tests locally!")
```
I do wish there was a better way to see these outputs other than sifting through the std callout's but better than nothing.

Thanks for finding this natural environment variable in the GitHub CI; that is a bit nicer for local override convenience!

.github/workflows/full-tests-linux.yml
CodyCBakerPhD, line 39: Same question as above...

> So it looks from https://github.com/catalystneuro/nwb-conversion-tools/runs/4456719722?check_suite_focus=true that the "CI" key being used in test_gin.py is inherent to the workflow environment; thus the TEST_ENV variable here is going unused?
>
> Do we need this line then?

Now that we have the output of https://github.com/catalystneuro/nwb-conversion-tools/runs/4456719722?check_suite_focus=true to display all the built-in environment variables, anyway. Good to have as reference.

tests/test_on_data/test_gin.py
CodyCBakerPhD: @luiztauffer I need to commit this change specifically ASAP because there are a bunch of PRs that branched from this and they're all failing because the CI path is a string, not `PathType`.

.github/workflows/full-tests-linux.yml
luiztauffer: no we don't, we can actually clean everything I added in this file... I don't think i managed to add the ENV here, very frustrating, but at least it seems like the CI key should do it

CodyCBakerPhD: CI stuff is indeed quite frustrating sometimes. You did successfully add the environment variable, though.

image

It's just that there's apparently a whole bunch of other latent environment variables I never knew about, and one of them is a boolean for if this is a CI or not, lol.

PR #312, opened 2021-12-04
Standardize file and folder input arguments

CodyCBakerPhD:
## Motivation

Moving this from the other PR, which had accumulated some issues from countless merge conflicts over the months. Also inherits from #310 to better ensure GIN test pass.

The new annotation method that gives us full control over path types also now allows us to control localized naming conventions rather than having to rely on the underlying package scheme. We therefore have an opportunity to standardize how all our interfaces operate here with respect to input argument naming schemes, which is a nice convenience IMO and came up briefly in our last team meeting.

This PR standardizes all `FilePathType`s to correspond to an input argument called `file_path`. It mostly affects the previous interfaces reliant on `neo` as the extractor, since these instead usually follow the convention of `filename`.

This would introduce a break to back-compatibility with previous versions, so if the convenience of this is agreed upon we can wait to merge until the next big version update (likely the SpikeInterface2 writers + Icephys). Something we could do now if this is approved is to add some deprecation warnings to the affected interfaces to let people know of the upcoming change.

## How to test the behavior?
`test_interfaces` ensures nothing breaks in the schema validation with the name changes; I had to update `test_gin` to properly use the adjusted naming, indicative of the break in back compatability

I also added a test to ensure the aliasing results in the correct `source_data` for each interface.

```
pytest
```

## Checklist


CodyCBakerPhD:
@Saksham20 If you could take a look at this since Ben is busy

nwb_conversion_tools/datainterfaces/ecephys/axona/axonadatainterface.py
```diff
@@ -135,9 +136,10 @@ def get_source_schema(cls):
             type="object",
         )
 
-    def __init__(self, filename: FilePathType, noise_std: float = 3.5):
-        super().__init__(filename=filename)
-        self.recording_extractor = se.AxonaUnitRecordingExtractor(filename=filename, noise_std=noise_std)
+    def __init__(self, file_path: FilePathType, noise_std: float = 3.5):
+        super().__init__(filename=file_path)
+        self.source_data = dict(file_path=file_path)
+        self.recording_extractor = se.AxonaUnitRecordingExtractor(filename=file_path, noise_std=noise_std)
```
Saksham20, line 143:
```suggestion
super().__init__(file_path=file_path)
```

Saksham20 (changes requested):

nwb_conversion_tools/datainterfaces/ecephys/axona/axonadatainterface.py
CodyCBakerPhD: However, looking at this particular subinterface and the way it inherits, it wouldn't work either way regardless. Added a debug for that.

Though, I don't think this interface is tested in any way so it is difficult to tell.

PR #313, opened 2021-12-04
Add Axona GIN tests

CodyCBakerPhD:
## Motivation

Unless I'm mistaken and it got lost in merge conflicts somehow, it looks like Axona data was never added to the GIN tests despite it being readily available.

This PR adds it and reveals some bugs. Assuming @sbuergers is too busy, who else has experience with the Axona format to help debug this?

## Checklist


alejoe91:
@CodyCBakerPhD the problem seems to be related to this:
The `get_metadata()` function of the `AxonaRecordingExtractorInterface` loads the experimenter (if available) from the file header and adds it the metadata as `metadata["NWBFile"]["experimenter"] = ['name of experimenter]`.

However, the `deep_dict_update` in the [`get_metadata` of the `NWBconverter` class](https://github.com/catalystneuro/nwb-conversion-tools/blob/add_axona_gin/nwb_conversion_tools/nwbconverter.py#L96) seems to remove the list notation, so that the field becomes `metadata["NWBFile"]["experimenter"] = 'name of experimenter`, which is not valid in the metadata schema. 
How should we handle this?

For the second error, we were missing an init (just pushed a [fix](https://github.com/catalystneuro/nwb-conversion-tools/pull/313/commits/b267c9a9c1f1e5336da3b63dd55f57d6b93ac4e9)).

nwb_conversion_tools/datainterfaces/ecephys/baselfpextractorinterface.py
```diff
@@ -13,6 +13,9 @@
 class BaseLFPExtractorInterface(BaseRecordingExtractorInterface):
     """Primary class for all LFP data interfaces."""
 
+    def __init__(self, **source_data):
+        super().__init__(**source_data)
```
CodyCBakerPhD, line 17:
```suggestion
    def __init__(self, **source_data):
        super().__init__(**source_data)
```
What does this do that inheriting the `__init__` from the parent doesn't? Note the `AxonaLFPDataInterface` has its own override to the `__init__`, the super should just be needed there instead.

CodyCBakerPhD:
> However, the deep_dict_update in the get_metadata of the NWBconverter class seems to remove the list notation, so that the field becomes metadata["NWBFile"]["experimenter"] = 'name of experimenter, which is not valid in the metadata schema.
How should we handle this?

@Saksham20 Is this intended behavior of the new `deep_dict_update` or another bug?

Saksham20:
> > However, the deep_dict_update in the get_metadata of the NWBconverter class seems to remove the list notation, so that the field becomes metadata["NWBFile"]["experimenter"] = 'name of experimenter, which is not valid in the metadata schema.
> > How should we handle this?
> 
> @Saksham20 Is this intended behavior of the new `deep_dict_update` or another bug?

No, this should be fixed after merging with master which contains #315

CodyCBakerPhD:
> No, this should be fixed after merging with master which contains #315

I've updated it with master, then. @alejoe91 Can you tell if that fixes it on your side?

nwb_conversion_tools/nwbconverter.py
```diff
@@ -93,7 +93,7 @@ def get_metadata(self):
         metadata = get_default_nwbfile_metadata()
         for interface in self.data_interface_objects.values():
             interface_metadata = interface.get_metadata()
-            metadata = dict_deep_update(metadata, interface_metadata)
+            metadata = dict_deep_update(metadata, interface_metadata, append_list=False)
```
CodyCBakerPhD, line 96: If this doesn't fix the problem, can you revert it?

alejoe91 (approved):

PR #316, opened 2021-12-09
Make GIN tests run explicitly

CodyCBakerPhD:
## Motivation

In situations where the GIN cache gets wiped or experienced a key mismatch, the current setup of `pytest` simply skips them since it thinks the intention is to run the lazy version of the tests. This essentially means the 'full tests' then do not test on any actual data if this happens, but this failure is a silent skip and the person reviewing the PR has to be careful to notice it.

This change to the infrastructure makes the GIN tests run and fail explicitly in their respective 'full' CI's. The 'partial' (lazy) CI's will run `pytest` the same way a user without any local GIN data should: with `pytest tests/test_internals/` (from root directory of `nwb-conversion-tools`) which therefore bypasses the `test_gin.py`, which would have explicitly failed in the case of not finding local data.

## Checklist


CodyCBakerPhD:
Cache issue is finally resolved. Still have some ideas for a better way to do it cross-OS, and I'm not thrilled with forcing `datalad` along the same pipeline, but the current limitations of `actions/cache@v2` don't really leave any other options.

alejoe91 (approved):

PR #317, opened 2021-12-09
Added docstrings and minor debug

CodyCBakerPhD:
## Motivation

@bendichter Requested docstrings and minor bug fixes for this for the DataJoint project.

## Checklist


nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -18,12 +30,50 @@ def __init__(
...
     ):
+        """
+        Initialize an Iterable object which returns DataChunks with data and their selections on each iteration.
+
+        Parameters
+        ----------
+        ephys_writer : si.BaseRecording
+          The BaseRecording object (from spikeinterface/spikeinterface) which handles the API for data transfer.
+        segment_index : int, optional
+          DESCRIPTION. The default is 0.
```
CodyCBakerPhD, line 41: @Saksham20 or @alejoe91 Can one of you add the description for this? I have no idea what it does.

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -18,12 +30,50 @@ def __init__(
...
+        unsigned_coercion : list, optional
+          For unsigned data types (e.g., uint16), this value shifts the dtype to a signed type.
+          The default is None.
```
CodyCBakerPhD, line 44: @alejoe91 Does the new spikeinterface object not have an offset field? If so, why are we asking for it to be set explicitly?

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -18,12 +30,50 @@ def __init__(
...
+        write_scaled : bool, optional
+          Whether to scale the data by the internal conversion factor ('gain', in spikeinterface).
+          The default is True.
```
CodyCBakerPhD, line 47: @Saksham20 Why is the default here `True`? For writing to NWB, `False` should _always_ be the default for applying the conversion factor.

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -18,12 +30,50 @@ def __init__(
...
+          The chunk_shape will be set implicitly by this argument.
+          The default is 1MB, which is highly recommended by the HDF5 group. For more details, see
+          https://support.hdfgroup.org/HDF5/doc/TechNotes/TechNote-HDF5-ImprovingIOPerformanceCompressedDatasets.pdf
+        chunk_shape : tuple, optional
+          If the user desires further control over the shaping of the buffer over all the axes,
+          they may set this explicitly as a tuple of the same length as the data shape indicating the desired shape
+          of each selection.
+          The default is None.
+        """
+        assert HAVE_SI_090, "spikeinterface v0.9 is not installed (pip install spikeinterface)!"
```
CodyCBakerPhD, line 68: @alejoe91 Is this lazy import message OK?

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -1,15 +1,27 @@
...
+
 
 class NwbEphysWriterDataChunkIterator(GenericDataChunkIterator):
-    """DataChunkIterator specifically for use on RecordingExtractor objects."""
+    """DataChunkIterator for use on RecordingExtractor (spikeinterface/spikeinterface) objects."""
 
     def __init__(
         self,
-        ephys_writer,
+        ephys_writer: si.BaseRecording,
```
CodyCBakerPhD, line 24: Ben also proposed renaming this more appropriately to `ephys_reader`, what do you guys thing?

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
alejoe91: It does. I think that we should set `unsigned_coercion` only if the dtype of the raw data is unsigned. What do you think?

Saksham20:
@CodyCBakerPhD 
After implementing the `unsigned_coercion` correction, the tests are failing for the intan data.
It seems the tests for all the other datasets are passing because their `channel_offeset` values are 0, which cause the `unsigned_coercion` to be zero thus not resulting in any correction in the datachunkiterator [here] (https://github.com/catalystneuro/nwb-conversion-tools/blob/260637196f204e2173a44790b0278cadc3059b47/nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py#L82) before writing to nwb. But not in the case of intan which have non-zero `channel_offset` values thus non-zero `unsigned_coercion`. 
Can you take a look by running this:
```python
nwbfile_path = 'nwbfile.nwb'
# download from [GIN](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/intan/intan_rhd_test_1.rhd)
interface_kwargs=dict(file_path='path-to-intan') 

class TestConverter(NWBConverter):
    data_interface_classes = dict(TestRecording=IntanRecordingInterface)

converter = TestConverter(source_data=dict(TestRecording=dict(interface_kwargs)))
converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True)
recording = converter.data_interface_objects["TestRecording"].recording_extractor
nwb_recording = NwbRecordingExtractor(file_path=nwbfile_path)

traces_recording = recording.get_traces(return_scaled=False)
traces_nwb = nwb_recording.get_traces(return_scaled=False)
traces_recording[:10]-traces_nwb[:10] # out: [32768, 32768, 32768 ... ]
```

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
CodyCBakerPhD: I would agree, and we should always check the offset field before that. Basically, the offset can serve two purposes

1) shifting the values of the series to some scientific unit
2) shifting the data type from unsigned to signed

since we don't know which is the use case ahead of time, but we do know the object has internally set offsets, we can start by using those. In pseudo-code

```
shifted_traces=unsigned_traces + internally set offset
if dtype(shifted_traces) are still unsigned:
    signed_traces = shifted_traces - 2**(n_bits-1)  # forces to be signed
```

where `n_bits=16` for `int16`, etc. 

Note that all this will only be an issue until the offset field becomes built-in for the PyNWB containers; once it is we can just set it there and write the series as is.

alejoe91: Another problem is that when we apply the `unsigned_coercion` we are actually removing the offset, but the offset property is saved when adding electrodes...Also currently, the `NwbEphysWriterDataChunkIterator` returns an unsigned dtype even when an `unsigned_coercion` is used. In that case, the `_get_dtype` function should return a signed dtype. Right?

CodyCBakerPhD: > offset property is saved when adding electrodes
We can probably stop doing that until we add it as a specific attribute assigned to datasets (long-term plan); it's enough now to simply take an unsigned type and make it a signed type.

> Also currently, the NwbEphysWriterDataChunkIterator returns an unsigned dtype even when an unsigned_coercion is used. In that case, the _get_dtype function should return a signed dtype. Right?
The way I'd see it here is in that case, we have the `dtype` that goes into the NWB conversion process (say, `uint16`), but the `dtype` we want to save it as in the NWBFile would be the signed version (say, `int16`). The only added cost is to doing the addition/subtraction at each buffer read, but from personal experiments on integer data types that isn't too costly an operation.

CodyCBakerPhD: > Another problem is that when we apply the unsigned_coercion we are actually removing the offset,

The way I have it outlined it above, the information as an attribute goes away but the information is instead absorbed into the series itself WLOG; even the `itemsize` is actually preserved except in an edge case where there was (a) the maximum value of the data itself was on the right edge of the unsigned range and (b) the offset was a non-zero integer.

Or I guess if the offset itself is a float is kind of the worst case scenario, but I've never seen that before. You can maybe add a prohibition of that until we can set the `offsett` as an NWB attribute.

alejoe91: I'm working on a solution! :) Seems to work locally but it needs some more testing. I'll push soon

alejoe91:
@CodyCBakerPhD @Saksham20 

I tried to solve the problem this way: when `unsigned_coercion` is used, we have to force the dtype at the DataChunkIterator level (e.g. `uint16` --> `int16`). In this case, we don't have to write offsets as columns to the elctrode table, because they are removed. So the addiiton of offsets and gains to the electrode table is now in the `add_electrical_series`.

An extra problem arises in testing formats with unsigned values (Intan). The original file will have unscaled `uint` traces, the resulting NWB file will have `int` types. For this reason, I added a `test_unscaled` argument to the `test_convert_recording_extractor_to_nwb` function. 

Let me know what you think!

CodyCBakerPhD:
@alejoe91 Thanks for working on this; I'll add some suggestions and final touches here and there.

> I tried to solve the problem this way: when unsigned_coercion is used, we have to force the dtype at the DataChunkIterator level (e.g. uint16 --> int16). In this case, we don't have to write offsets as columns to the elctrode table, because they are removed. So the addiiton of offsets and gains to the electrode table is now in the add_electrical_series.

That all sounds fine.

> An extra problem arises in testing formats with unsigned values (Intan). The original file will have unscaled uint traces, the resulting NWB file will have int types. For this reason, I added a test_unscaled argument to the test_convert_recording_extractor_to_nwb function.

I don't think blocking the test is the way to go. As I recall, there should be a way to call the signed `dtype` through that `RecordingExtractor` object (since I recall adding this to `pyintan` at some point), so the added argument to the tests should just control the `dtype` that is tested for unscaled step. I can mess with that, though.

nwb_conversion_tools/utils/basenwbephyswriter.py
```diff
@@ -412,43 +420,88 @@ def add_electrical_series(self, segment_index):
...
+                        unsigned_coercion = unsigned_coercion.astype(int)
+                    unsigned_coercion = list(unsigned_coercion)
+
+                    # in this case we have to force the uint type to int
+                    dtype_str = recording_dtype.str
+                    # to avoid overflow, we need to double the number of bytes
+                    nbits = int(dtype_str[dtype_str.find("u") + 1 :])
+                    new_dtype_str = f"{dtype_str[0]}i{2*nbits}"
+                    data_chunk_dtype = np.dtype(new_dtype_str)
+
```
CodyCBakerPhD, line 457: This seems like a lot of untested internal logic/machinery for handling `dtype` promotion; I'll investigate established ways of doing it.

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -1,48 +1,98 @@
...
+        ephys_writer: BaseNwbEphysWriter,
         segment_index: int = 0,
         unsigned_coercion: list = None,
-        write_scaled: bool = True,
+        write_scaled: bool = False,
         buffer_gb: float = None,
         buffer_shape: tuple = None,
         chunk_mb: float = None,
         chunk_shape: tuple = None,
+        dtype: Union[str, np.dtype] = None,
```
CodyCBakerPhD, line 24: The only thing about adding this new argument is now, I really feel like this function is getting overloaded. Especially considering the interactions between `unsigned_coercion`, `write_scaled`, and `dtype`; these aren't being fully tested in that capacity and at various levels may be redundant.

I understand we will control these ourselves in specified ways at the `add_electrical_series` call, but this Iterator is also intended to be capable of standalone usage.

I'll do a separate PR into this that tries to reduce what's needed to get to the bare minimum of required information.

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -1,48 +1,98 @@
...
+            end_frame=selection[0].stop,
+            return_scaled=self.write_scaled,
+            segment_index=self.segment_index,
         )
 
+        if self.dtype is not None:
+            data = data.astype(self.dtype)
+
+        data = data + self.unsigned_coercion[channel_idxs]
+        return data
```
CodyCBakerPhD, line 89: Note that it's possible, the way it's currently setup, for the modification by the coercion to promote the `dtype` in a way that wouldn't return the explicitly set `dtype`. This links to the redundancy problem mentioned above when using this Iterator in a really arbitrary way.

I.e., `NwbEphysWriterDataChunkIterator(ephys_writer=my_writer_with_unit8_data, dtype="int8", unsigned_coercion=[-(2**15+1)]*len(series), return_scaled=True)` is a problem because what gets returned at each buffer is none of those things requested by the arguments.

alejoe91:
> > An extra problem arises in testing formats with unsigned values (Intan). The original file will have unscaled uint traces, the resulting NWB file will have int types. For this reason, I added a test_unscaled argument to the test_convert_recording_extractor_to_nwb function.
> 
> I don't think blocking the test is the way to go. As I recall, there should be a way to call the signed `dtype` through that `RecordingExtractor` object (since I recall adding this to `pyintan` at some point), so the added argument to the tests should just control the `dtype` that is tested for unscaled step. I can mess with that, though.

@CodyCBakerPhD I think that for Intan or any other unsigned type recordings we shouldn't test the unscaled option simply because we are actively removing the "unsignedness" when saving to NWB. So currently, there is no general way to test the `return_scaled=False` (Intan is an Exception, but the `dtype` mechanism in `pyintan` is also kind of a trick to make it work IMO ;) )

CodyCBakerPhD:
We'll continue thinking about technical discussions brought up in this PR, but moving them elsewhere and closing for now.

PR #319, opened 2021-12-13
Progress bars for recording iterators

CodyCBakerPhD:
## Motivation

As a part of the DJ project, @bendichter requested built-in progress bars for the various inheritors of `GenericDataChunkIterators` with `tqdm` so one can visualize the rate of conversion speed for those hefty data. The `MovieDataInterface` with non-external file mode has always supported this.

I doubt the `hdmf` would want that added to the core upstream version of this since it would require yet another addition to their minimal requirements, so I just added it into our own locally defined instances.

Anyway, this solution invokes a progress bar in both the classical iterator context
```
for buffer in RecordingExtractorDataChunkIterator(recording=my_recording):
    print(buffer.shape)
```
as well as in our primary usage
```
class MyConverter(NWBConverter):
    data_interface_classes = dict(SpikeGLX=SpikeGLXRecordingInterface)
my_converter = MyConverter(source_data=dict(SpikeGLX=dict(file_path=".../my_recording_file.ap.bin")))
my_converter.run_conversion(nwbfile_path=".../my_nwbfile.nwb")
```

## How to test the behavior?
I have no idea how to test `tqdm` output to a console. Anyone have any ideas? 

## Checklist

### DIFF
--- a/nwb_conversion_tools/utils/genericdatachunkiterator.py
+++ b/nwb_conversion_tools/utils/genericdatachunkiterator.py
@@ -102,7 +102,7 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
             self.progress_bar_options = progress_bar_options
         else:
             self.progress_bar_options = dict()
-        progress_bar_defaults = dict(disable=True, position=0, leave=False)
+        progress_bar_defaults = dict(disable=True, leave=False)
         for default_key, default_value in progress_bar_defaults.items():
             if default_key not in self.progress_bar_options:
                 self.progress_bar_options.update({default_key: default_value})
@@ -147,7 +147,6 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
             try:
                 from tqdm import tqdm
 
-                print(self.progress_bar_options)
                 self.progress_bar = tqdm(total=self.num_buffers, **self.progress_bar_options)
             except ImportError:
                 warn(
--- a/nwb_conversion_tools/utils/genericdatachunkiterator.py
+++ b/nwb_conversion_tools/utils/genericdatachunkiterator.py
@@ -1,9 +1,11 @@
 """Authors: Cody Baker, Saksham Sharda, and Oliver Ruebel."""
+import sys
 from typing import Iterable, Tuple, Optional
 import numpy as np
 import psutil
 from abc import abstractmethod
 from itertools import product, chain
+from tqdm import tqdm
 
 from hdmf.data_utils import AbstractDataChunkIterator, DataChunk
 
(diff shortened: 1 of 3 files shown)


nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -17,15 +19,19 @@ def __init__(
         buffer_shape: tuple = None,
         chunk_mb: float = None,
         chunk_shape: tuple = None,
+        display_progress: bool = True,
```
bendichter, line 22:
```suggestion
        display_progress: bool = False,
```

nwb_conversion_tools/utils/genericdatachunkiterator.py
```diff
@@ -1,9 +1,11 @@
 """Authors: Cody Baker, Saksham Sharda, and Oliver Ruebel."""
+import sys
 from typing import Iterable, Tuple, Optional
 import numpy as np
 import psutil
 from abc import abstractmethod
 from itertools import product, chain
+from tqdm import tqdm
```
bendichter, line 8:
```suggestion
```

nwb_conversion_tools/utils/nwbephyswriterdatachunkiterator.py
```diff
@@ -17,15 +19,19 @@ def __init__(
...
         self.write_scaled = write_scaled
         self.ephys_writer = ephys_writer
+        self.display_progress = display_progress
         self.channel_ids = list(ephys_writer.recording.get_channel_ids())
         self.unsigned_coercion = (
             np.zeros([len(self.channel_ids)]) if unsigned_coercion is None else np.array(unsigned_coercion)
         )
         super().__init__(buffer_gb=buffer_gb, buffer_shape=buffer_shape, chunk_mb=chunk_mb, chunk_shape=chunk_shape)
+        if self.display_progress:
+            self.progress_bar = tqdm(total=self.num_buffers, position=0, leave=False)
```
bendichter, line 34:
```suggestion
            from tqdm import tqdm
            self.progress_bar = tqdm(total=self.num_buffers, position=0, leave=False)
```

bendichter (changes requested):
If you do it like this then we can add this to hdmf without requiring them to have tqdm as a requirement. What do you think?

Also, could you add a way to pass `desc` to tqdm?

CodyCBakerPhD:
Testing out moving the code into the Generic level; I do think this is the best level since it would universally apply to any use of the iterator. And it's true, if the default is to disable the progress bar then we could have it as a lazy import in hdmf, which would be the way to go.

My recent changes to the method of doing this insofar as the Generic level is concerned now allows full argument passing, including the requested `desc` field, but also anything other than the `total` counter which must be calculated by the iterator itself. That way a user can control other properties of the display such as `position` (relevant if more than one iterator) or more complicated format appearances (you can pass f-strings for example to control the display). `tqdm` allows an awful lot of such customizable options so I'm only overriding the very basic stuff here and allowing user full control through the `progress_par_options` field.

This currently works well for this example:
```
for buffer in RecordingExtractorDataChunkIterator(
    recording=recording,
    progress_bar_options=dict(
        disable=False,
        desc="Writing SpikeGLX recording..."
    )
):
    sleep(0.25)
```
However, getting it integrated into the new interface workflow is proving quite challenging. I'll keep working on it but it will take a while.

CodyCBakerPhD:
All changes here propagated up to `hdmf`.

PR #327, opened 2021-12-17
Adding reading matlab new format capabilities to the CellExplorerInterface

### LINKED ISSUE #318: Extend CellExplorerInterface to MATLAB v7.3+
This would be a good first upstream PR from the Buzsaki project for @h-mayorquin, since he's already looked into it a bit.

Basically, the current `CellExplorerSortingInterface` only strictly supports reading and data retrieval from MATLAB files through `scipy.io.loadmat`.

This can be extended to the other MATLAB file versions with `hdf5storage.loadmat`. It will change the way data is retrieved from many of the keys, however, so I've added a task list to get started.

- [x] Find two good example `.mat` files for this task; one that is readable in `scipy`, one that is not. We've already found some like this in Tingley, just need to report exact session IDs here for clarity.
- [ ] The neo team currently has a PR on GIN for `scipy` readable CellExplorer files; when you find a good example of a `hdf5storage` one, send it my way and I'll stub it and send it their way.
- [ ] Add GIN tests for the CellExplorerSortingInterface acting on the `scipy` readable files.
    -  [ ] When the `hdf5storage` files are added after stubbing, add tests for them as well.
- [x] Make a spikeextractors PR for the `sessionInfo.mat` file access.

For later, when we find a `.cellinfo` file with these properties...
- [ ] At first attempt to load the file in the Interface, do a `try/except` to determine the file version and track the result through an attribute, _e.g._, `self.matlab_file_version`.
- [ ] Based on the result of that attribute, for each data retrieval the interface performs, find the analog that perfectly matches it for the `hdf5storage` version.
    - Technically, this would be way more efficient if we made a specialized reader class for all MATLAB files that always handled data access from all source files through the same API, but this is a tad complicated for a first PR. We can always refactor it a bit later to reduce code.

As always, do not hesitate to reach out if you ever have a question, no matter how small.

## Checklist

h-mayorquin:
fix #318  

## Motivation
Newer versions of matlab can only be opened with `hdf5storage` instead of `scipy.io`. The purpose of this pull is to allow `CellExplorerInterface` to read the two types of formats. Moreover,  an optional file_path for both `sessionInfo` and `spikes.Info` files will be added. 

## How to test the behavior?
This PR will include tests for both formats as soon as they are available on gin.

## Checklist

### DIFF
--- a/tests/test_on_data/test_gin.py
+++ b/tests/test_on_data/test_gin.py
@@ -3,6 +3,7 @@ import unittest
 from pathlib import Path
 import numpy.testing as npt
 import os
+from nwb_conversion_tools.datainterfaces.ecephys.cellexplorer.cellexplorerdatainterface import CellExplorerSortingInterface
 
 import pytest
 from spikeextractors import NwbRecordingExtractor, NwbSortingExtractor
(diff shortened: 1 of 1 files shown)


h-mayorquin:
I found a spikes.info file in new format (that is unreadable by scipy). The files belong to the repository corresponding to the cell explorer paper and can be found in the globus data system:
`/PetersenP/CellExplorerExampleData/MS22/Peter_MS22_180629_110319_concat/` 
@CodyCBakerPhD

PR #328, opened 2021-12-20
Run NWB Conversion from YAML specification

CodyCBakerPhD:
## Motivation

This PR introduces the most straightforward way of writing a `.yml` file that specifies all information needed to run a typical conversion using non-custom `DataInterfaces`.

The `.yml` language allows three levels for custom information (both metadata and `DataInterface` specifications) to be specified: (i) on the outermost level, the information applies to all sessions of all experiments, (ii) at the experiment level, only applies to sessions within that experiment nesting, (iii) specific metadata to only a particular session.

This PR does require some of the `utils.conversion_tools` functions to be split into a `utils.nwbfile_tools` to avoid circular import with the new function.

A separate follow-up PR will extend the `.yml` language to allow reverse f-string glob specification whenever their dataset is hierarchically structured, possibly using keywords that can also serve as NWBFile metadata such as `session_id`, `subject_id`, and so on. This approach will greatly reduce redundancy in the `.yml` language for datasets that rigorously adhere to the specified structure, but will not be so rigorous that it cannot support custom sessions outside this structure (which is mainly what this current PR resolves).

## How to test the behavior?
A testing `.yml` file is included, which uses actual GIN data in its construction. The parsing function runs on this `.yml` to produce an `.nwb` file, whose internal structure is then verified to match the intended specification.

## Checklist


CodyCBakerPhD:
Still remembered some to-do's:

- [ ] Define JSON schema (with version) for the YAML specification
- [ ] Add validation of `.yml` file against this schema

PR #332, opened 2021-12-28
Add YAML conversion specification schema

CodyCBakerPhD:
## Motivation

This is a pre-PR for reducing the changelog for PR #328 that defines the JSON schema for the YAML conversion specification feature and generalizes some schema-related tools and tests in the process.

## How to test the behavior?

Separated the `Draft7Validator` tests to a separate test file, since they are largely independent of the `test_interfaces.py` file they were previously within.

Added test to ensure this new schema is indeed a valid `Draft7` version schema definition.

## Checklist


nwb_conversion_tools/schemas/yaml_specification_schema.json
```diff
@@ -0,0 +1,91 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "yaml.specification.schema.json",
+  "title": "YAML specification schema",
```
bendichter, line 4: Can you make these more specific?

nwb_conversion_tools/schemas/yaml_specification_schema.json
```diff
@@ -0,0 +1,91 @@
...
+      "additionalProperties": true
+    },
+
+    "data_interfaces": {
+      "type": "array",
+      "items": {
+        "title": "DataInterface",
+        "type": "string"
+      }
+    },
```
bendichter, line 27: These are defined multiple places. Consider referencing a definition instead.

nwb_conversion_tools/schemas/yaml_specification_schema.json
```diff
@@ -0,0 +1,91 @@
...
+              "type": "string"
+            }
+          },
+
+          "sessions": {
+            "type": "array",
+            "items": {
+              "title": "Single-session specification",
+              "type": "object",
+              "required": ["nwbfile_name", "source_data"],
```
bendichter, line 57: I don't think nwbfile_name should be required. Let's just automatically use the DANDI convention

bendichter:
What about something like this?

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "yaml.specification.schema.json",
  "title": "YAML specification schema",
  "description": "Schema for a YAML specified conversion.",
  "version": "0.1.0",
  "type": "object",
  "required": ["experiments"],
  "allOf": [{"$ref":  "#/$defs/level"}],
  "properties": {
    "experiments": {
      "type": "object",
      "additionalProperties": {
        "type": "object",
        "required": ["sessions"],
        "allOf": [{"$ref":  "#/$defs/level"}],
        "properties": {
          "sessions": {
            "type": "array",
            "items": {
              "title": "Single-session specification",
              "type": "object",
              "required": ["nwbfile_name", "source_data"],
              "allOf": [{"$ref":  "#/$defs/level"}],
              "properties": {
                "nwbfile_name": {
                  "type": "string"
                },
                "source_data": {
                  "type": "object",
                  "additionalProperties": true
                }
              }
            }
          }
        }
      }
    }
  },
  "$defs": {
    "level": {
      "metadata": {
        "$ref": "./metadata_schema.json#"
      },

      "conversion_options": {
        "type": "object",
        "additionalProperties": true
      },

      "data_interfaces": {
        "type": "array",
        "items": {
          "title": "DataInterface",
          "type": "string"
        }
      }
    }
  }
}
```

nwb_conversion_tools/schemas/yaml_specification_schema.json
CodyCBakerPhD: Yep, `allOf` with the definition works great.

nwb_conversion_tools/schemas/yaml_specification_schema.json
CodyCBakerPhD: I see how non-specific that is; changed all references to this as 'YAML conversion specification' to be clearer.

CodyCBakerPhD:
@bendichter Thanks for the comments; I was about to ask if you knew of a good way to reduce that repeating duplication of the metadata/options stuff, the `allOf` utility is exactly what we wanted there.

With some minor tweaks this is validating against the example file we later use for the full GIN tests for this feature: https://github.com/catalystneuro/nwb-conversion-tools/pull/334

nwb_conversion_tools/schemas/yaml_specification_schema.json
CodyCBakerPhD: Agreed

PR #344, opened 2022-01-03
Run NWB Conversion from YAML specification

CodyCBakerPhD:
## Motivation

This PR introduces the most straightforward way of writing a `.yml` file that specifies all information needed to run a typical conversion using non-custom `DataInterfaces`.

The `.yml` language allows three levels for custom information (both metadata and `DataInterface` specifications) to be specified: (i) on the outermost level, the information applies to all sessions of all experiments, (ii) at the experiment level, only applies to sessions within that experiment nesting, (iii) specific metadata to only a particular session.

This PR does require some of the `utils.conversion_tools` functions to be split into a `utils.nwbfile_tools` to avoid circular import with the new function.

A separate follow-up PR will extend the `.yml` language to allow reverse f-string glob specification whenever their dataset is hierarchically structured, possibly using keywords that can also serve as NWBFile metadata such as `session_id`, `subject_id`, and so on. This approach will greatly reduce redundancy in the `.yml` language for datasets that rigorously adhere to the specified structure, but will not be so rigorous that it cannot support custom sessions outside this structure (which is mainly what this current PR resolves).

## How to test the behavior?
Several testing `.yml` files are included, which use actual GIN data in their construction. The parsing function runs on these `.yml` files to produce `.nwb` files, whose internal structure is then verified to match the intended specification.

## Checklist


CodyCBakerPhD:
To-Do from pre-PR: use DANDI default naming scheme for optional `nwbfile_name`

nwb_conversion_tools/utils/conversion_tools.py
```diff
@@ -1,67 +1,109 @@
...
+            metadata = converter.get_metadata()
+            for metadata_source in [global_metadata, experiment_metadata, session.get("metadata", dict())]:
+                metadata = dict_deep_update(metadata, metadata_source)
 
-def check_regular_timestamps(ts):
-    """Check whether rate should be used instead of timestamps."""
-    time_tol_decimals = 9
-    uniq_diff_ts = np.unique(np.diff(ts).round(decimals=time_tol_decimals))
-    return len(uniq_diff_ts) == 1
+            print(metadata["NWBFile"])
```
bendichter, line 93: did you want to keep this?

```suggestion
```

nwb_conversion_tools/utils/conversion_tools.py
```diff
@@ -1,67 +1,109 @@
...
+            if "nwbfile_name" not in session:
+                # 'subject_id' is required by schema validation if the 'Subject' is specified in metadata
+                assert "Subject" in metadata and "session_start_time" in metadata.get("NWBFile"), (
+                    "If not specifying an explicit name for the NWBFile ('nwbfile_name'), then both "
+                    "metadata['Subject']['subject_id'] and metadata['NWBFile']['session_start_time'] are required!"
```
bendichter, line 98: why does session_start_time have to do with the naming?

nwb_conversion_tools/utils/conversion_tools.py
```diff
@@ -1,67 +1,109 @@
...
+                )
+                subject_file_name = metadata["Subject"]["subject_id"].replace(" ", "_")
+                nwbfile_name = f"{subject_file_name}_{metadata['NWBFile']['session_start_time']}"
```
bendichter, line 101: could we use the dandi naming code here instead? https://github.com/dandi/dandi-cli/blob/master/dandi/organize.py#L70

nwb_conversion_tools/utils/conversion_tools.py
CodyCBakerPhD: The datetime shows up as a part of the DANDI structure of naming to specify session IDs; I agree though that the actual DANDI functions ought to be used

nwb_conversion_tools/utils/conversion_tools.py
CodyCBakerPhD: Yep, good idea

CodyCBakerPhD:
Returning to draft until I figure out the DANDI organization integration

CodyCBakerPhD:
@bendichter Usage of DANDI `organize` convention for establishing unique filenames is now established.

For any session in the `.yml` file that does not have an `nwbfile_name` set, it is temporarily assigned an alias `temp_nwbfile_name_{#}.nwb`; this is because DANDI `organize` finds the minimal amount of metadata to include in the file name to uniquely distinguish it from all the other NWBFiles in the directory, meaning it can't really be resolved at the level of individual files.

After all the NWBFiles are written, it then triggers a renaming only for the temporary aliases (anything with an `nwbfile_name` pre-specified stays as-is) based on the organization convention, but considering all the metadata across all the files for maximum distinction.

Right now, if there isn't enough metadata for DANDI to make an intelligent decision, it raises an `assertion` prior to renaming, but keeps the written NWBFile with the temporary alias to avoid losing the data in the file. The user can then manually rename the file if they wish. This is mostly triggered by the special case of using the `.yml` to convert only a single session.

For more than a single session, there will usually be enough for the organize to distinguish based on `session_start_time` alone, which sends us back to the age-old question of **should we set `session_start_time` to be a required metadata field in all cases?**

CodyCBakerPhD:
Code is all mostly ready to go, but waiting on next release to do a version bump for altering the tool import structure.

bendichter (approved):

PR #350, opened 2022-01-04
Patch for propagation of gains with neuroscope and stub_test=True

### LINKED ISSUE #330: Gain not propagated to final nwb files as a consequence of SubRecording extractor
## Description
I am running a conversion pipeline from the Buzsaki lab and I am finding that the gain is not propagated to either the LFP nor the raw data. For this I am using the `Neuroscope` data interfaces and I think the problem consists of two parts.

1) First, when the Neuroscope is instantiated the _init_function calls `subset_shank_channels` whose role is to remove non meaningful channels based on the XML file:
https://github.com/catalystneuro/nwb-conversion-tools/blob/0809bb4f845d2fdc7d8f85b8b953d9b4fafd7e32/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py#L107-L111
`subset_shank_channels` in turn calls `se.SubRecordingExtractor` but this **does not inherit all the properties from the base recorder**:
https://github.com/catalystneuro/nwb-conversion-tools/blob/0809bb4f845d2fdc7d8f85b8b953d9b4fafd7e32/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py#L23-L29
This means, that the new instance of the recording extractor ends without the appropriate gains even if those are correclly passed as an argument.

This issue can be solved quickly by adding a gain property to the auxiliar function already in place that adds the properties to the new `SubRecordingExtractor` after it is instantiated:
https://github.com/catalystneuro/nwb-conversion-tools/blob/0809bb4f845d2fdc7d8f85b8b953d9b4fafd7e32/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py#L207-L210

I already tried a a patch in a local branch and that does propagates the gains to where they should be.
However,

2)  The second issue manifest itself when running stubs. The problem is that when running the conversion pipeline with `stub_test=True` the writer class re-instantiates the recording extractor again. The base recording extractor instantiates the writer class in the spot in the following place:
https://github.com/catalystneuro/nwb-conversion-tools/blob/0809bb4f845d2fdc7d8f85b8b953d9b4fafd7e32/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py#L88-L99

 Which in turn ends up calling again `se.SubRecordingExtractor`. In my case, this is done through the `SI013NwbEphysWriter`:
https://github.com/catalystneuro/nwb-conversion-tools/blob/0809bb4f845d2fdc7d8f85b8b953d9b4fafd7e32/nwb_conversion_tools/utils/si013nwbephyswriter.py#L29-L39

There the `self._make_recording_stub()` is defined in the following way as you can see ends up calling `se.SubRecordingExtractor` :
https://github.com/catalystneuro/nwb-conversion-tools/blob/0809bb4f845d2fdc7d8f85b8b953d9b4fafd7e32/nwb_conversion_tools/utils/si013nwbephyswriter.py#L41-L43  

Which ends up with the gains being overwritten again. For this second issue I have not patched locally. I guess the solution can be like one, to add the extractor properties again after calling `se.SubRecordingExtractor`.

However, it seems that we would kill two birds with the same shot if somehow we manage to get `se.SubRecordingExtractor` to inherit all the properties of the parent extractor. Has this been try before and there are dragons lurking there that I don't know about? What do you think?
 
## Steps to Reproduce
TBD

## Environment
Runing the master branch from `nwb_conversion_tools` 

## Checklist

- [] Have you included a minimal code snippet that reproduces the issue you are encountering?

h-mayorquin:
## Motivation
Fixes #330

As mentioned in the issue, so far the neuroscope interfaces do not propagate the gains when stub_test=True. The problems and reasoning behind this are more clearly described in the linked issue. 

This is PR was requested by @CodyCBakerPhD  to discuss the temporary patch.

## Checklist

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
@@ -93,6 +93,10 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
             kwargs.update(channel_ids=self.subset_channels)
 
         recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+
+        # Here, the gains need to be added again.
+        # But shis is specific, the code coulc check for neuroscope, and then call the function to add gains again.
+
         return recording_extractor
 
     def run_conversion(
--- a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
@@ -93,10 +93,6 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
             kwargs.update(channel_ids=self.subset_channels)
 
         recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
-
-        # Here, the gains need to be added again.
-        # But shis is specific, the code coulc check for neuroscope, and then call the function to add gains again.
-
         return recording_extractor
 
     def run_conversion(
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -607,8 +607,15 @@ def add_electrical_series(
     # channels gains - for RecordingExtractor, these are values to cast traces to uV.
     # For nwb, the conversions (gains) cast the data to Volts.
     # To get traces in Volts we take data*channel_conversion*conversion.
-    channel_conversion = recording.get_channel_gains()
-    channel_offset = recording.get_channel_offsets()
+    # If the recording is a sub-recording we recover the gains from the parent recording
+
+    if isinstance(recording, se.SubRecordingExtractor):
+        channel_conversion = recording._parent_recording.get_channel_gains()
+        channel_offset = recording._parent_recording.get_channel_offsets()
+    else:
+        channel_conversion = recording.get_channel_gains()
+        channel_offset = recording.get_channel_offsets()
+
     unsigned_coercion = channel_offset / channel_conversion
     if not np.all([x.is_integer() for x in unsigned_coercion]):
         raise NotImplementedError(
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -9,22 +9,15 @@ from typing import Optional, List
 from warnings import warn
 from collections import defaultdict
 
-import spikeextractors as se
 import pynwb
+from spikeextractors import RecordingExtractor, SortingExtractor
 from numbers import Real
 from hdmf.data_utils import DataChunkIterator
 from hdmf.backends.hdf5.h5_utils import H5DataIO
 from .json_schema import dict_deep_update, OptionalFilePathType, ArrayType
 
 from .recordingextractordatachunkiterator import RecordingExtractorDataChunkIterator
-
-
-def list_get(li: list, idx: int, default):
-    """Safe index retrieval from list."""
-    try:
-        return li[idx]
-    except IndexError:
-        return default
+from ..utils.conversion_tools import get_module
 
 
 def set_dynamic_table_property(
@@ -66,33 +59,13 @@ def set_dynamic_table_property(
             dynamic_table.add_column(name=property_name, description=description, data=values, index=index, table=table)
 
 
-def check_module(nwbfile, name: str, description: str = None):
-    """
-    Check if processing module exists. If not, create it. Then return module.
-    Parameters
-    ----------
-    nwbfile: pynwb.NWBFile
-    name: str
-    description: str | None (optional)
-    Returns
-    -------
-    pynwb.module
-    """
-    assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
-    if name in nwbfile.modules:
-        return nwbfile.modules[name]
-    else:
-        if description is None:
-            description = name
-        return nwbfile.create_processing_module(name, description)
-
-
-def get_nwb_metadata(recording: se.RecordingExtractor, metadata: dict = None):
+def get_nwb_metadata(recording: RecordingExtractor, metadata: dict = None):
     """
     Return default metadata for all recording fields.
+
     Parameters
     ----------
-    recording: RecordingExtractor
+    recording: SpikeInterfaceRecording
     metadata: dict
         metadata info for constructing the nwb file (optional).
     """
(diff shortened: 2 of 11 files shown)


nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -93,6 +93,10 @@ def subset_recording(self, stub_test: bool = False):
             kwargs.update(channel_ids=self.subset_channels)
 
         recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+
+        # Here, the gains need to be added again.
+        # But shis is specific, the code coulc check for neuroscope, and then call the function to add gains again.
+
```
h-mayorquin, line 99: The problem is here. Here, some machinery to recover the gains that are eliminated by "subrecording" should be instantiated. Some ideas are:
1) call the methods in self .recording_extractor before to extract the gains (if available) and then re-assigned them at the end [if available]
2) call the method used in the `neuroscopedatainterface` to solve the problem. We can check if self.recording_extractor is the right one (the recorders have a name attribute) and then use the method.

Both seem kind *very* hack-y and I am not sure I feel comfortable with them. Moreover, will not generate the double scaling problem that @alejoe91 was mentioning in today (2022-01-04) meeting.

h-mayorquin (commented):
The patch to solve the problem at the neuroscope level is straighforward. However, the second issue described in #330 is more tricky because it is in the very general class of `baserecordingextractor`, see the details in the code review.

alejoe91:
@CodyCBakerPhD @h-mayorquin after today's discussion, should we wait for the usage of the new SI objects? The problem is inherently fixed there

CodyCBakerPhD:
Yeah, this is trickier than expected. If we just add the gains back onto the `data_interface.recording_extractor` after each time it is sub-setted without clearing them from the parent object then we end up with that situation Alessio describe where the calls to `get_traces` during the conversion process would apply the gain multiple times.

I guess the most short-term solution would actually be to adjust `add_electrical_series` to retrieve gains from top-level parent objects if the passed `recording` is a `SubRecordingExtractor`.

The plan to update to `BaseRecording` objects with the proper subsetting behavior is definitely the better long-term solution, it just comes down to how long it happens to take those changes to go through. We'll probably know more after the meetings tomorrow.

h-mayorquin:
@CodyCBakerPhD 
> Yeah, this is trickier than expected. If we just add the gains back onto the `data_interface.recording_extractor` after each time it is sub-setted without clearing them from the parent object then we end up with that situation Alessio describe where the calls to `get_traces` during the conversion process would apply the gain multiple times.
> 
> I guess the most short-term solution would actually be to adjust `add_electrical_series` to retrieve gains from top-level parent objects if the passed `recording` is a `SubRecordingExtractor`.
> 
> The plan to update to `BaseRecording` objects with the proper subsetting behavior is definitely the better long-term solution, it just comes down to how long it happens to take those changes to go through. We'll probably know more after the meetings tomorrow.

All right, I did added some extra functionality to `add_electrical_series` as indicated. I tried the conversion locally and it is in fact propagated to the final nwb file.

As I don't understand fully the problem of double scaling I am not totally confident about where the extraction of gains from parent recording should take place (see the review). Also, it might be relevant to note that the be noted that `BinDatRecordingExtractor` which is the parent class of `NeuroscopeRecordingExtractor` inherits does not use the `write_scaled` argument in the `get_traces` function:

https://github.com/catalystneuro/spikeextractors/blob/e7fa83719deaa2cebbb1a741d124e877375aaeea/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py#L119-L134 

This makes me confused about how the double scaling will happen. I guess we can discuss this in today meeting.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -642,6 +649,8 @@ def add_electrical_series(
     else:
         raise NotImplementedError(f"iterator_type ({iterator_type}) should be either 'v1' or 'v2' (recommended)!")
 
+    # Not sure if recovering the gains from the parent recording should be done here AFTER the data is extracted.
+
```
h-mayorquin: As discussed, maybe the extraction of gains should be here. Maybe it is working in our case because `BinDatRecorder` does not really use `return_scaled`.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -607,8 +607,15 @@ def add_electrical_series(
     # channels gains - for RecordingExtractor, these are values to cast traces to uV.
     # For nwb, the conversions (gains) cast the data to Volts.
     # To get traces in Volts we take data*channel_conversion*conversion.
-    channel_conversion = recording.get_channel_gains()
-    channel_offset = recording.get_channel_offsets()
+    # If the recording is a sub-recording we recover the gains from the parent recording
+
+    if isinstance(recording, se.SubRecordingExtractor):
+        channel_conversion = recording._parent_recording.get_channel_gains()
```
CodyCBakerPhD, line 613:
```suggestion
        temp_recording = recording
        while isinstance(temp_recording , se.SubRecordingExtractor):
            temp_parent = temp_recording ._parent_recording
            channel_conversion = temp_parent .get_channel_gains()
            temp_recording = temp_parent 
```

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -607,8 +607,15 @@ def add_electrical_series(
...
+        channel_offset = recording._parent_recording.get_channel_offsets()
```
CodyCBakerPhD, line 614:
```suggestion
        temp_recording = recording
        temp_parent = temp_recording._parent_recording
        channel_conversion = temp_parent.get_channel_gains()
        channel_offset = temp_parent.get_channel_offsets()
        while isinstance(temp_recording , se.SubRecordingExtractor) and channel_conversion != np.ones(len(channel_conversion)):
            temp_parent = temp_recording._parent_recording
            channel_conversion = temp_parent.get_channel_gains()
            channel_offset = recording._parent_recording.get_channel_offsets()
            temp_recording = temp_parent 
```

h-mayorquin:
All the tests work on my personal environment. I am not sure why I am getting the errors in CI test. They seem to be a failure to import `se`.

CodyCBakerPhD:
> All the tests work on my personal environment. I am not sure why I am getting the errors in CI test. They seem to be a failure to import `se`.

This is because the code there does relative imports (`from spikeextractors import RecordingExtractor`, etc.). Went ahead and fixed that for you.

I'll work on getting the CI debugged; some kind of loose dependency issue on the minimal requirements.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin LGTM - great job on the tests!

PR #348, opened 2022-01-04
Version 0.10.3 changelog - Not for direct merge

CodyCBakerPhD:
## Motivation

This PR is the set of changes introduces from the now un-merged `SpikeInterface` PR. The changelog has been properly updated to reflect current version on `main` (> `v0.9.4`,  < `v0.10.*`). This PR is not intended to be directly merged altogether, but just a reference point for things to do in separate, smaller scale future PRs.

We do believe that much of this code can be useful going forward, but we are going to ask that each sub-feature be introduces in the smallest possible PR (less than +/- 300 changes or so) each with very thorough tests and documentation included to minimize the chance that something breaks in the process.

## Checklist

- [X] Have you checked our [Contributing](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/docs/contribute.rst) document?
- [X] Have you ensured the PR description clearly describes the problem and solutions?
- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/nwb-conversion-tools/pulls) for the same change?

PR #351, opened 2022-01-04
Update ecephys write for BaseRecording

CodyCBakerPhD:
## Motivation

This is a simple PR to update the ecephys write functions to work on `BaseRecording`'s, including `RecordingExtractors` wrapped by the API map from SpikeInterface.

This may also be the better solution to automatically resolving @h-mayorquin gain issue with interface that have multiple `SubRecording` nestings.

## How to test the behavior?
Tests just need to be debugged, maybe add some new ones to verify new functionality.

## Checklist


CodyCBakerPhD:
@alejoe91 Hopefully the CI finishes soon; I see errors on my end related to the non-existence of a `get_channel_property_names` analog from `spikeextractors`. What is the new function for doing this sort of thing?

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -607,36 +551,37 @@ def add_electrical_series(
...
+            raise NotImplementedError(
+                "Unable to coerce underlying unsigned data type to signed type, which is currently required for NWB "
+                "Schema v2.2.5! Please specify 'write_scaled=True'."
+            )
+        elif np.any(unsigned_coercion != 0):
+            warnings.warn(
+                "NWB Schema v2.2.5 does not officially support channel offsets. The data will be converted to a signed "
+                "type that does not use offsets."
+            )
+            unsigned_coercion = unsigned_coercion.astype(int)
```
alejoe91, line 565: @CodyCBakerPhD ATM `unsigned_coercion` is computed but not used...must be propagated to the DataChunkIterator as we did in the dev branch (with the dtype correction)

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1013,177 +944,86 @@ def write_units(
...
+
+    # 2. fill with provided custom descriptions
+    for x in metadata["units"]:
+        if x["name"] not in list(unit_columns):
+            raise ValueError(
+                f'"{x["name"]}" not a property of sorting object, set it first and rerun')
+        unit_columns[x["name"]]["description"] = x["description"]
+
+    # 3. For existing electrodes table, add the additional columns and fill with default data:
+    add_properties_to_dynamictable(
```
alejoe91: We need to change this function to allow to add property to a dynamic table created outside of an NWBFile

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -607,36 +548,37 @@ def add_electrical_series(
...
-        )
-        unsigned_coercion = unsigned_coercion.astype(int)
-    if write_scaled:
-        eseries_kwargs.update(conversion=1e-6)
-    else:
-        if len(np.unique(channel_conversion)) == 1:  # if all gains are equal
-            eseries_kwargs.update(conversion=channel_conversion[0] * 1e-6)
-        else:
+    if recording.has_scaled_traces():
+        channel_conversion = recording.get_channel_gains()
```
h-mayorquin, line 552: I have a question.
In my case, the problem is that making a subrecording in `baserecordingextractor` eliminates the gains from the recording object (before passing it to the `write_recording` function). 

Is the goal here to modify the code in a way that we can use `se.SubRecordingExtractor` without having to remove the gains. 

I am failing how this is going to solve the problem described in #330. Or maybe this is somewhere else in the code?

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: `v0.9.*` returns us to the practice of include offsets as a property of the electrode columns: https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L247 thus tests for intan pass without needing the older `unsigned_coercion` part, which means we could remove it. But we're also getting movement on adding offsets to NWB so maybe we should just wait for that most proper of solutions for now

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: As Alessio mentioned, gain propagation through `spikeinterface/spikeinterface` method (`FrameSliceRecording`) seen here: https://github.com/catalystneuro/nwb-conversion-tools/pull/348/files#diff-e1af4b855c44fd131e9ea6bc47af5c62c403c3a044b260362b692ae41dc97a20R51-R53 should in principle resolve the issue naturally without using `SubRecordingExtractors`. So we would also have to replace those subset calls in this PR

CodyCBakerPhD:
Has now been fully absorbed across other PRs

PR #355, opened 2022-01-05
change example prop to float

bendichter:
attempt to fix pynwb 2.0.0 tests

## Motivation

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,4 +1,4 @@
-pynwb==1.5.1
+pynwb==2.0.0
 tqdm==4.60.0
 natsort==7.1.1
 numpy==1.19.3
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD (commented):
@bendichter Can you bump the `requirements-full.txt` `pynwb` version to make sure this passes with `2.0.0`?

bendichter:
see #356 for the root problem, which I think should really be addressed in a better way than this PR

CodyCBakerPhD:
#356 should be directly addressed as a part of the steady incorporation of the refactored ephys writer, hopefully won't take long

PR #358, opened 2022-01-05
Update recordings only

CodyCBakerPhD:
## Motivation

Splinter of #351 that provides new SpikeInterface API changes for a couple of the helper functions related to `write_recording`.

## How to test the behavior?

Will review if more tests are needed.

## Checklist


CodyCBakerPhD:
@alejoe91 Sorry I wasn't totally clear on the intention of #351; the goal with these initial PRs is to make the smallest number of changes possible to update these API, starting with Recording objects only. The goal is to keep both the changelogs and impacts to code behavior minimal.

The goal of this PR is just to take the `add_*` helper functions to `write_recording` work on `BaseRecording` objects, coercing `RecordingExtractor` objects to them in the process, while keeping original write behaviors as close as possible to what they were before so that we can more carefully review the eventual refactors you included in #351 individually piece by piece.

To achieve this here, it's mostly going in and changing API calls such as the `property_names = recording.get_property_keys()`, `.frame_to_time()` -> `.get_times()`, including `segment_index` in the iterator and `add_electrical_series` and stuff of that nature.

Currently, the tests seem to be failing for similar reasons to #351, but they seem specific to either how the test objects are created or how the `OldToNewAPI` is mapping things - any idea how to fix this?

CodyCBakerPhD:
@alejoe91 Also, when I locally try debugging specifically using `test_gin.py` (the CI will not reach that point until the `test_si.py` passes), I run into this first error with `SpikeGLX`, indicating the property maybe isn't getting copied over?

image

I'm using https://github.com/SpikeInterface/spikeinterface/pull/428, too.

alejoe91:
@CodyCBakerPhD what's the status?

CodyCBakerPhD:
@alejoe91 Sorry got distracted by some other things. Double checking now but I think I fixed everything

CodyCBakerPhD:
@alejoe91 Ready for your final review; the only bit of magic going on is the gain inference in `SubRecordingExtractors` at the start of `add_electrical_series`. Might be nice to keep that functionality somehow going forward, would be fine keeping the code here or if it's something you think the `old_api_utils` deserve, that's cool too.

CodyCBakerPhD:
Oh, and as far as compatibility goes for the tests and `spikeinterface/spikeextractors/NwbRecordingExtractor` reading of the `offset_to_uV`, I think it's OK to keep the tests as is just for the moment and do a roundtrip update for that in the coming weeks.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -500,6 +523,30 @@ def add_electrical_series(
...
+            default_channel_conversion = np.ones_like(channel_conversion)
+            if np.any(channel_conversion != default_channel_conversion):
+                break
+            else:
+                temp_recording = parent_recording
+        checked_recording = OldToNewRecording(oldapi_recording_extractor=recording)
+        if np.any(channel_conversion != default_channel_conversion):
+            checked_recording.set_channel_gains(gains=channel_conversion)
+        if np.any(channel_offset != np.zeros_like(channel_offset)):
+            checked_recording.set_channel_offsets(offsets=channel_offset)
```
alejoe91, line 547: @CodyCBakerPhD this could also happen when the recording is a `BasePreprocessorRecordingExtractor`. I woudl suggest to merge as is and fix it later. Do you agree?

alejoe91 (approved):

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: Very interesting - good to know! Fine to delay that since we've never made use of that specific type.

CodyCBakerPhD: @alejoe91 If you think the logic applies generally enough, sure! I know it could be interpreted as a problem on other non-NWB RecordingExtractor contexts of translating an old to a new.

PR #359, opened 2022-01-05
handle missing prop data

### LINKED ISSUE #356: filling property data only works for float
https://github.com/catalystneuro/nwb-conversion-tools/blob/14fa9bf99f93f627c56b98a089791fb29d8775dc/nwb_conversion_tools/utils/spike_interface.py#L1097

This line will only work if the original data is float. It will not work if the original data is int (in which case the data needs to be promoted to float) or string (in which case we should fill with '')

bendichter:
## Motivation

fix #356 

Fills missing property data based on data type:

if float, missing data is NaN
if str, missing data is ''
if int and contains some missing data, promote to float and missing data is NaN

### DIFF
--- a/.github/workflows/testing-linux.yml
+++ b/.github/workflows/testing-linux.yml
@@ -62,7 +62,6 @@ jobs:
           datalad get -r ./spikegadgets/
           datalad get -r ./spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/
           datalad get -r ./phy/phy_example_0/
-          datalad get -r ./blackrock/
           datalad get -r ./axona/
       - name: Run full pytest with coverage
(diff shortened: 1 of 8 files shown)


CodyCBakerPhD:
@bendichter Would it be possible to add specific tests for this behavior?

CodyCBakerPhD:
@bendichter Thanks, just waiting for tests to finish.

I know this kind of thing was resolved somewhere in the #348 changes, but pinning it down is tricky. I'm sure that'll come up again in the piecewise refactors but this is good as a quick standalone fix for our current purposes.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@bendichter Oh my, tests are actually failing now: https://github.com/catalystneuro/nwb-conversion-tools/runs/4719902648?check_suite_focus=true

Looks like a simple indexing issue though, should be easy to fix.

CodyCBakerPhD (approved):

PR #368, opened 2022-01-13
Add write_electrical_series option to write_recording

alejoe91:
Fix https://github.com/catalystneuro/nwb-conversion-tools/pull/368

## Motivation

Add `write_electrical_series` option to `write_recording` to optionally avoid writing time series .

## How to test the behavior?
```
write_recording(..., write_electrical_series=False)
```

## Checklist

### DIFF
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -750,7 +750,7 @@ def add_epochs(recording: RecordingExtractor, nwbfile=None, metadata: dict = Non
                 )
 
 
-def add_device_electrode_info(recording: RecordingExtractor, nwbfile=None, metadata: dict = None):
+def add_electrodes_info(recording: RecordingExtractor, nwbfile=None, metadata: dict = None):
     """
     Adds device, electrode_groups, and electrodes info to the nwbfile
 
@@ -859,7 +859,7 @@ def add_all_to_nwbfile(
     if nwbfile is not None:
         assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
 
-    add_device_electrode_info(recording=recording, nwbfile=nwbfile, metadata=metadata)
+    add_electrodes_info(recording=recording, nwbfile=nwbfile, metadata=metadata)
 
     if write_electrical_series:
         add_electrical_series(
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -748,8 +748,8 @@ def add_epochs(recording: RecordingExtractor, nwbfile=None, metadata: dict = Non
                     stop_time=recording.frame_to_time(epoch["end_frame"]),
                     tags=epoch_name,
                 )
-                
-                
+
+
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
@alejoe91 This being an alternative to directly calling

```
add_device(recording=recording, nwbfile=nwbfile, metadata=metadata)
add_electrode_groups(...)
add_electrodes(...)
```

which itself could be wrapped into a function (`add_electrode_info` or something like that) to reduce lines of code, with the main difference being that for your changes, the `nwbfile` object is also created or opened at the same time?

I assume this is building towards a way of specifying electrode information for SortingInterfaces in advance of the WaveformExtractor incorporation?

CodyCBakerPhD:
@alejoe91 A couple other things

- [ ] Can you add a test for this new option?
- [ ] Whenever we add these kind of arguments to `write_recording` we usually propagate them up to the `DataInterface` levels. Can you do that as well? (should only require adding it to `BaseRecordingDataInterface`)

alejoe91:
> @alejoe91 This being an alternative to directly calling
> 
> ```
> add_device(recording=recording, nwbfile=nwbfile, metadata=metadata)
> add_electrode_groups(...)
> add_electrodes(...)
> ```
> 
> which itself could be wrapped into a function (`add_electrode_info` or something like that) to reduce lines of code, with the main difference being that for your changes, the `nwbfile` object is also created or opened at the same time?
> 
> I assume this is building towards a way of specifying electrode information for SortingInterfaces in advance of the WaveformExtractor incorporation?

@CodyCBakerPhD I made an extra function (`add_device_electrode_info`). However I think that it's more user friendly to keep the `write_electrical_series` option in the `write_recording` instead of creatung a new `write_` function.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -750,6 +750,46 @@ def add_epochs(recording: RecordingExtractor, nwbfile=None, metadata: dict = Non
                 )
 
 
+def add_device_electrode_info(recording: RecordingExtractor, nwbfile=None, metadata: dict = None):
```
CodyCBakerPhD, line 753:
```suggestion
def add_electrodes_info(recording: RecordingExtractor, nwbfile=None, metadata: dict = None):
```
Minor naming change; the device and electrode groups are both information fields of the electrodes, so I'd say the purpose of this function is just to add the ElectrodeTable and any of its dependencies.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -815,27 +859,23 @@ def add_all_to_nwbfile(
...
-        metadata=metadata,
-        write_as=write_as,
-        es_key=es_key,
-        write_scaled=write_scaled,
-        compression=compression,
-        compression_opts=compression_opts,
-        iterator_type=iterator_type,
-        iterator_opts=iterator_opts,
-    )
+    add_device_electrode_info(recording=recording, nwbfile=nwbfile, metadata=metadata)
```
CodyCBakerPhD, line 862:
```suggestion
    add_electrodes_info(recording=recording, nwbfile=nwbfile, metadata=metadata)
```

alejoe91:
@CodyCBakerPhD applied suggested changes. Can you accept and merge?

CodyCBakerPhD (approved):

Issue #163, opened 2021-04-07
Dynamic NWBConverter class generation and/or add_interface


CodyCBakerPhD:
sneakers-the-rat's [fork of the repo](https://github.com/sneakers-the-rat/nwb-conversion-tools/blob/add_interface/nwb_conversion_tools/nwbconverter.py#L74-L140) diverges in a number of ways. One of the sub goals of that project involves the automated generation of NWBConverter classes that may contain different interfaces.

**Is your feature request related to a problem? Please describe.**
We normally specify a subclass of a NWBConverter as

```
    class MyLabNWBConverter(NWBConverter):
        data_interface_classes = dict(InterfaceName=DataInterface1, ...)
```

where the user manually specifies the interface classes at the level or moment of class definition. Not really an issue, but if the user isn't familiar with classes or programming in general it might be a bit scary or confusing upon first glance.

Even beyond that, looking ahead to a converter class structure or something similar that can operate on folders, a more dynamic way of specifying the interface classes might be desired.

**Describe the solution you'd like**
sneaker's suggests a function called `add_interface(...)` which essentially adds a desired class to an instantiated converter object. This is, right away, a problematic break of our API which has classes defined before instantiation of the converter object, which necessarily leads to the instantiation of the data interface objects themselves.

I'll adjust that proposition instead to be a class method called prior to instantiating the converter object on a converter subclass which itself _contains an empty set of_ `data_interface_classes`, and instead has its classes added via the `MyLabNWBConverter.add_interface(...)` method used before creating `my_nwb_converter = MyLabNWBConverter(source_data)`.

**Describe alternatives you've considered**
Another way to perform this exact functionality would be to override the `type` functionality acting on an NWBConverter where we pass the class list in as an argument. One could construct that argument in the same iterative way of `add_interface`, but instead of it modifying some existing class structure, it returns an entirely new one.

This approach, while not something we would want to ask new users (or possibly any user) to actually do in practice, it is what could work under the hood of a function that wraps it in a nicer looking way.

**Additional context**
In general, an `add_interface` style of function implies an iterative approach to assembling a converter class, rather than a one-shot approach we normally do. I'd just point out that from a high level, we can also just assemble the list of interface classes iteratively ahead of time and then still build a class structure from the existing one-shot method without needing any work beyond that, and especially to the core of what has already been made and tested.

[...]

PR #371, opened 2022-01-16
Update sorting only

alejoe91:
## Motivation

Splinter of #351 that provides new SpikeInterface API changes for add_units.

## How to test the behavior?

Tests are passing with new API changes

## Checklist


nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1148,160 +1155,37 @@ def write_units(
...
-            aggregated_unit_properties[key] = [np.nan if x is None else x for x in val]
+        aggregated_unit_properties[pr] = checked_sorting.get_property(pr)
 
     for i, unit_id in enumerate(unit_ids):
         if use_times:
-            spkt = sorting.frame_to_time(sorting.get_unit_spike_train(unit_id=unit_id))
-        else:
-            spkt = sorting.get_unit_spike_train(unit_id=unit_id) / sorting.get_sampling_frequency()
+            warnings.warn(
+                "'use_times' is not supported yet by the new SI API. " "Computing spike times using sampling frequency"
```
CodyCBakerPhD, line 1181: @alejoe91 Would it make sense to delay this refactor until the new SI API supports everything it used to? What new features or improvements do the new Sorting objects bring?

alejoe91: I believe only times are missing for the sorting object. Will work on it next week and we can wait until then 🙂

alejoe91: @CodyCBakerPhD this implements the time mechanism: https://github.com/SpikeInterface/spikeinterface/pull/464

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1065,7 +1067,7 @@ def get_nspikes(units_table: pynwb.misc.Units, unit_id: int):
         return units_table["spike_times_index"].data[index] - units_table["spike_times_index"].data[index - 1]
 
 
-def write_units(
+def add_units(
     sorting: SortingExtractor,
```
CodyCBakerPhD, line 1071:
```suggestion
    sorting: SpikeInterfaceSorting,
```
Suggest defining `SpikeInterfaceSorting = Union[SortingExtractor, BaseSorting]` above

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1113,8 +1115,15 @@ def write_units(
     units_description : str (optional)
         Text description of the sorting table; recommended to included parameters of sorting method, curation, etc.
     """
-    unit_ids = sorting.get_unit_ids()
-    fs = sorting.get_sampling_frequency()
+    if isinstance(sorting, SortingExtractor):
+        fs = sorting.get_sampling_frequency()
```
CodyCBakerPhD, line 1119:
```suggestion
        sampling_frequency = sorting.get_sampling_frequency()
```
We've been getting in the habit lately of using fuller variable names to make the code easier to read for people unfamiliar with the code.

CodyCBakerPhD:
@alejoe91 So, to confirm, the new `spikeinterface/spikeinterface` `BaseSorting` objects completely absorb or map all previous `spikeinterface/spikeextractor` `SortingExtractor` `unit_spike_features`?

What types of `unit_spike_features` used to be present (either possible, or commonly found) and what is the new method of setting/retrieving them? Are they types of properties now?

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -1148,160 +1151,36 @@ def write_units(
...
-
-    # handle missing data differently depending on type of data
-    for key, val in aggregated_unit_properties.items():
-        if all(isinstance(x, int) or x is None for x in val) and any(x is None for x in val):
-            aggregated_unit_properties[key] = [np.nan if x is None else float(x) for x in val]
-        if all(isinstance(x, str) or x is None for x in val):
-            aggregated_unit_properties[key] = [x or "" for x in val]
-        if all(isinstance(x, float) or x is None for x in val):
-            aggregated_unit_properties[key] = [np.nan if x is None else x for x in val]
+        aggregated_unit_properties[pr] = checked_sorting.get_property(pr)
```
CodyCBakerPhD, line 1172: I assume that `spikeinterface/spikeinterface` `BaseSorting` objects handle missing values based on their data type in this same manner?

alejoe91:
> @alejoe91 So, to confirm, the new `spikeinterface/spikeinterface` `BaseSorting` objects completely absorb or map all previous `spikeinterface/spikeextractor` `SortingExtractor` `unit_spike_features`?
> 
> What types of `unit_spike_features` used to be present (either possible, or commonly found) and what is the new method of setting/retrieving them? Are they types of properties now?

The new API relies on the `WaveformExtractor` to deal with any spike-specific info. So spike features don't exist anymore in the Sorting object.

CodyCBakerPhD:
> > @alejoe91 So, to confirm, the new `spikeinterface/spikeinterface` `BaseSorting` objects completely absorb or map all previous `spikeinterface/spikeextractor` `SortingExtractor` `unit_spike_features`?
> > What types of `unit_spike_features` used to be present (either possible, or commonly found) and what is the new method of setting/retrieving them? Are they types of properties now?
> 
> The new API relies on the `WaveformExtractor` to deal with any spike-specific info. So spike features don't exist anymore in the Sorting object.

So from a design point, refactoring the `write_sorting` for minimal loss of back-compatibility should also include (a) instantiating a `WaveformExtractor` that has the features from a passed `SortingExtractor`, and (b) allowing an optional `WaveformExtractor` to be passed in if a `BaseSorting` is the main object; then instead of outright removal of the spike features here, refactor to use the `WaveformExtractor` (however it reached that step) to perform those operations?

alejoe91:
> > > @alejoe91 So, to confirm, the new `spikeinterface/spikeinterface` `BaseSorting` objects completely absorb or map all previous `spikeinterface/spikeextractor` `SortingExtractor` `unit_spike_features`?
> > > What types of `unit_spike_features` used to be present (either possible, or commonly found) and what is the new method of setting/retrieving them? Are they types of properties now?
> > 
> > 
> > The new API relies on the `WaveformExtractor` to deal with any spike-specific info. So spike features don't exist anymore in the Sorting object.
> 
> So from a design point, refactoring the `write_sorting` for minimal loss of back-compatibility should also include (a) instantiating a `WaveformExtractor` that has the features from a passed `SortingExtractor`, and (b) allowing an optional `WaveformExtractor` to be passed in if a `BaseSorting` is the main object; then instead of outright removal of the spike features here, refactor to use the `WaveformExtractor` (however it reached that step) to perform those operations?

Option (a) would be impossible because the `WaveformExtactor` needs a recording too. Option (b) is ideally possible, but I think it would be better to just use `write_waveforms` (also in general I don't think we recommended to write features often)

Is there some sort of caching of the environment in the tests? The requirements point to spikeinterface@master, but apparently tests fail to resolve `get_unit_spike_train(..., return_times=True)`, which is already merged in [SI master](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/basesorting.py#L56)

CodyCBakerPhD:
> Option (a) would be impossible because the WaveformExtactor needs a recording too. Option (b) is ideally possible, but I think it would be better to just use write_waveforms (also in general I don't think we recommended to write features often)

I mean, the information about the features would be somehow contained within the `SortingExtractor`; surely then you could initialize a `RecordingExtractor` + `WaveformExtractor` both from the information in the old object?

> Is there some sort of caching of the environment in the tests? The requirements point to spikeinterface@master, but apparently tests fail to resolve get_unit_spike_train(..., return_times=True), which is already merged in SI master

Nope; fresh install every time. `requirements-minimal.txt` points to current master of SI, but `requirements-full.txt` points to specific commit hashes in place of official released versions (which would be preferable, but not possible here). If you believe the commit hash on that is out of date feel free to bump it on any given PR that needs it.

alejoe91:
> > Option (a) would be impossible because the WaveformExtactor needs a recording too. Option (b) is ideally possible, but I think it would be better to just use write_waveforms (also in general I don't think we recommended to write features often)
> 
> I mean, the information about the features would be somehow contained within the `SortingExtractor`; surely then you could initialize a `RecordingExtractor` + `WaveformExtractor` both from the information in the old object?
> 

That would require to re-save the waveforms and indices into new files and I don't think that we could initialize a recording from the sorting object only, as we should have no info about channels, num frames, etc.. I think it's fine to leave features behind in my opinion with this refactoring, but we can ask what others think

CodyCBakerPhD:
S> The new API relies on the WaveformExtractor to deal with any spike-specific info. So spike features don't exist anymore in the Sorting object.

So let's say in the old API, I'd want to add a non-waveform unit spike feature with the goal of classifying certain spikes across all units in a certain manner.

For example, for `spike_1` of `unit_2`, I want to set or associated a string value equal to "evoked_response". To `spike_2` of `unit_3` I likewise assign the label "spontaneous". For the sake of simplicity, never mind the missing data for NWB write of `spike_1` of `unit_3` and all spikes of `unit_1` (just empty strings I suppose).

I would have done this previously via something like

```python
my_sorting.set_unit_spike_features(
    unit_ids=[2], feature_name="classification", value="evoked_response", indexes=[1]
)
my_sorting.set_unit_spike_features(
    unit_ids=[3], feature_name="classification", value="spontaneous", indexes=[2]
)
```

My question is with the new `WaveformExtractor`, what is the new analogous API call?

alejoe91:
@CodyCBakerPhD that is not possible in the new API. In general, one could add array properties and respective indices in the property tables. So bottom line is that some back-compatibility is broken with the new API

alejoe91:
@CodyCBakerPhD @bendichter IMO this can be merged

alejoe91:
> I'd rather have some logic for writing non-waveform features from old `SortingExtractor` objects in the event that is the type passed; similar to how we handle epochs from old `RecordingExtractor` now

Ok we can make a separate function that is only called if the object is an old extractor

tests/test_internals/test_si.py
```diff
@@ -293,7 +293,8 @@ def test_write_sorting(self):
 
             # check that missing props are filled correctly
             np.testing.assert_array_equal(nwbfile.units["float_prop"][:], [80.0, np.nan, np.nan])
-            np.testing.assert_array_equal(nwbfile.units["int_prop"][:], [80.0, np.nan, np.nan])
+            # int is not filled with nan in SI
+            # np.testing.assert_array_equal(nwbfile.units["int_prop"][:], [80.0, np.nan, np.nan])
```
CodyCBakerPhD, line 297:
```suggestion
            # np.testing.assert_array_equal(nwbfile.units["int_prop"][:], [80, 0, 0])
```
Should update test to reflect and communicate new standard default.

@bendichter Note that this means all integer valued properties have a default fill value of `0` now, is this OK with you? If not, an alternative for us in this PR would be to restore a part of the aggregated unit code, and coerce integer valued properties to float in order to designate items truly set as `0` from missing values filled by `np.nan`.

CodyCBakerPhD:
Summary of changes:

i) Input argument now expects either `SortingExtractor` or `BaseSorting`; all non-`unit_spike_feature` related operations are performed on a `BaseSorting` object, either the one passed or one coerced from the `old_to_new_api` mapping.

ii) All property reading from the sorting object is now done with a `BaseSorting`, which naturally does a lot of the 'nice' data retrieval features we had previously hard-coded here. Some minor changes in behavior, in particular the default values of integer valued properties.

iii) writing `unit_spike_features` as additional columns of the table is only supported for `SortingExtractor` objects, no change of behavior is expected here.

tests/test_internals/test_si.py
alejoe91: @CodyCBakerPhD we could also add the option `missing_value` to Spikeinterface when setting property on a subset of channels. In case one uses `np.nan`, then the property becomes float, but this would leave to possibility to set say -1 as default missing option. What do you think?

CodyCBakerPhD: Talked to Ben about it and he supports the `float` casting of `int` types. Increases data sizes a tad, but this isn't that heavy a data type anyway. Really we would want to use something like numpy masks to distinguish missing values in all cases, but NWB doesn't really have a great analog of that, although it is important for us to properly distinguish missing vs. non-missing data values.

@alejoe91 At what level do you think we should resolve that? In this PR, or in SI?

alejoe91: I think in SI, will try to implement the `missing_value` option tomorrow :) Will ping you when I'm done, thanks for the reminder

alejoe91: @CodyCBakerPhD to be tested with this patch: https://github.com/SpikeInterface/spikeinterface/pull/483

Now we can provide the `missing_value` argument when setting property. Note that setting `np.nan` for an int property will result in the property values being handled as `object` array, which gives more flexibility (e.g. you can have a nan value in a string array, etc.)

alejoe91: @CodyCBakerPhD after discussion with Sam, we decided to be more strict about missing values. If you set an int property, the `missing_value` must also be int. In case one has int values and wants to set nans as missing, she should cast the property values outside of the property function. I think that this stricter approach will be less error-prone and "magic" in the long run

bendichter: I disagree. There is no int value that works for missing values.

CodyCBakerPhD: @alejoe91 @samuelgarcia I do agree with Ben as per my justifications above, but I see what you're saying about giving the user control via requiring them to set it as float if they want to do that themselves. The thing is, they have to think about that pretty far in advance, whereas it would be the kind of thing that is easy to slip the mind unless you remember every time you add a property to (a) check if you have missing values and (b) remember that, in SI, the way you specify missing values is to pre-cast as float. I'd recommend just changing line #163 in the BaseExtractor (https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/base.py#L163) to `if values.dtype.kind in ['f', 'i']:` if I were you guys.

If you don't budge on that though, we should still have code on this PR that handles the float upcasting.

samuelgarcia: Hi @CodyCBakerPhD @alejoe91 @bendichter.
I think that in the spikeinterface context we should not change the dtype on the fly and magically depending missing values. When an int is set to property, generally it is on purpose (indexing).
Having missing values, in context of spike sorting is a corner case contary to data management.
For property handling we have a big constrain of dump to dict  for parralel processing and dump to json for provenance. Handling in missing memory is easy, we could use mask array of numpy but the dump to json will be pain.
This is why I suggest that in SI, there is no magic upcasting on the fly. An internal property should keep the same dtype.
Upcasting should be external.
So yes it will force us to have some code here in conversion tools but I think it is a good practice to have clear rules for upcasting here and not in SI.

CodyCBakerPhD: But for now, I'll just get something in this PR to fix the issue temporarily.

CodyCBakerPhD: Furthermore, from lines https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/base.py#L157-L170 I don't actually see a way for us to take a Sorting object this far downstream and handle this issue on our side, since by that time the information of both the `ids` set at function call and original `values` aren't kept anywhere and so the information of missing values is lost.

This is a step back even from the old `SortingExtractors`, where these were dynamic dict-of-dicts that allowed not only allowed users full control over missing values, but also preserved that information internally so it could be used downstream.

@alejoe91 Just so we can get this PR done, what I would recommend in the meantime is for you to restore some of the previous code for inferring missing data for integer values properties in the case where the object passed was a `SortingExtractor`, similar to (and likely containable within) the block of code at the end that handles the `unit_spike_features`; just so we can still maintain support for that via old `SortingExtractor` objects until the new SI makes the masking changes.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@alejoe91 @samuelgarcia Looking good over here, going to merge now. Thanks for working on this!

PR #364, opened 2022-01-10
Restore movie data chunk iterator

Saksham20:
## Motivation

Rebasing #265 

Additionally, the user has the options to specify multiple video files to go in the same ImageSeries container only when the `external_mode=True`. The `external_file` attribute of the `ImageSeries` gets the list of video files. 

## How to test the behavior:

```python
from nwb_conversion_tools import (
    NWBConverter,
    MovieInterface
)
from pynwb import NWBHDF5IO

movie_file_paths = ['file1.avi', 'file2.avi']
nwbfile_path = 'nwbfile.nwb'

class MovieTestNWBConverter(NWBConverter):
    data_interface_classes = dict(Movie=MovieInterface)

source_data = dict(Movie=dict(file_paths= movie_file_paths]))
converter = MovieTestNWBConverter(source_data)
metadata = converter.get_metadata()
metadata.update(
            Behavior=dict(
                Movies=[
                    dict(
                        name="CustomName",
                        description="CustomDescription",
                        unit="CustomUnit",
                        resolution=12.3,
                        comments="CustomComments",
                    ),
                    dict(
                        name="CustomName", # the same name. Thus the imageseries will contain a list of video files under external_file
                        description="CustomDescription",
                        unit="CustomUnit",
                        resolution=12.3,
                        comments="CustomComments",
                    )
                ]
            )
        )

converter.run_conversion(metadata=metadata, nwbfile_path=nwbfile_path, overwrite=True)
with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
    nwbfile = io.read()
    custom_name = metadata["Behavior"]["Movies"][0]["name"]
    external_path_list = nwbfile.acquisition[custom_name].external_path
assert set(external_path_list)==set(movie_file_paths)
```

## Checklist

- [x] Have you checked our [Contributing](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/docs/contribute.rst) document?
- [x] Have you ensured the PR description clearly describes the problem and solutions?
- [x] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/nwb-conversion-tools/pulls) for the same change?
- [x] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?

Issue #287, opened 2021-10-25
Convert many tests to unittest.TestCases


CodyCBakerPhD:
Adding this to the global to-do list, low priority. Would be another good first issue to tackle for someone learning the workflows.

**Is your feature request related to a problem? Please describe.**
No issue with the tests, though could make visual output cleaned similar to how the `test_gin.py` utilizes naming functions.

**Describe the solution you'd like**

Many of the current tests, such as those in the `test_interface.py`, were put together ad hoc to get initial coverage up and running. But these could be refactored for code reduction as there is a lot of duplicity in stuff like making temporary test directories and cleanup after each run. This is the sort of thing that `unittest.TestCase` is really built for.

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/catalystneuro/nwb-conversion-tools/issues)?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [X] Have you checked our [Contributing](https://nwb-conversion-tools.readthedocs.io/en/master/developer_guide.html) document?

CodyCBakerPhD:
Compared to what it used to be they've largely gotten better

PR #372, opened 2022-01-17
Add context class movie

Saksham20:
## Motivation

Adding a video capture context manager with various properties. Based on dicussion [here](https://github.com/catalystneuro/nwb-conversion-tools/pull/265#issuecomment-938687660)

This PR breaks #364 as suggested by @CodyCBakerPhD [here](https://github.com/catalystneuro/nwb-conversion-tools/pull/364#issuecomment-1009162935)

## How to test the behavior?
Have setup tests for it in a separate file: https://github.com/catalystneuro/nwb-conversion-tools/blob/952b7f4a530bd8b62ab7185569060d8d1878087d/tests/test_internals/test_movie_interface.py

## Checklist

### DIFF
--- a/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
@@ -11,11 +11,9 @@ from pynwb.image import ImageSeries
 from hdmf.backends.hdf5.h5_utils import H5DataIO
 from hdmf.data_utils import DataChunkIterator
 
-from .movie_utils import get_movie_timestamps, get_movie_fps, get_frame_shape
 from ....basedatainterface import BaseDataInterface
 from ....utils.conversion_tools import check_regular_timestamps, get_module
 from ....utils.json_schema import get_schema_from_hdmf_class, get_base_schema
-
 
 try:
     import cv2
@@ -83,6 +81,7 @@ class MovieInterface(BaseDataInterface):
         chunk_data: bool = True,
         module_name: Optional[str] = None,
         module_description: Optional[str] = None,
+        compression: str = "gzip"
     ):
         """
         Convert the movie data files to ImageSeries and write them in the NWBFile.
@@ -105,6 +104,8 @@ class MovieInterface(BaseDataInterface):
             and may contain most keywords normally accepted by an ImageSeries
             (https://pynwb.readthedocs.io/en/stable/pynwb.image.html#pynwb.image.ImageSeries).
              The list for the 'Movies' key should correspond one to one to the movie files in the file_paths list.
+             If multiple movies need to be in the same ImageSeries, then supply the same value for "name" key.
+             Multiple movies in same ImageSeries is only supported if 'external_mode'=True.
         stub_test : bool, optional
             If True, truncates the write operation for fast testing. The default is False.
         external_mode : bool, optional
@@ -128,6 +129,8 @@ class MovieInterface(BaseDataInterface):
             If the processing module specified by module_name does not exist, it will be created with this description.
             The default description is the same as used by the conversion_tools.get_module function.
         """
+        from .movie_utils import VideoCaptureContext
+
         file_paths = self.source_data["file_paths"]
 
         if stub_test:
@@ -151,88 +154,83 @@ class MovieInterface(BaseDataInterface):
             f"({len(image_series_kwargs_list)}) vs. file_paths ({len(self.source_data['file_paths'])})!"
         )
 
-        for j, file in enumerate(file_paths):
-            timestamps = starting_times[j] + get_movie_timestamps(movie_file=file)
+        # check for duplicates in image_series_kwargs list keys:
+        def _check_duplicates(image_series_kwargs_list):
+            image_series_kwargs_list_keys = [i["name"] for i in image_series_kwargs_list]
+            if len(set(image_series_kwargs_list_keys)) < len(image_series_kwargs_list_keys):
+                assert external_mode, "for multiple video files under the same ImageSeries name, use exernal_mode=True"
+            keys_set = []
+            image_series_kwargs_list_unique = []
+            for no, image_series_kwargs in enumerate(image_series_kwargs_list):
+                if image_series_kwargs["name"] not in keys_set:
+                    keys_set.append(image_series_kwargs["name"])
+                    image_series_kwargs_list_unique.append(dict(image_series_kwargs, data=[file_paths[no]]))
+                else:
+                    idx = keys_set.index(image_series_kwargs["name"])
+                    image_series_kwargs_list_unique[idx]["data"].append(file_paths[no])
+            return image_series_kwargs_list_unique
 
-            if len(starting_times) != len(file_paths):
-                starting_times.append(timestamps[-1])
+        image_series_kwargs_list_updated = _check_duplicates(image_series_kwargs_list)
 
-            image_series_kwargs = dict(image_series_kwargs_list[j])
-            if check_regular_timestamps(ts=timestamps):
-                fps = get_movie_fps(movie_file=file)
-                image_series_kwargs.update(starting_time=starting_times[j], rate=fps)
-            else:
-                image_series_kwargs.update(timestamps=H5DataIO(timestamps, compression="gzip"))
-
+        for j, image_series_kwargs in enumerate(image_series_kwargs_list_updated):
+            file_list = image_series_kwargs.pop("data")
             if external_mode:
-                image_series_kwargs.update(format="external", external_file=[file])
+                image_series_kwargs.update(format="external", external_file=file_list)
+                with VideoCaptureContext(str(file_list[0]), stub=stub_test) as vc:
+                    fps = vc.get_movie_fps()
+                    image_series_kwargs.update(starting_time=0.0, rate=fps)  # TODO manage custom starting_times
             else:
+                file = file_list[0]
                 uncompressed_estimate = Path(file).stat().st_size * 70
                 available_memory = psutil.virtual_memory().available
-                if not chunk_data and uncompressed_estimate >= available_memory:
+                if not chunk_data and not stub_test and uncompressed_estimate >= available_memory:
                     warn(
                         f"Not enough memory (estimated {round(uncompressed_estimate/1e9, 2)} GB) to load movie file as "
                         f"array ({round(available_memory/1e9, 2)} GB available)! Forcing chunk_data to True."
                     )
                     chunk_data = True
-
-                total_frames = len(timestamps)
-                frame_shape = get_frame_shape(movie_file=file)
-                maxshape = [total_frames]
-                maxshape.extend(frame_shape)
-                best_gzip_chunk = (1, frame_shape[0], frame_shape[1], 3)
-                tqdm_pos, tqdm_mininterval = (0, 10)
+                video_capture_ob = VideoCaptureContext(str(file), stub=stub_test)
+                total_frames = video_capture_ob.get_movie_frame_count()
                 if chunk_data:
-
-                    def data_generator(file, count_max):
-                        cap = cv2.VideoCapture(str(file))
-                        for _ in range(min(count_max, total_frames)):
-                            success, frame = cap.read()
-                            yield frame
-                        cap.release()
-
-                    mov = DataChunkIterator(
-                        data=tqdm(
-                            iterable=data_generator(file=file, count_max=count_max),
-                            desc=f"Copying movie data for {Path(file).name}",
-                            position=tqdm_pos,
-                            total=total_frames,
-                            mininterval=tqdm_mininterval,
-                        ),
+                    frame_shape = video_capture_ob.get_frame_shape()
+                    maxshape = (total_frames, *frame_shape)
+                    best_gzip_chunk = (1, frame_shape[0], frame_shape[1], 3)
+                    iterable = DataChunkIterator.from_iterable(
+                        data=video_capture_ob,
                         iter_axis=0,  # nwb standard is time as zero axis
                         maxshape=tuple(maxshape),
+                        dtype=video_capture_ob.get_movie_frame_dtype(),
                     )
-                    image_series_kwargs.update(data=H5DataIO(mov, compression="gzip", chunks=best_gzip_chunk))
+                    data = H5DataIO(iterable, compression=compression, chunks=best_gzip_chunk)
                 else:
-                    cap = cv2.VideoCapture(str(file))
-                    mov = []
+                    iterable = []
+                    tqdm_pos, tqdm_mininterval = (0, 10)
                     with tqdm(
                         desc=f"Reading movie data for {Path(file).name}",
                         position=tqdm_pos,
                         total=total_frames,
                         mininterval=tqdm_mininterval,
                     ) as pbar:
-                        for _ in range(min(count_max, total_frames)):
-                            success, frame = cap.read()
-                            mov.append(frame)
+                        for frame in video_capture_ob:
+                            iterable.append(frame)
                             pbar.update(1)
-                    cap.release()
-                    image_series_kwargs.update(
-                        data=H5DataIO(
-                            DataChunkIterator(
-                                tqdm(
-                                    iterable=np.array(mov),
-                                    desc=f"Writing movie data for {Path(file).name}",
-                                    position=tqdm_pos,
-                                    mininterval=tqdm_mininterval,
-                                ),
-                                iter_axis=0,  # nwb standard is time as zero axis
-                                maxshape=tuple(maxshape),
-                            ),
-                            compression="gzip",
-                            chunks=best_gzip_chunk,
-                        )
-                    )
+                    iterable = np.array(iterable)
+
+                    data = H5DataIO(iterable, compression=compression)
+
+                    # capture data in kwargs:
+                    image_series_kwargs.update(data=data)
+                # capture time info in kwargs:
+                with VideoCaptureContext(str(file), stub=stub_test) as vc:
+                    timestamps = starting_times[j] + vc.get_movie_timestamps()
+                    if len(starting_times) != len(file_paths):
+                        starting_times.append(timestamps[-1])
+                    if check_regular_timestamps(ts=timestamps):
+                        fps = vc.get_movie_fps()
+                        image_series_kwargs.update(starting_time=starting_times[j], rate=fps)
+                    else:
+                        image_series_kwargs.update(timestamps=timestamps)
+
             if module_name is None:
                 nwbfile.add_acquisition(ImageSeries(**image_series_kwargs))
             else:
--- a/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
@@ -218,8 +218,8 @@ class MovieInterface(BaseDataInterface):
 
                     data = H5DataIO(iterable, compression=compression)
 
-                    # capture data in kwargs:
-                    image_series_kwargs.update(data=data)
+                # capture data in kwargs:
+                image_series_kwargs.update(data=data)
                 # capture time info in kwargs:
                 with VideoCaptureContext(str(file), stub=stub_test) as vc:
                     timestamps = starting_times[j] + vc.get_movie_timestamps()
--- a/tests/test_internals/test_movie_interface.py
+++ b/tests/test_internals/test_movie_interface.py
@@ -64,16 +64,14 @@ def assert_nwbfile_conversion(
     module_name: str = None,
     module_description: str = None,
     metadata: dict = None,
-    iterator_type: str = "v2",
     stub_test: bool = False,
     external_mode: bool = False,
-    chunk_data: bool = True,
+    chunk_data: bool = False,
 ):
     metadata = converter.get_metadata() if metadata is None else metadata
     custom_names = [metadata["Behavior"]["Movies"][i]["name"] for i in range(len(metadata["Behavior"]["Movies"]))]
     conversion_opts = dict(
         starting_times=starting_times,
-        iterator_type=iterator_type,
         stub_test=stub_test,
         external_mode=external_mode,
         chunk_data=chunk_data,
@@ -116,6 +114,7 @@ def test_conversion_default(movie_converter, create_movies, nwbfile_path):
 
 def test_conversion_custom(movie_converter, nwbfile_path):
     if HAVE_OPENCV:
+        starting_times = [np.float(np.random.randint(200)) for i in range(len(create_movies))]
         module_name = "TestModule"
         module_description = "This is a test module."
         assert_nwbfile_conversion(
@@ -123,17 +122,22 @@ def test_conversion_custom(movie_converter, nwbfile_path):
             nwbfile_path=nwbfile_path,
             module_description=module_description,
             module_name=module_name,
+            starting_times=starting_times
         )
 
 
 def test_conversion_options(movie_converter, nwbfile_path):
     if HAVE_OPENCV:
+        starting_times = [np.float(np.random.randint(200)) for i in range(len(create_movies))]
         conversion_options_testing_matrix = [
-            dict(external_mode=False, stub_test=True, chunk_data=i, iterator_type=j)
-            for i, j in product([True, False], ["v1", "v2"])
+            dict(external_mode=False, stub_test=True, chunk_data=i)
+            for i in [True, False]
         ]
         for conv_ops in conversion_options_testing_matrix:
-            assert_nwbfile_conversion(converter=movie_converter, nwbfile_path=nwbfile_path, **conv_ops)
+            assert_nwbfile_conversion(converter=movie_converter,
+                                      nwbfile_path=nwbfile_path,
+                                      starting_times=starting_times,
+                                      **conv_ops)
 
 
 def test_conversion_external_mode(movie_converter, nwbfile_path):
--- a/tests/test_internals/test_movie_interface.py
+++ b/tests/test_internals/test_movie_interface.py
@@ -1,0 +1,141 @@
+from itertools import product
+
+import numpy as np
+import pytest
+from pynwb import NWBHDF5IO
+
+from nwb_conversion_tools import NWBConverter, MovieInterface
+
+try:
+    import cv2
+
+    HAVE_OPENCV = True
+except ImportError:
+    HAVE_OPENCV = False
+
+
+@pytest.fixture(scope="module")
+def create_movies(tmp_path_factory):
+    if HAVE_OPENCV:
+        base_path = tmp_path_factory.mktemp("movie_tests")
+        movie_file1 = base_path / "test1.avi"
+        movie_file2 = base_path / "test2.avi"
+        (nf, nx, ny) = (50, 640, 480)
+        writer1 = cv2.VideoWriter(
+            filename=str(movie_file1),
+            apiPreference=None,
+            fourcc=cv2.VideoWriter_fourcc("M", "J", "P", "G"),
+            fps=25,
+            frameSize=(ny, nx),
+            params=None,
+        )
+        writer2 = cv2.VideoWriter(
+            filename=str(movie_file2),
+            apiPreference=None,
+            fourcc=cv2.VideoWriter_fourcc("M", "J", "P", "G"),
+            fps=25,
+            frameSize=(ny, nx),
+            params=None,
+        )
+        for k in range(nf):
+            writer1.write(np.random.randint(0, 255, (nx, ny, 3)).astype("uint8"))
+            writer2.write(np.random.randint(0, 255, (nx, ny, 3)).astype("uint8"))
+        writer1.release()
+        writer2.release()
+        return [str(movie_file1), str(movie_file2)]
+
+
+@pytest.fixture(scope="module")
+def movie_converter(create_movies):
+    if HAVE_OPENCV:
+
+        class MovieTestNWBConverter(NWBConverter):
+            data_interface_classes = dict(Movie=MovieInterface)
+
+        source_data = dict(Movie=dict(file_paths=create_movies))
+        converter = MovieTestNWBConverter(source_data)
+        return converter
+
+
+def assert_nwbfile_conversion(
+    converter,
+    nwbfile_path,
+    starting_times: list = None,
+    module_name: str = None,
+    module_description: str = None,
+    metadata: dict = None,
+    iterator_type: str = "v2",
+    stub_test: bool = False,
+    external_mode: bool = False,
+    chunk_data: bool = True,
+):
+    metadata = converter.get_metadata() if metadata is None else metadata
+    custom_names = [metadata["Behavior"]["Movies"][i]["name"] for i in range(len(metadata["Behavior"]["Movies"]))]
+    conversion_opts = dict(
+        starting_times=starting_times,
+        iterator_type=iterator_type,
+        stub_test=stub_test,
+        external_mode=external_mode,
+        chunk_data=chunk_data,
+    )
+    if module_name:
+        conversion_opts.update(module_name=module_name)
+    if module_description:
+        conversion_opts.update(module_description=module_description)
+    converter.run_conversion(
+        metadata=metadata, nwbfile_path=nwbfile_path, overwrite=True, conversion_options=dict(Movie=conversion_opts)
+    )
+    module_name = module_name if module_name is not None else "acquisition"
+    with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
+        nwbfile = io.read()
+        mod = nwbfile.modules[module_name]
+        assert module_name in mod
+        if module_description:
+            assert module_description == mod.description
+        for no, name in enumerate(custom_names):
+            if external_mode:
+                for file_name in converter.data_interfaces["Movie"].source_data["file_paths"]:
+                    assert file_name in mod[name].external_file
+            else:
+                assert name in mod
+            if starting_times:
+                assert starting_times[no] == mod[name].starting_time
+
+
+@pytest.fixture(scope="module")
+def nwbfile_path(tmp_path_factory):
+    nwbfile_path = str(tmp_path_factory.mktemp("movie_tests") / "test.nwb")
+    return nwbfile_path
+
+
+def test_conversion_default(movie_converter, create_movies, nwbfile_path):
+    if HAVE_OPENCV:
+        starting_times = [np.float(np.random.randint(200)) for i in range(len(create_movies))]
+        assert_nwbfile_conversion(converter=movie_converter, nwbfile_path=nwbfile_path, starting_times=starting_times)
+
+
+def test_conversion_custom(movie_converter, nwbfile_path):
+    if HAVE_OPENCV:
+        module_name = "TestModule"
+        module_description = "This is a test module."
+        assert_nwbfile_conversion(
+            converter=movie_converter,
+            nwbfile_path=nwbfile_path,
+            module_description=module_description,
+            module_name=module_name,
+        )
+
+
+def test_conversion_options(movie_converter, nwbfile_path):
+    if HAVE_OPENCV:
+        conversion_options_testing_matrix = [
+            dict(external_mode=False, stub_test=True, chunk_data=i, iterator_type=j)
+            for i, j in product([True, False], ["v1", "v2"])
+        ]
+        for conv_ops in conversion_options_testing_matrix:
+            assert_nwbfile_conversion(converter=movie_converter, nwbfile_path=nwbfile_path, **conv_ops)
+
+
+def test_conversion_external_mode(movie_converter, nwbfile_path):
+    if HAVE_OPENCV:
+        assert_nwbfile_conversion(converter=movie_converter, nwbfile_path=nwbfile_path, external_mode=True)
--- a/nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
+++ b/nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
@@ -1,7 +1,9 @@
-"""Authors: Cody Baker."""
+"""Authors: Saksham Sharda, Cody Baker."""
 from pathlib import Path
+from typing import Union, Tuple
+
 import numpy as np
-from typing import Union
+from tqdm import tqdm
 
 try:
     import cv2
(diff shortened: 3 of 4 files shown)


Saksham20:
TODO:

- [ ] fix segmentation errors in tests

Saksham20:
@CodyCBakerPhD you had [added this function](https://github.com/catalystneuro/nwb-conversion-tools/blob/335c71f54ca241e7178d99bbc6ad00aff163aa85/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py#L239) to add a custom processing module to hold the `ImageSeries`. But we would need to put the ImageSeries under a NWBContainer, NWBDataInterface. Calling `mod.add(ImageSeries)` creates an error.
I'm not sure which existing data interface holds a list of `ImageSeries` type.

CodyCBakerPhD:
@Saksham20 If you can write a test to show how `.add` or `.add_interface` causes an error in the way you describe that would be the best way to understand the issue you're having. See https://github.com/catalystneuro/nwb-conversion-tools/blob/335c71f54ca241e7178d99bbc6ad00aff163aa85/tests/test_internals/test_interfaces.py#L178-L184 for the current test of this functionality

tests/test_internals/test_movie_interface.py
```diff
@@ -0,0 +1,143 @@
...
+    import cv2
+
+    HAVE_OPENCV = True
+except ImportError:
+    HAVE_OPENCV = False
+
+
+@pytest.fixture(scope="module")
+def create_movies(tmp_path_factory):
+    if HAVE_OPENCV:
```
CodyCBakerPhD, line 19: Could you implement this testing refactor through `unittest.TestCase`? See https://github.com/catalystneuro/nwb-conversion-tools/issues/287

Using common `setUp` techniques could simplify a lot of the logic, and it's always good to have separate tests for each option/use-case as you are starting to separate these

CodyCBakerPhD:
@Saksham20 Mark as ready for review once tests are passing; I'm also kind of curious what is so severe here that we get segmentation faults?

tests/test_internals/test_movie_interface.py
```diff
@@ -83,13 +83,15 @@ def assert_nwbfile_conversion(
...
         nwbfile = io.read()
-        mod = nwbfile.modules[module_name]
-        assert module_name in mod
-        if module_description:
-            assert module_description == mod.description
+        if module_name is None:
+            mod = nwbfile.acquisition
+        else:
+            assert module_name in nwbfile.processing
+            mod = nwbfile.processing[module_name]
```
bendichter, line 92: maybe it would be better to use `get_module` here

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -220,16 +220,14 @@ def _check_duplicates(image_series_kwargs_list):
...
-                    else:
-                        image_series_kwargs.update(timestamps=timestamps)
+                timestamps = starting_times[j] + video_capture_ob.get_movie_timestamps()
+                if len(starting_times) != len(file_paths):
+                    starting_times.append(timestamps[-1])
+                if check_regular_timestamps(ts=timestamps):
+                    fps = video_capture_ob.get_movie_fps()
+                    image_series_kwargs.update(starting_time=starting_times[j], rate=fps)
+                else:
+                    image_series_kwargs.update(timestamps=timestamps)
```
bendichter, line 230: what is wrong with the way you were doing it before?

tests/test_internals/test_movie_interface.py
Saksham20: get module [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/cfc01eb11868c7d666ec5dca4e6d0ef251bac538/nwb_conversion_tools/utils/conversion_tools.py#L13) creates the processing module if it does not exist. This test expects that module name to exist already.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: before, we were using `DataChunkIterator` for both `chunk_data=True/False`. 
When chunking is disabled then we should not use the `DataChunkIterator` at all and just wrap the array with HDF5IO to compress. What do you think @CodyCBakerPhD since you had set it up like this initially?

Additionally, in another PR we would be using the custom `MovieDataChunkIterator ` which inherits from `GenericDataChunkIterator` [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/main/nwb_conversion_tools/utils/genericdatachunkiterator.py). That you have a built-in progress bar right @CodyCBakerPhD ?

tests/test_internals/test_movie_interface.py
Saksham20: so do we plan to use Unittest all across as a convention? 
The fixtures behave like the `setUp` and with the `yield` statement, they also have a teardown functionality by running the code after yield during cleanup. I have appropriately kept some fixtures with module level scope and others with function scope (default for `purest.fixture`). 
I find `pytest` easy to use, is there a reason why you think `unittest` is better?
I have made some changes from your last comment, please review: https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/64cf4639f0173ba82837eb26a6bfe1c22cf43073

CodyCBakerPhD: Convention, yes, requirement, no.

Overall it offers superior code organization since you can group similar tests (data requirements, behavioral methods, etc.) into similar classes all in the same file. Also `hdmf.testing TestCase` (child class of `unittest`) already has a ton of added useful functionality for various things, such as testing exact assertions. It's also easier for new users to understand, since a `yield` -> clean-up approach actually gets pretty complicated, as does defining multiple helper functions across the scope for things that would otherwise be simple attributes of the class.

I grant you, the original tests for this interface were not well organized either way so anything can be done here to improve them; but in the future if we do attempt a refactor with `unittest`, I'd swap to it if it offers clearer organization or code reduction.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD: Not really sure what code change this selection is referring to, but for Saksham's comments:

there was a reason behind wrapping the DataChunkIterator due, I recall, to get the `tqdm` progress bar to work correctly as it wrote the actual chunks (in the HDF5 Dataset sense) to the NWBFile (line https://github.com/catalystneuro/nwb-conversion-tools/blob/main/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py#L224 ensures no actual buffering is going on); I agree that if there is an alternative way of doing it without that then it's not needed, but remembering this PR is focused on the context class alone I think a change to the `chunk_data` option can wait for the other PR refactoring the `MovieDataInterface`.

bendichter: > When chunking is disabled then we should not use the DataChunkIterator at all and just wrap the array with HDF5IO to compress. 

This is not correct. "chunking" means different things in these two different contexts.

> but remembering this PR is focused on the context class alone

yes, this change is incorrect, needs to be removed, and is holding up the rest of the PR

CodyCBakerPhD: > This is not correct. "chunking" means different things in these two different contexts.

Yeah, this really should be changed to `buffer_data` as well. Might even want to not make it boolean, and instead apply a size scale similar to `buffer_gb`; if the user wants to load the entire movie, they can just set an arbitrarily (but hopefully not unsafe) buffer size.

Saksham20: I have added back the `DataChunkIterator` like it was before this PR. https://github.com/catalystneuro/nwb-conversion-tools/pull/372/files#diff-57b78875c81d904caa745de93f7771de666b821e9d52586cd03a8328e7a74c6bR226

CodyCBakerPhD:
@Saksham20 Any clue what is causing these segmentation faults (one of the worst errors that can occur via python)? I'd really like to have a full understanding of why they are occurring and how to avoid them before OK'ing these changes, since this has been a problem since the first commit.

CodyCBakerPhD:
@Saksham20 It occurred to me that it may be safer, for the context class, to not inherit from the `VideoCapture` but rather be a stand-alone context class that contains an attribute object that is a `VideoCapture`, similar to how our `DataInterfaces` contain `{submodality}_extractors}`. Then just define behavior explicitly for enter/exit, init/close that wrap the equivalent operations on the internal object; that way it's not inheriting any extra magic that may be causing the isssues.

What do you think? Also were you still planning on a separate PR for introducing the context with unit tests and no interface integration?

Saksham20:
> @Saksham20 It occurred to me that it may be safer, for the context class, to not inherit from the `VideoCapture` but rather be a stand-alone context class that contains an attribute object that is a `VideoCapture`, similar to how our `DataInterfaces` contain `{submodality}_extractors}`. Then just define behavior explicitly for enter/exit, init/close that wrap the equivalent operations on the internal object; that way it's not inheriting any extra magic that may be causing the isssues.
> 
> What do you think? Also were you still planning on a separate PR for introducing the context with unit tests and no interface integration?

yes, a good point, I'll try not inheriting from VideoCapture and see if it works. I still would want to keep #388 as a separate PR for now

Saksham20:
> > @Saksham20 It occurred to me that it may be safer, for the context class, to not inherit from the `VideoCapture` but rather be a stand-alone context class that contains an attribute object that is a `VideoCapture`, similar to how our `DataInterfaces` contain `{submodality}_extractors}`. Then just define behavior explicitly for enter/exit, init/close that wrap the equivalent operations on the internal object; that way it's not inheriting any extra magic that may be causing the isssues.
> > What do you think? Also were you still planning on a separate PR for introducing the context with unit tests and no interface integration?
> 
> yes, a good point, I'll try not inheriting from VideoCapture and see if it works. I still would want to keep #388 as a separate PR for now

@CodyCBakerPhD , I have added the VIdeoCapture as an attribute and there seem to be no seg faults. Lets continue with this PR. I have closed #388 . Could you review this and the lets merge

CodyCBakerPhD:
> @CodyCBakerPhD , I have added the VIdeoCapture as an attribute and there seem to be no seg faults. Lets continue with this PR. I have closed #388 . Could you review this and the lets merge

I would still **greatly appreciate** a separate pre-PR that does only these things:
- [ ] Defines this context class as you have done
- [ ] Under `tests/test_internals/movie_util_unit_tests.py`, add one unit test for each expected behavior of this class.
- [ ] This pre-PR would not integrate at all with the MovieDataInterface, which is the main purpose of this PR.

I'll go ahead and peel the MovieInterface tests out of `test_interfaces` to alleviate general changelog inflation for both PRs

Saksham20:
@CodyCBakerPhD I think the testing workflows are not running correctly after your automated changes

CodyCBakerPhD:
@Saksham20 They are never triggered by 'Automated Changes' since no actual functional behavior is changed by running `black`, so it would be a waste of computational resources. You can see they are triggered and passing from looking at individual commits https://github.com/catalystneuro/nwb-conversion-tools/runs/5269356329?check_suite_focus=true, in particular the one right before the  'Automated Changes' was triggered.

The bug is actually with the web display for the PR; it expects the tests to run on each and every push. I might swap to an external CI like we use on NWBInspector now, since it does resolve that issue. However, the ultimate behavior is the same in that you can always avoid triggering it by calling `black .` on the directory before adding+commiting+pushing changes.

CodyCBakerPhD:
Reminder to self: squash this PR prior to merge.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -129,12 +121,10 @@ def run_conversion(
             If the processing module specified by module_name does not exist, it will be created with this description.
             The default description is the same as used by the conversion_tools.get_module function.
         """
+        from .movie_utils import VideoCaptureContext
```
CodyCBakerPhD, line 124: Is there a reason to do this import local to the function? I think it would be fine to invoke up at the top

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -84,6 +73,7 @@ def run_conversion(
         chunk_data: bool = True,
         module_name: Optional[str] = None,
         module_description: Optional[str] = None,
+        compression: str = "gzip",
```
CodyCBakerPhD, line 76:
```suggestion
        compression: str = "gzip",
        compression_options: Optional[dict] = None,
```
Might as well include compression options as well.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -106,6 +96,8 @@ def run_conversion(
             and may contain most keywords normally accepted by an ImageSeries
             (https://pynwb.readthedocs.io/en/stable/pynwb.image.html#pynwb.image.ImageSeries).
              The list for the 'Movies' key should correspond one to one to the movie files in the file_paths list.
+             If multiple movies need to be in the same ImageSeries, then supply the same value for "name" key.
+             Multiple movies in same ImageSeries is only supported if 'external_mode'=True.
```
CodyCBakerPhD, line 100:
```suggestion
             Multiple movies in the same ImageSeries is only supported if 'external_mode'=True.
```

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,49 +142,55 @@ def run_conversion(
             f"({len(image_series_kwargs_list)}) vs. file_paths ({len(self.source_data['file_paths'])})!"
         )
 
-        for j, file in enumerate(file_paths):
-            timestamps = starting_times[j] + get_movie_timestamps(movie_file=file)
-
-            if len(starting_times) != len(file_paths):
-                starting_times.append(timestamps[-1])
+        # check for duplicates in image_series_kwargs list keys:
```
CodyCBakerPhD, line 145:
```suggestion
```
Code below comment is self-explanatory

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,49 +142,55 @@ def run_conversion(
...
+        def _check_duplicates(image_series_kwargs_list):
+            image_series_kwargs_list_keys = [i["name"] for i in image_series_kwargs_list]
+            if len(set(image_series_kwargs_list_keys)) < len(image_series_kwargs_list_keys):
+                assert external_mode, "for multiple video files under the same ImageSeries name, use exernal_mode=True"
```
CodyCBakerPhD, line 149:
```suggestion
                assert external_mode, "For multiple video files under the same ImageSeries name, use exernal_mode=True."
```

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,49 +142,55 @@ def run_conversion(
...
+            keys_set = []
+            image_series_kwargs_list_unique = []
+            for no, image_series_kwargs in enumerate(image_series_kwargs_list):
```
CodyCBakerPhD, line 152:
```suggestion
            for n, image_series_kwargs in enumerate(image_series_kwargs_list):
```
I'd avoid using `no` since it's a standalone English word with different contextual meaning related to boolean archtypes.

Some common iteration variable names, especially for `enumerate`, are `i` (though more common in C/C++/C#), `j`, `k`, `n`, `m`. Basically anything that is also a common integer indexing variable from discrete & combinatorial mathematics.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,49 +142,55 @@ def run_conversion(
...
+                if image_series_kwargs["name"] not in keys_set:
+                    keys_set.append(image_series_kwargs["name"])
+                    image_series_kwargs_list_unique.append(dict(image_series_kwargs, data=[file_paths[no]]))
```
CodyCBakerPhD, line 155:
```suggestion
                    image_series_kwargs_list_unique.append(dict(image_series_kwargs, data=[file_paths[n]]))
```

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: yes, thanks for that. Indeed we can use compression options to input to `H5DataIO` https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/bb68d9c4c651252350560d2b0104e09ad1a8adc1

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: yeah it can move up, https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/8b8ca554b247cd2ff537a2361e47dde9001c1188

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -128,13 +122,15 @@ def run_conversion(
         module_description: str, optional
             If the processing module specified by module_name does not exist, it will be created with this description.
             The default description is the same as used by the conversion_tools.get_module function.
+        compression: str, optional
+            compression strategy to use for HFDataIO.
+            https://hdmf.readthedocs.io/en/latest/hdmf.backends.hdf5.h5_utils.html#hdmf.backends.hdf5.h5_utils.H5DataIO
+        compression_options: int, optional
+            parameter for compression filter. See HFDataIO doc.
+            https://hdmf.readthedocs.io/en/latest/hdmf.backends.hdf5.h5_utils.html#hdmf.backends.hdf5.h5_utils.H5DataIO
```
CodyCBakerPhD, line 130:
```suggestion
            Compression strategy to use for HFDataIO. For full list of currently supported filters, see
            https://docs.h5py.org/en/latest/high/dataset.html#lossless-compression-filters
        compression_options: int, optional
            Parameter(s) for compression filter. Currently only supports the compression level (integer from 0 to 9) of
            compression="gzip".
```

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -84,6 +74,8 @@ def run_conversion(
         chunk_data: bool = True,
         module_name: Optional[str] = None,
         module_description: Optional[str] = None,
+        compression: Optional[str] = "gzip",
+        compression_options: Optional[dict] = None,
```
CodyCBakerPhD, line 78:
```suggestion
        compression_options: Optional[int] = None,
```
As correctly indicated in the documentation, this only currently supports `int` values for `gzip`.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,49 +148,54 @@ def run_conversion(
...
+        image_series_kwargs_list_updated = _check_duplicates(image_series_kwargs_list)
 
+        for j, image_series_kwargs in enumerate(image_series_kwargs_list_updated):
+            file_list = image_series_kwargs.pop("data")
             if external_mode:
-                image_series_kwargs.update(format="external", external_file=[file])
+                image_series_kwargs.update(format="external", external_file=file_list)
+                with VideoCaptureContext(str(file_list[0])) as vc:
+                    fps = vc.get_movie_fps()
+                image_series_kwargs.update(starting_time=0.0, rate=fps)  # TODO manage custom starting_times
```
CodyCBakerPhD, line 174: Note the `starting_times` input argument; can you utilize this?

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD: Of course, it will need aliasing in the actual `H5DataIO` call since there it is called ` compression_opts` (less descriptive imo)

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,49 +148,54 @@ def run_conversion(
...
-
-                    def data_generator(file, count_max):
-                        cap = cv2.VideoCapture(str(file))
-                        for _ in range(min(count_max, total_frames)):
-                            success, frame = cap.read()
-                            yield frame
-                        cap.release()
-
-                    mov = DataChunkIterator(
+                    _ = video_capture_ob.__enter__()
```
CodyCBakerPhD, line 195: This is very highly irregular usage of the magic function. Is it not able to be opened and used as an iterable like in https://github.com/catalystneuro/nwb-conversion-tools/blob/main/tests/test_internals/test_movie_utils.py#L71-L78?

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
...
-                    image_series_kwargs.update(data=H5DataIO(mov, compression="gzip", chunks=best_gzip_chunk))
+                    data = H5DataIO(
+                        iterable,
+                        compression=compression,
+                        compression_options=compression_options,
+                        chunks=best_gzip_chunk,
+                    )
+                    data = H5DataIO(
+                        iterable, compression=compression, compression_opts=compression_options, chunks=best_gzip_chunk
+                    )
```
CodyCBakerPhD, line 215:
```suggestion
                    data = H5DataIO(
                        iterable, compression=compression, compression_opts=compression_options, chunks=best_gzip_chunk
                    )
```
?

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
...
-                            DataChunkIterator(
-                                tqdm(
-                                    iterable=np.array(mov),
-                                    desc=f"Writing movie data for {Path(file).name}",
-                                    position=tqdm_pos,
-                                    mininterval=tqdm_mininterval,
-                                ),
-                                iter_axis=0,  # nwb standard is time as zero axis
-                                maxshape=tuple(maxshape),
+                    iterable = []
```
CodyCBakerPhD, line 217:
```suggestion
                    iterable = np.zeros(shape=maxshape, dtype="uint8")
```
For better performance (mainly to avoid allocating the memory twice), let's just allocate a single array and fill its values iteratively.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
                         iter_axis=0,  # nwb standard is time as zero axis
                         maxshape=tuple(maxshape),
```
CodyCBakerPhD, line 205: Does this still need to be cast as a tuple?

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
...
+                    iterable = []
+                    with VideoCaptureContext(str(file)) as video_capture_ob:
+                        with tqdm(
+                            desc=f"Reading movie data for {Path(file).name}",
+                            position=tqdm_pos,
+                            total=total_frames,
+                            mininterval=tqdm_mininterval,
+                        ) as pbar:
+                            for frame in video_capture_ob:
+                                iterable.append(frame)
```
CodyCBakerPhD, line 226:
```suggestion
                            for n, frame in enumerate(video_capture_ob):
                                iterable[n, :, :, :] = frame
```

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
...
+                                pbar.update(1)
+                    data = H5DataIO(
+                        DataChunkIterator(
+                            tqdm(
+                                iterable=np.array(iterable),
```
CodyCBakerPhD, line 231:
```suggestion
                                iterable=iterable,
```

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
...
+                                desc=f"Writing movie data for {Path(file).name}",
+                                position=tqdm_pos,
+                                mininterval=tqdm_mininterval,
                             ),
-                            compression="gzip",
-                            chunks=best_gzip_chunk,
-                        )
+                            iter_axis=0,  # nwb standard is time as zero axis
+                            maxshape=tuple(maxshape),
```
CodyCBakerPhD, line 237: Does this still need to be casted as a tuple?

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,49 +148,54 @@ def run_conversion(
...
+            keys_set = []
+            image_series_kwargs_list_unique = []
+            for n, image_series_kwargs in enumerate(image_series_kwargs_list):
+                if image_series_kwargs["name"] not in keys_set:
+                    keys_set.append(image_series_kwargs["name"])
+                    image_series_kwargs_list_unique.append(dict(image_series_kwargs, data=[file_paths[n]]))
+                else:
+                    idx = keys_set.index(image_series_kwargs["name"])
+                    image_series_kwargs_list_unique[idx]["data"].append(file_paths[n])
+            return image_series_kwargs_list_unique
```
CodyCBakerPhD, line 164: The main goal of this, as I understand, is to handle multiple file path inputs to a single ImageSeries object via external mode.

It's misleading to fold in the `data` keyword to the `image_series_kwargs` in this manner, when the `data` field is actually nothing like that even when you use `external_mode=True`, which is what leads to the awkward popping in the iteration of `image_series_kwargs_list_updated` only to be properly re-added at the end of the method when `external_mode=False`.

It might be better design to simply use this like `file_list = _extract_file_list(image_series_kwargs_list)` which can then return the `file_list` without any popping, and can still assert uniqueness of names when `external_mode=False`. This would also make the code more readable as well.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD, line 174:
```suggestion
                with VideoCaptureContext(str(file_list[0])) as vc:
                    fps = vc.get_movie_fps()
                image_series_kwargs.update(starting_time=0.0, rate=fps, format="external", external_file=file_list)  # TODO manage custom starting_times
```
Code reduction.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
...
+                        ),
+                        compression="gzip",
+                        compression_opts=compression_options,
+                        chunks=best_gzip_chunk,
                     )
+
+                # capture data in kwargs:
+                image_series_kwargs.update(data=data)
+                if len(starting_times) != len(file_paths):
+                    starting_times.append(timestamps[-1])
```
CodyCBakerPhD, line 247:
```suggestion
```
Looks like this is being asserted at the start of the method and so can never enter the logic.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -203,37 +204,52 @@ def data_generator(file, count_max):
...
+                            iter_axis=0,  # nwb standard is time as zero axis
+                            maxshape=tuple(maxshape),
+                        ),
+                        compression="gzip",
+                        compression_opts=compression_options,
+                        chunks=best_gzip_chunk,
                     )
+
+                # capture data in kwargs:
+                image_series_kwargs.update(data=data)
```
CodyCBakerPhD, line 245:
```suggestion
```
Self-explanatory from code

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: yes that was pending , fixed it here https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/921b050cad40da93b297c3dd690e1b8763f576f9

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: not necessary

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: this just reinstantiates the `self.vc=cv2.VideoCapture()`, but no the whole object. We could however reinstantiate since its not expensive

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: yes we dont need to do that
https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/afbde8305f560b1e40a4c3aad1848e2e24b2d77a

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD: Yeah, I would always re-instantiate explicitly; the `__enter__` method really should never be called explicitly like that

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: yeah okay, indeed its not very readable. See here https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/62b39d288848d34462fe08dab9db7eb540f506cc

Saksham20: yes I agree its not very readable : https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/62b39d288848d34462fe08dab9db7eb540f506cc

CodyCBakerPhD:
@Saksham20 Just about there. Always be sure to check the [codecov reports](https://app.codecov.io/gh/catalystneuro/nwb-conversion-tools/blob/add_context_class_movie/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py) for any uncovered lines that relate to expected functional behavior. Can you:

- [ ] Add a test for duplicated metadata names with `external_mode=True` (https://github.com/catalystneuro/nwb-conversion-tools/pull/372/files#diff-57b78875c81d904caa745de93f7771de666b821e9d52586cd03a8328e7a74c6bR160-R161)
- [ ] `starting_times = None` (https://github.com/catalystneuro/nwb-conversion-tools/pull/372/files#diff-57b78875c81d904caa745de93f7771de666b821e9d52586cd03a8328e7a74c6bR173)

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,88 +144,113 @@ def run_conversion(
...
+            starting_times = [0.0] * len(image_series_kwargs_list_updated)
 
+        for j, (image_series_kwargs, file_list) in enumerate(zip(image_series_kwargs_list_updated, file_paths_list)):
             if external_mode:
-                image_series_kwargs.update(format="external", external_file=[file])
+                with VideoCaptureContext(str(file_list[0])) as vc:
+                    fps = vc.get_movie_fps()
+                image_series_kwargs.update(
+                    starting_time=starting_times[j], rate=fps, format="external", external_file=file_list
+                )
```
CodyCBakerPhD, line 181: If https://github.com/catalystneuro/nwb-conversion-tools/pull/372/files#diff-57b78875c81d904caa745de93f7771de666b821e9d52586cd03a8328e7a74c6bR173 is triggered here, all the `starting_times` for all the separate `ImageSeries` will be `0.0`.

The logic of the previous interface was to at least assume that the videos were consecutive, and adjusting the `starting_times` of each so that they start when the last one ended. You can calculate the time of that by also pulling the total number of frames and dividing by the `fps`, which I would do at initialization (https://github.com/catalystneuro/nwb-conversion-tools/pull/372/files#diff-57b78875c81d904caa745de93f7771de666b821e9d52586cd03a8328e7a74c6bR173) instead of setting them all to `0.0`.

Saksham20: I think its better for us to have starting times as Nans instead of 0s if they are not specified. Its more accurate than assuming that videos are continuous.

Saksham20:
> @Saksham20 Can you point me to the logic in the code responsible for handling `stub_test=True` when `external_mode=False`? The one that shortens the number of frames loaded.
> 
> Also if you can add some tests that ensure such a method works correctly.

I missed adding the logic to change the `frame_count` of the video capture object when stub_test is true. https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/bca08f18ff7df1bc896969d91072501d52258bf6

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,88 +144,126 @@ def run_conversion(
...
-            if check_regular_timestamps(ts=timestamps):
-                fps = get_movie_fps(movie_file=file)
-                image_series_kwargs.update(starting_time=starting_times[j], rate=fps)
-            else:
-                image_series_kwargs.update(timestamps=H5DataIO(timestamps, compression="gzip"))
+        image_series_kwargs_list_updated, file_paths_list = _check_duplicates(image_series_kwargs_list)
+        if starting_times is not None:
+            assert isinstance(starting_times, list) and all(
+                [isinstance(x, float) for x in starting_times]
+            ), "Argument 'starting_times' must be a list of floats."
```
CodyCBakerPhD, line 168:
```suggestion
```

Already being asserted at beginning of function.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,88 +144,126 @@ def run_conversion(
...
+            assert len(starting_times) == len(image_series_kwargs_list_updated), (
+                "starting times list length must be equal number of unique ImageSeries " "containers to write to nwb"
+            )
+        else:
+            warn("starting_times not provided, setting to 'None'")
+            starting_times = [None] * len(image_series_kwargs_list_updated)
```
CodyCBakerPhD, line 174:
```suggestion
            starting_times = [np.nan] * len(image_series_kwargs_list_updated)
```

What you said earlier makes more sense; `NaN` would be better

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,88 +144,126 @@ def run_conversion(
...
-                total_frames = len(timestamps)
-                frame_shape = get_frame_shape(movie_file=file)
-                maxshape = [total_frames]
-                maxshape.extend(frame_shape)
+                with VideoCaptureContext(str(file)) as video_capture_ob:
+                    if stub_test:
+                        video_capture_ob.frame_count = 10
+                    total_frames = video_capture_ob.get_movie_frame_count()
+                    frame_shape = video_capture_ob.get_frame_shape()
+                    start_time = starting_times[j] if starting_times[j] is not None else 0.0
```
CodyCBakerPhD, line 199:
```suggestion
```
With `NaN` as default, this is unneeded.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD, line 174:
```suggestion
            warn("starting_times not provided, setting to 'NaN'")
            starting_times = [np.nan] * len(image_series_kwargs_list_updated)
```

What you said earlier makes more sense; NaN would be better

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -152,88 +144,126 @@ def run_conversion(
...
+                image_series_kwargs.update(data=data)
+                if check_regular_timestamps(ts=timestamps):
+                    image_series_kwargs.update(starting_time=starting_times[j], rate=fps)
+                else:
+                    image_series_kwargs.update(timestamps=timestamps)
+                    if starting_times[j] is None:
+                        current_description = image_series_kwargs.get("description", "")
+                        image_series_kwargs.update(
+                            description=current_description + "Start time not provided. Timestamps may not be accurate."
+                        )
```
CodyCBakerPhD, line 265:
```suggestion
                    if np.isnan(starting_times[j]):
                        warn("Start time not provided. Timestamps may not be accurate.")
```

TBH people rarely read the descriptions; I would just issue a warning to the person running the conversion to (in addition with the warning above) inform them that timing information might not be accurate.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: it would make all timestamps as Nans also, I dont think that would be a good idea. 
If starting time is not specified, its better to still have timestamps but with an updated description of the `ImageSeries` like: https://github.com/catalystneuro/nwb-conversion-tools/blob/212fb41916d849f1a1594a3c7dcce0133b2641ea/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py#L265

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: a warning is important for the user doing the conversion. 
But description is important to communicate this to a user in the far future who might access the file. Agreed, Its more likely to be overlooked but its good to still have it in there.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: with Nans in there it might fail any downstream analysis/plots one might do by recreating the timestamps etc. And this might be hard to debug. 
So keeping it as none (which the `ImageSeries` would default to 0.0 internally) and updating the description is safer IMO.

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD: In general, I don't think we should automatically adjust object descriptions if specified by the user. We can adjust it as you propose in the case where the default description is being generated, but it's weird that a user would have to specify a data-related argument to prevent text-based metadata from being appended to a manually set one.

Example: Let's say the video spans the entire session duration, and starts exactly at relative time `0.0`. They took the time to write a nice detailed description for the ImageSeries in the metadata, but it didn't occur to the user to have to hard-specify `starting_times=[0.0]` since it seemed obvious to them in this case. They didn't check the output NWBFile super carefully, and missed that we automatically added extra information onto the description. Now everyone that uses the file afterward and does take the time to read the new description in full will question the `timestamps`.

Saksham20: then should we just make starting_times as a required argument?

CodyCBakerPhD: > then should we just make starting_times as a required argument? When there are >1 video files to be written?

I thought about that but they could be simultaneous captures of the same experiment (say, from different angles) so that's not fair either. And `starting_time` isn't required on any other interfaces so it would be weird to do that here.

I would just not adjust the description unless we're filling it with a default one.

CodyCBakerPhD:
Related - an old issue from the first version of this interface: https://github.com/NeurodataWithoutBorders/pynwb/issues/1318

Can you look into it and see if it can be closed? Or if it needs more action propagated upward?

Can we set multiple `starting_times` when `external_mode=True` and more than one file path is being specified?

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
Saksham20: a bit like this? https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/0125ee9a5c2bbbd912a9e739e676b2a6a5632f7b

Saksham20: IMO, if its a rare case of simultaneous capture, then we should enforce the user to specify the `starting_times`. I think its important info for dataset in the `ImageSeries`.

Saksham20: I think its better to be strict about it rather than generate a ton of nwbfiles that dont have this important info no?

Saksham20:
> Related - an old issue from the first version of this interface: https://github.com/NeurodataWithoutBorders/pynwb/issues/1318
> Can you look into it and see if it can be closed? Or if it needs more action propagated upward?

> Can we set multiple starting_times when external_mode=True and more than one file path is being specified?

Yeah thanks for bringing that up. I have though of that too, I've recreated the issue with a simpler code snippet https://github.com/NeurodataWithoutBorders/pynwb/issues/1433

nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD: Sure

CodyCBakerPhD:
@Saksham20 Just need a test that explicitly verifies the size using `stub_test=False` is exactly 10 frames (for a video > 10 frames).

Then should be good to go.

Saksham20:
> @Saksham20 Just need a test that explicitly verifies the data size using stub_test=False and external_mode=False is exactly 10 frames (for a video > 10 frames).

Oh yeah, missed that. Done https://github.com/catalystneuro/nwb-conversion-tools/pull/372/commits/bd67a7bd5f27582562aea2d43c60e025b3cadef0

PR #376, opened 2022-01-19
return nwbfile

bendichter:
returning the nwbfile object in write_recording and write_sorting allows us to build the nwbfile in memory

### DIFF
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -1026,6 +1026,7 @@ def write_recording(
             iterator_opts=iterator_opts,
             write_electrical_series=write_electrical_series,
         )
+    return nwbfile
 
 
 def get_nspikes(units_table: pynwb.misc.Units, unit_id: int):
@@ -1398,3 +1399,4 @@ def write_sorting(
             units_name=units_name,
             units_description=units_description,
         )
+    return nwbfile


CodyCBakerPhD:
@bendichter We currently operate on the `nwbfile` in memory as a pointer (https://github.com/catalystneuro/nwb-conversion-tools/blob/main/nwb_conversion_tools/nwbconverter.py#L171-L172): this is how the `NWBConverter` class collects data across multiple interfaces prior to a single write operation. If you really do want a return structure for some other reason, we should mirror the optional argument behavior with the original behavior as default to avoid breaking back-compatibility (https://github.com/catalystneuro/nwb-conversion-tools/blob/main/nwb_conversion_tools/nwbconverter.py#L116 -> https://github.com/catalystneuro/nwb-conversion-tools/blob/main/nwb_conversion_tools/nwbconverter.py#L155 -> https://github.com/catalystneuro/nwb-conversion-tools/blob/main/nwb_conversion_tools/nwbconverter.py#L181)

bendichter:
I don't understand, what's wrong with returning `nwbfile`?

CodyCBakerPhD:
I see; you are relying on the first call to a  `write_*` method to instantiate your `nwbfile` object, so something like

```python
metadata = dict(NWBFile=dict(session_start_time=datetime(...), ...), ...)  # required to specify the correct start time
nwbfile = write_recording(recording=my_ap_recording, metadata=metadata)
write_recording(recording=my_lfp_recording, nwbfile=nwbfile, metadata=metadata)
```

in which case this change should be fine. It shouldn't break the current in-place usage since the returned output value would simply not be assigned to a value in the functions namespace.

CodyCBakerPhD (dismissed):

CodyCBakerPhD:
Actually it occurs to be looking over those functions that a couple things will need to be changed for this to work as expected; I'll open them in a separate PR pointing to this and test it out

CodyCBakerPhD (approved):

PR #377, opened 2022-01-19
Update docs and arguments for NWBFile object

CodyCBakerPhD:
## Motivation
Updating docstrings, annotation typing, and required vs. optional behavior for inputs across the helper functions to make the new return behavior clearer.

## How to test the behavior?
It's on the long-term To-Do list to add many tests the these assertions are actually triggered.

## Checklist

### DIFF
--- a/nwb_conversion_tools/utils/spike_interface.py
+++ b/nwb_conversion_tools/utils/spike_interface.py
@@ -12,7 +12,7 @@ from collections import defaultdict
 import pynwb
 from spikeinterface import BaseRecording
 from spikeinterface.core.old_api_utils import OldToNewRecording
-from spikeextractors import RecordingExtractor, SortingExtractor, SubRecordingExtractor
+from spikeextractors import RecordingExtractor, SortingExtractor
 from numbers import Real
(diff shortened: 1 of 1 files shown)


nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -95,17 +95,15 @@ def get_nwb_metadata(recording: SpikeInterfaceRecording, metadata: dict = None):
     return metadata
 
 
-def add_devices(recording: SpikeInterfaceRecording, nwbfile=None, metadata: dict = None):
+def add_devices(nwbfile: pynwb.NWBFile, metadata: dict = None):
```
bendichter, line 98: so add_devices doesn't use the recording object?

CodyCBakerPhD: Nope lol, we've known that for a long time but kept the `recording` argument for posterity. Maybe the new ProbeInterface might have some kind of API value related to devices, but neither spikeextractors nor spikeinterface have explicit usage for that, so this is all dependent on `metadata` and default values

bendichter (approved):

PR #379, opened 2022-01-20
update spikeglx to use new spikeinterface api

bendichter:


CodyCBakerPhD:
In general, I think it would be worth it as we update the `DataInterfaces` to the new SI versions, to allow users to be able to control which version of `Recording` object is used, similar to how we have conversion options controlling which version of `DataChunkIterator` to use (`v1` for original `DataChunkIterator` and `v2` for `GenericDataChunkIterator`). So something like `recording_version="v1"` for the `RecordingExtractor` and `recording_version="v2"` for the `BaseRecording`, if one exists for that format. We'd only support future debugging for the newer version, of course, but it would be nice to have the old ones to fall back on in any situation where the old one does something the new one doesn't.

CodyCBakerPhD:
Other than that, some questions about this version of `SpikeGLX`:

1) does the channel metadata for `lf` files match `ap` files? This line indicates different probe loading behavior: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/neoextractors/spikeglx.py#L40

2) If you have multiple `ap` files in the same folder, how is this handled by `neo` and how should we handle it at the `DataInterface` level? (likely need to add `stream_id` argument)

CodyCBakerPhD:
With the new neo release, this would be a good candidate to test out as our first one to get refactored this way within the next couple of weeks

Saksham20:
> does the channel metadata for lf files match ap files? This line indicates different probe loading behavior: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/neoextractors/spikeglx.py#L40

For this condition, 'location', 'contact_vector' and 'group' are added [here](https://github.com/SpikeInterface/spikeinterface/blob/f724d60e6367bd946814631dd091ce4dccaffae4/spikeinterface/core/baserecording.py#L355). 
If we dont add that property info in `get_metadata()` then `write_electrodes` will pick the names itself as the description. I dont think the user is expected to provide any custom description for this so its okay for `write_electrodes()` to do what it does [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/3eb8461ffc520794399c424188ee2fceeddf6b8c/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L314)

Saksham20:
>If you have multiple ap files in the same folder, how is this handled by neo and how should we handle it at the DataInterface level? (likely need to add stream_id argument)

I think Neo expects that there will be one 'meta' file for each '.ap'/'.lf' file. [here](https://github.com/NeuralEnsemble/python-neo/blob/0cfe62d487d3333754ab5d12e4a59e53a5485707/neo/rawio/spikeglxrawio.py#L232)

Additionally, I have added `stream_id` as optional.

Saksham20:
> I'd be interested to have a record of comparison between the two versions; using a decently largish file, how much time does it take to write to NWB for the RecordingExtractor version and the BaseRecording (neo backend)?

TODO:
- [ ] write comparison for old vs new version

Saksham20:
> So something like recording_version="v1" for the RecordingExtractor and recording_version="v2" for the BaseRecording, if one exists for that format.

One design implementation possibility:

We could move some logic of checking for the `recording_version` into `BaseRecordingExtractorInterface`. Something like:
```python
class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
    """Primary class for all RecordingExtractorInterfaces."""

    RecordingExtractorClass= None

    def __init__(self, version: str = "v1", **source_data):
       super().__init__(**source_data)
       self.version = version
       self.recording_extractor = self._get_recording_class(**source_data)
       self.subset_channels = None

    def _get_recording_class(self, version):
      if self.version=="v1":
         recordingextclass = eval('spikeextractors.' + RecordingExtractorClass)
      else:
         recordingextclass = eval('spikeinterface.extractors.' + RecordingExtractorClass)
      return recordingextclass

   @abstractmethod
   def _validate_version_args(self): # something that each child class can use to validate v1 or v2 args. 
      pass
```
Then downstream spikeglx can define the class variable `RecordingExtractorClass="SpikeGLXRecordingExtractor"`. 
In case of the above implementation, then the [line 59](https://github.com/catalystneuro/nwb-conversion-tools/pull/379/files#diff-bd6a6b9bab0b4c4748806940c25adf25f13489d568b08b25aa81a2a06973e97aR59) function can become `_validate_version_args`:

```python
def _validate_version_args(self, file_path, folder_path, **args):
    if version == "v1":
        assert file_path is not None and Path(file_path).is_file(), f"{file_path} should be a file for version='v1'"
        rx_kwargs = dict(file_path=str(file_path), **kwargs)
        source_data = file_path
    elif version == "v2":
        assert (
            folder_path is not None and Path(folder_path).is_dir()
        ), f"{folder_path} should be a folder for version='v1'"
        rx_kwargs = dict(folder_path=str(folder_path), **kwargs)
        source_data = folder_path
    else:
        raise ValueError("specify version='v1' for spikeextractors " "version='v2' for spikeinterface")
    return rx_kwargs, source_data
```

This is how its done in spike interface [here](https://github.com/SpikeInterface/spikeinterface/blob/d57db1b7aa81d5737e2f6b113b1c021ed539f31a/spikeinterface/extractors/neoextractors/spikeglx.py#L31)

Saksham20:
Some tests are currently failing because:

The new `spikeinterface` api supports `channel_ids` as **non-int** values, which we now support writing to nwb (#415). These non-int channel ids get written into the electrodes table under `channel_name` column [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/28eba22d7d7a0fb3447a27b2830f2eebfb683e69/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L323) by @h-mayorquin.
However, the `NWBRecordingExtractor` of both `spikeinterface` and `spikeextractors` pulls ids from the electrodes table index (SI implementation [here](https://github.com/SpikeInterface/spikeinterface/blob/cf53b64836dfef715fd6c21d0f919109cb155698/spikeinterface/extractors/nwbextractors.py#L126) and SE implementation [here](https://github.com/SpikeInterface/spikeextractors/blob/72ed8111bb914d7b29df9c1c60fb4b669a2aade6/spikeextractors/extractors/nwbextractors/nwbextractors.py#L220) ). Thus the round trip tests checking equality of recordings fail at `RX.get_channel_ids()` [here](https://github.com/SpikeInterface/spikeinterface/blob/36c60a352779b61ffd16f45195dccfcf9f5af31a/spikeinterface/core/testing.py#L10)
Specifically:
```python
# RX_si is new spikeinterace recording containing char channel_ids
chan_ids = RX_si.get_channel_ids()  # (a char array)
# write RX_si to nwb using run_conversion() to 'nwbfile.nwb'
RX_si_round_trip = si.NwbRecordingExtractor(file_path = 'nwbfile.nwb')
channel_ids_rt = RX_si_round_trip.get_channel_ids()  # get only ints

chan_ids != channel_ids_rt 
```

We may either,
@alejoe91 
1. need to update the `spikeinterface.core.testing.check_recordings_equal` to not assert equality of channel_ids `assert np.allclose(RX1.get_channel_ids(), RX2.get_channel_ids())`. 
2. Change `si.NwbRecordingExtractor` to optionally output `char` channel_ids (look for `channel_name` in the electrodes table and return that else return the table ids which are `ints`). This is dependent on our convention of putting char channel_ids under `channel_name` column. 

How do you all think we should approach this problem? @bendichter @CodyCBakerPhD

alejoe91:
@Saksham20 should be good for the new SI in master: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/nwbextractors.py#L126-L129

This implements exactly option 2.

Saksham20:
> @Saksham20 should be good for the new SI in master: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/nwbextractors.py#L126-L129
> 
> This implements exactly option 2.

Oh, thanks, that recent change was not updated in my si master. Now the tests work!

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -1,59 +1,120 @@
...
+    if isinstance(recording, se.SpikeGLXRecordingExtractor):
+        if isinstance(recording, se.SubRecordingExtractor):
+            current_recording = recording._parent_recording
+        else:
+            current_recording = recording
+        n_shanks = int(current_recording._meta.get("snsShankMap", [1, 1])[1])
+        if n_shanks > 1:
+            raise NotImplementedError("SpikeGLX metadata for more than a single shank is not yet supported.")
+        session_start_time = datetime.fromisoformat(current_recording._meta["fileCreateTime"]).astimezone()
+        metadata_update["NWBFile"]["session_start_time"] = (session_start_time.strftime("%Y-%m-%dT%H:%M:%S"),)
```
bendichter, line 55: why is this a tuple?

bendichter: why are we converting from iso format to datetime and then back to iso format?

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -1,59 +1,120 @@
...
+
+    return dict_deep_update(metadata, metadata_update)
+
+
+def _init_recording(version, file_path, folder_path, **kwargs):
+    if version == "v1":
```
bendichter, line 61: why are we keeping v1 around?

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -1,59 +1,120 @@
...
+    if isinstance(recording, se.SpikeGLXRecordingExtractor):
+        if isinstance(recording, se.SubRecordingExtractor):
+            current_recording = recording._parent_recording
+        else:
+            current_recording = recording
+        n_shanks = int(current_recording._meta.get("snsShankMap", [1, 1])[1])
+        if n_shanks > 1:
+            raise NotImplementedError("SpikeGLX metadata for more than a single shank is not yet supported.")
+        session_start_time = datetime.fromisoformat(current_recording._meta["fileCreateTime"]).astimezone()
+        metadata_update["NWBFile"]["session_start_time"] = str(session_start_time)
```
bendichter, line 55: @Saksham20 you did not fix the underlying issue

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
Saksham20: its to get the timezone content. 
```python
dt = datetime.fromisoformat("2020-11-03T10:35:10").astimezone()
datetime.isoformat(dt) # "2020-11-03T10:35:10+5:00"
```

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
Saksham20: > In general, I think it would be worth it as we update the DataInterfaces to the new SI versions, to allow users to be able to control which version of Recording object is used, similar to how we have conversion options controlling which version of DataChunkIterator to use (v1 for original DataChunkIterator and v2 for GenericDataChunkIterator). So something like recording_version="v1" for the RecordingExtractor and recording_version="v2" for the BaseRecording, if one exists for that format. We'd only support future debugging for the newer version, of course, but it would be nice to have the old ones to fall back on in any situation where the old one does something the new one doesn't.

I agree with what @CodyCBakerPhD suggests

Saksham20:
> @Saksham20 should be good for the new SI in master: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/nwbextractors.py#L126-L129

This implements exactly option 2.

Since this is not yet release on pypi, its failing on CI

Saksham20:
now using an approach similar to the neuroscope pr #432 which is now in main branch

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -1,60 +1,115 @@
...
+        if spikeextractors_backend:
+            assert file_path is not None and Path(file_path).is_file(), f"{file_path} should be a file for version='v1'"
+            self.RX = se.SpikeGLXRecordingExtractor
+            super().__init__(file_path=str(file_path), **kwargs)
+            self.recording_extractor = OldToNewRecording(oldapi_recording_extractor=self.recording_extractor)
+            self.folder_path = file_path.parent
+        else:
+            assert (
+                folder_path is not None and Path(folder_path).is_dir()
+            ), f"{folder_path} should be a folder for version='v2'"
```
h-mayorquin, line 103: The string here should reference the backend instead of the version, correct?

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -81,27 +134,37 @@ def get_conversion_options(self):
...
-        """Compile input schema for the RecordingExtractor."""
-        source_schema = get_schema_from_method_signature(class_method=cls.__init__, exclude=["x_pitch", "y_pitch"])
-        source_schema["properties"]["file_path"]["description"] = "Path to SpikeGLX file."
-        return source_schema
+        return _get_source_schema(cls)
+
+    def __init__(
+        self,
+        folder_path: FilePathType = None,
+        file_path: FilePathType = None,
```
h-mayorquin, line 144: Should not we just keep the `file_path` only for backwards compatibility and extract the folder/directory path from it for so we don't add extra arguments to the function?

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
Saksham20: yes, thanks, have changed that

nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
Saksham20: Since we want to encourage using spike interface, I think using supporting arguments (si uses `folder_path` with `stream_id` to specify the specific file) is better. 
IMOI, If we still want to give users the option to specify the backend, then the recognised arg (`file_path`) should also be present.

h-mayorquin: My argument again it is redundancy. The argument cuts both ways though. We could get `file_path` from `folder_path` or vice-versa. Less sure about the `stream_id`.

In a more general note, in principle "using `spikeinterface` backend" is different from "make our functions call as similar to their recording objects calls". My feeling is that we are concentrating in the former but I am less sure about the latter, maybe am I wrong and we actually want to have similar function calls?

Saksham20: yes, we can always get `file_path` given `folder_path` and `stream_id`:
File for spikeglx are of the form `somepath/name_g0_t0.imec0.lf.bin` with `folder_path = "somepath"` and `stream_id="imec0.lf"`. So we can get `file_name` like so:
```python
from pathlib import Path
folder_path = Path(somepath/name_g0_t0.imec0.lf.bin)
stream_id = "imec0.lf"
file_path = list(folder_path.glob(f"*{stream_id}*"))[0]
```
or just specify the `file_path` and get the folder and stream_id like so:
```python
from pathlib import Path
file_path = Path("somepath/name_g0_t0.imec0.lf.bin")
folder_path = str(file_path.parent)
stream_id = "".join(file_path.suffixes[:-1])[1:]  # "imec0.lf"
```
If we are making the design choice to not mimic the args of the spikeinterface/spikeextractors backend, then, IMO the second one where we allow only the `file_path` (like you also mentioned) is better and less confusing for the user. 

I noticed something else: we are not propagating `stream_id` argument in `write_recording` [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/f2ceb69e744ef800d907ebb09668e09037b1ee24/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L848) which is used by `add_electrical_series` [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/f2ceb69e744ef800d907ebb09668e09037b1ee24/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L454) then why not remove it?

h-mayorquin: Yes, it would be good to get @bendichter and @CodyCBakerPhD thoughts and whether we are should try to emulate the calling signature of `spikeinterface` to some degree. 

Yes, plus is only a single argument which makes it more simple. I feel that many of our functions calls are already to complicated. 

```python
stream_id = "".join(file_path.suffixes[:-1])[1:]  # "imec0.lf"
```
Nice use of suffixes : )

Saksham20: > whether we are should try to emulate the calling signature of spikeinterface to some degree.
as discussed with @bendichter lets keep the old arguments: `file_path` instead of `folder_path` + `stream_id`

PR #383, opened 2022-01-25
Modify neuroscope to use spike interface

### LINKED ISSUE #382: Gain mismatch with the neuroscope and spikeinterface
## Description
This is to discuss an error that arises when a neuroscope recorder subsets some of its channels. In short, there is a shape mismastch between the channels gains inherited from the neuroscope recorder and the ones that are assigned to `spikeinterface`.  See the bottom of this issue for the error trace. In this case, the gain is propagated correctly but no the shape.

In more detail:
The problem occurs in the conversion of the recording extractors to a spike interface object in the `spike-interface` of `nwb-conversion-tools`. This is carred ou int `old_api_utils`:

https://github.com/SpikeInterface/spikeinterface/blob/ec7adf991835e851926c89adb3be7f1d97f8d629/spikeinterface/core/old_api_utils.py#L171-L198

The machinery in `old_ap_utils` correctly gets all the gains from the uppermost level of the original recording extractor (lines 189) but the new spikeinterface object is initiated with the number of channels in the recording extractor that was passed to it (see the arguments passed to the class in lines 172). This generates a size clash in in the new functions in charge of settings the properties as the shape of the two arrays don't match (lines 193).

Now, in this case for Neuroscope the channels are subset because we have some functions that remove the channels not related to neural data  in the `neuroscopedatainterface`:

https://github.com/catalystneuro/nwb-conversion-tools/blob/1bbefafa10e6b2d7b5daa67fc5d6f273bf13a487/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py#L23-L33

So, some channels are removed by this function. Then, in the transformation from spikeextractor to spikeinterface the gains are extracted from the original channel but the new spikeinterface object inherits the number of channels from the subrecording creating a mismatch. 

What gives and what should be done? the gain values should indeed be extracted from the upermost level as it is the one that in fact holds the correct gain values. On the other hand, I think the shape should be broadcasted to be equal to that one in the subrecording as that is the one that will actually be written.

What do you think?
@CodyCBakerPhD @alejoe91 

Error trace:
```python
~/development/buzsaki-lab-to-nwb/buzsaki_lab_to_nwb/tingley_code_septal/convert_tingley_septal.py in <module>
    126         metadata=metadata,
    127         conversion_options=conversion_options,
--> 128         overwrite=True,
    129     )
    130     print("Done with conversion")

~/development/nwb-conversion-tools/nwb_conversion_tools/nwbconverter.py in run_conversion(self, metadata, save_to_file, nwbfile_path, overwrite, nwbfile, conversion_options)
    170 
    171                 for interface_name, data_interface in self.data_interface_objects.items():
--> 172                     data_interface.run_conversion(nwbfile, metadata, **conversion_options.get(interface_name, dict()))
    173 
    174                 io.write(nwbfile)

~/development/nwb-conversion-tools/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py in run_conversion(self, nwbfile, metadata, stub_test, starting_time, use_times, save_path, overwrite, write_as, write_electrical_series, es_key, compression, compression_opts, iterator_type, iterator_opts)
    179             compression_opts=compression_opts,
    180             iterator_type=iterator_type,
--> 181             iterator_opts=iterator_opts,
    182         )

~/development/nwb-conversion-tools/nwb_conversion_tools/utils/spike_interface.py in write_recording(recording, save_path, overwrite, nwbfile, starting_time, use_times, metadata, write_as, es_key, write_electrical_series, write_scaled, compression, compression_opts, iterator_type, iterator_opts)
   1013             iterator_type=iterator_type,
   1014             iterator_opts=iterator_opts,
-> 1015             write_electrical_series=write_electrical_series,
   1016         )
   1017     return nwbfile

~/development/nwb-conversion-tools/nwb_conversion_tools/utils/spike_interface.py in add_all_to_nwbfile(recording, nwbfile, starting_time, use_times, metadata, write_as, es_key, write_electrical_series, write_scaled, compression, compression_opts, iterator_type, iterator_opts)
    824         assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
    825 
--> 826     add_electrodes_info(recording=recording, nwbfile=nwbfile, metadata=metadata)
    827 
    828     if write_electrical_series:

~/development/nwb-conversion-tools/nwb_conversion_tools/utils/spike_interface.py in add_electrodes_info(recording, nwbfile, metadata)
    751     """
    752     add_devices(nwbfile=nwbfile, metadata=metadata)
--> 753     add_electrode_groups(recording=recording, nwbfile=nwbfile, metadata=metadata)
    754     add_electrodes(recording=recording, nwbfile=nwbfile, metadata=metadata)
    755 

~/development/nwb-conversion-tools/nwb_conversion_tools/utils/spike_interface.py in add_electrode_groups(recording, nwbfile, metadata)
    169     assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
    170     if isinstance(recording, RecordingExtractor):
--> 171         checked_recording = OldToNewRecording(oldapi_recording_extractor=recording)
    172     else:
    173         checked_recording = recording

~/miniconda3/envs/buzsaki-lab-to-nwb-env/lib/python3.7/site-packages/spikeinterface/core/old_api_utils.py in __init__(self, oldapi_recording_extractor)
    191         if gains is not None:
    192             if np.any(gains != 1):
--> 193                 self.set_channel_gains(gains)
    194         if offsets is not None:
    195             if np.any(offsets != 0):

~/miniconda3/envs/buzsaki-lab-to-nwb-env/lib/python3.7/site-packages/spikeinterface/core/baserecording.py in set_channel_gains(self, gains, channel_ids)
    493         if np.isscalar(gains):
    494             gains = [gains] * self.get_num_channels()
--> 495         self.set_property('gain_to_uV', gains, ids=channel_ids)
    496 
    497     def get_channel_gains(self, channel_ids=None):

~/miniconda3/envs/buzsaki-lab-to-nwb-env/lib/python3.7/site-packages/spikeinterface/core/base.py in set_property(self, key, values, ids)
    150         values = np.asarray(values)
    151         if ids is None:
--> 152             assert values.shape[0] == size
    153             self._properties[key] = values
    154         else:

AssertionError: 
```

## Steps to Reproduce
TBD

## Environment
Running the latest versions of `nwb-conversion-tools` and `spikeinterface`

## Checklist

- [x] Have you ensured the feature or change was not already [reported](https://github.com/catalystneuro/nwb-conversion-tools/issues)?
- [x] Have you included a brief and descriptive title?
- [x] Have you included a clear description of the problem you are trying to solve?
- [ ] Have you included a minimal code snippet that reproduces the issue you are encountering?
- [x] Have you checked our [Contributing](https://nwb-conversion-tools.readthedocs.io/en/master/developer_guide.html) document?

h-mayorquin:
## Motivation
This should fix #382

As discussed in the issue there is a shape mismatch between gains when the spikeinterface transforms a spikeextractor object to a spikeinterface one. The mismatch occurs when some of the channels are subset and the example in the issue was for neuroscope.

I was discussed with @bendichter yesterday and he mentioned that it would be a good idea to just change fully to `spikeinterface` for neuroscope as we intend to do it anyway at some point. This pull request is the first attempt in this direction and I start by implementing @alejoe91 suggestion in the issue to showcase some problems, discuss and get your feedback.

There are three modifications in the diff:
1) As per @alejoe91  suggestion the `subset_shank_channels` method in the neuroscope interface has been modified to transform the recorder and use the `channel_slice` method instead of subrecording.

2) To keep the changes minimal, the function `add_recording_extractor_properties` in `neuroscopedatainterface` is now called before the sub-recording as the `channel_slice` machinery includes functionality to copy the properties. Though, we change the SX attribute to the corresponding. spikeinterface we will need to modified the `add_recording_extractor_properties` consequently. 

3) As a consequence the method that creates stubs in `baserecordinginterface` has been modified to account for possible `si.BaseRecording` objects as input.

I converted some files with this and the error indeed disappears. I am going to move ahead and change the `SX` attribute to a new itnterface object to see if this still work. Any advice on pits to look for in this direciton @CodyCBakerPhD ?

## Problems
1) The data-type of the nwb files for both the raw data and the lfp is still a float after using this. I was under the wrong impression that the latest version of `nwb-conversion-tools` would avoid this.

2) For some reason there is one test failing and I could not pin it down: `test_converter_recording_extractor_to_nwb` is failing. The problem is that for neuroscope the `NWBRecordingExtractor.get_traces()` method seems to return the data transposed as in time is not the first dimension. This seems strange to me as the files that I produced do have the right dimension: `(stub_time, n_channels)`

## Checklist

### DIFF
--- a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
@@ -95,7 +95,7 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
         if isinstance(self.recording_extractor, se.RecordingExtractor):
             recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
         elif isinstance(self.recording_extractor, si.BaseRecording):
-            recording_extractor = si.FrameSliceRecording(self.recording_extractor, end_frame=end_frame)
+            recording_extractor = self.recording_extractor.frame_slice(start_frame=0, end_frame=end_frame)
         else:
             raise f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording"
 
--- a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
@@ -97,7 +97,7 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
         elif isinstance(self.recording_extractor, si.BaseRecording):
             recording_extractor = self.recording_extractor.frame_slice(start_frame=0, end_frame=end_frame)
         else:
-            raise f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording"
+            raise TypeError(f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording")
 
         return recording_extractor
 
--- a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
@@ -99,8 +99,6 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
         else:
             raise f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording"
 
-        print("Passing by here")
-
         return recording_extractor
 
     def run_conversion(
--- a/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
@@ -4,6 +4,8 @@ from typing import Optional
 import numpy as np
 
 import spikeextractors as se
+import spikeinterface as si
+
 from pynwb import NWBFile
 from pynwb.device import Device
 from pynwb.ecephys import ElectrodeGroup
@@ -90,7 +92,15 @@ class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
         if self.subset_channels is not None:
             kwargs.update(channel_ids=self.subset_channels)
 
-        recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+        if isinstance(self.recording_extractor, se.RecordingExtractor):
+            recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+        elif isinstance(self.recording_extractor, si.BaseRecording):
+            recording_extractor = si.FrameSliceRecording(self.recording_extractor, end_frame=end_frame)
+        else:
+            raise f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording"
+
+        print("Passing by here")
+
         return recording_extractor
 
     def run_conversion(
--- a/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Optional, Union
 
 import spikeextractors as se
+import spikeinterface as si
 from spikeinterface.core.old_api_utils import OldToNewRecording
 
 from pynwb.ecephys import ElectricalSeries
--- a/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
@@ -31,6 +31,7 @@ def subset_shank_channels(
     else:
         recording = recording_extractor
     shank_channels = get_shank_channels(xml_file_path=xml_file_path)
+    
     if shank_channels is not None:
         channel_ids = [channel_id for group in shank_channels for channel_id in group]
         sub_recording = recording.channel_slice(channel_ids)
--- a/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
+++ b/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
@@ -1,8 +1,10 @@
 """Authors: Cody Baker and Ben Dichter."""
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union
 
 import spikeextractors as se
+import spikeinterface as si
+
 from pynwb.ecephys import ElectricalSeries
 
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
@@ -20,16 +22,20 @@ except ImportError:
 INSTALL_MESSAGE = "Please install lxml to use this interface!"
 
 
-def subset_shank_channels(recording_extractor: se.RecordingExtractor, xml_file_path: str) -> se.SubRecordingExtractor:
+def subset_shank_channels(
+    recording_extractor: Union[se.RecordingExtractor, si.BaseRecording], xml_file_path: str
+) -> si.BaseRecording:
     """Attempt to create a SubRecordingExtractor containing only channels related to neural data."""
+    if isinstance(recording_extractor, se.RecordingExtractor):
+        recording = si.core.old_api_utils.OldToNewRecording(recording_extractor)
+    else:
+        recording = recording_extractor
     shank_channels = get_shank_channels(xml_file_path=xml_file_path)
     if shank_channels is not None:
-        sub_recording = se.SubRecordingExtractor(
-            parent_recording=recording_extractor,
-            channel_ids=[channel_id for group in shank_channels for channel_id in group],
-        )
+        channel_ids = [channel_id for group in shank_channels for channel_id in group]
+        sub_recording = recording.channel_slice(channel_ids)
     else:
-        sub_recording = recording_extractor
+        sub_recording = recording
     return sub_recording
 
 
@@ -43,7 +49,8 @@ def add_recording_extractor_properties(recording_extractor: se.RecordingExtracto
     group_electrode_numbers = [x for channels in channel_groups for x, _ in enumerate(channels)]
     group_nums = [n + 1 for n, channels in enumerate(channel_groups) for _ in channels]
     group_names = [f"Group{n}" for n in group_nums]
-    for channel_id in recording_extractor.get_channel_ids():
+
+    for channel_id in channel_map.keys():
         recording_extractor.set_channel_groups(channel_ids=[channel_id], groups=group_nums[channel_map[channel_id]])
         recording_extractor.set_channel_property(
             channel_id=channel_id, property_name="group_name", value=group_names[channel_map[channel_id]]
@@ -105,10 +112,12 @@ class NeuroscopeRecordingInterface(BaseRecordingExtractorInterface):
         if xml_file_path is None:
             xml_file_path = get_xml_file_path(data_file_path=file_path)
         super().__init__(file_path=file_path, gain=gain, xml_file_path=xml_file_path)
+
+        # Add the properties
+        add_recording_extractor_properties(recording_extractor=self.recording_extractor, xml_file_path=xml_file_path)
         self.recording_extractor = subset_shank_channels(
             recording_extractor=self.recording_extractor, xml_file_path=xml_file_path
         )
-        add_recording_extractor_properties(recording_extractor=self.recording_extractor, xml_file_path=xml_file_path)
 
     def get_metadata_schema(self):
         metadata_schema = super().get_metadata_schema()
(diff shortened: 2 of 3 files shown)


nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -90,7 +92,13 @@ def subset_recording(self, stub_test: bool = False):
...
             kwargs.update(channel_ids=self.subset_channels)
 
-        recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+        if isinstance(self.recording_extractor, se.RecordingExtractor):
+            recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+        elif isinstance(self.recording_extractor, si.BaseRecording):
+            recording_extractor = self.recording_extractor.frame_slice(start_frame=0, end_frame=end_frame)
+        else:
+            raise f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording"
+
```
h-mayorquin, line 101: has been modified to account for possible si.BaseRecording objects as input.

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -20,16 +22,21 @@
...
-        sub_recording = se.SubRecordingExtractor(
-            parent_recording=recording_extractor,
-            channel_ids=[channel_id for group in shank_channels for channel_id in group],
-        )
+        channel_ids = [channel_id for group in shank_channels for channel_id in group]
+        sub_recording = recording.channel_slice(channel_ids)
     else:
-        sub_recording = recording_extractor
+        sub_recording = recording
     return sub_recording
```
h-mayorquin, line 41: As per @alejoe91 suggestion the subset_shank_channels method in the neuroscope interface has been modified to transform the recorder and use the channel_slice method instead of subrecording.

h-mayorquin:
So in order to test this with `spikeinterface` I added some machinery to the gin tests. The test for return_scaled=True seems to be still failing. I will investigate this now.

h-mayorquin:
I think this pull request should be merged as it is. 
Right now:
* The pull request solved the problem that it was intended to solve. The shape mismatch between gains at different levels of stacks.
* The Neuroscope corresponding test was modified to account for having `spikeinterface` objects as recorders paving the way for the future change. 

Moreover, I think that the`NeuroScopeRecordingExtractor` of `spikeinterface` has some sort of bug (which goes all the way to `neo`) when not all channels have an assigned group in the `xml`:

https://github.com/SpikeInterface/spikeinterface/issues/486

I found this while testing it against some of the data from the last Buzsaki conversion. 

So, my suggestion is to merge this as it stands as it solves the problem (and to keep pull requests smaller) and that I start to work in the second pull request as soon as we clarify the issue with `spikeinterface`.

What do you think @bendichter @CodyCBakerPhD

nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -90,7 +92,13 @@ def subset_recording(self, stub_test: bool = False):
         if self.subset_channels is not None:
             kwargs.update(channel_ids=self.subset_channels)
 
-        recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+        if isinstance(self.recording_extractor, se.RecordingExtractor):
+            recording_extractor = se.SubRecordingExtractor(self.recording_extractor, **kwargs)
+        elif isinstance(self.recording_extractor, si.BaseRecording):
+            recording_extractor = self.recording_extractor.frame_slice(start_frame=0, end_frame=end_frame)
+        else:
+            raise f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording"
```
CodyCBakerPhD, line 100:
```suggestion
            raise ValueError(f"{self.recording_extractor} should be either se.RecordingExtractor or si.BaseRecording")
```
When raising an error, there are one of two approaches you can take:

(a) `assert "Some message."

which returns an `AssertionError` type, or

(b) `raise SomeError("Some message.")

which returns whatever error type you specify from the following list: https://docs.python.org/3/tutorial/errors.html

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -1,8 +1,10 @@
 """Authors: Cody Baker and Ben Dichter."""
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union
 
 import spikeextractors as se
+import spikeinterface as si
```
CodyCBakerPhD, line 6:
```suggestion
from spikeinterface.core.old_api_utils import OldToNewRecording
```
One trick that can be done to return horizontal line usage across primary code blocks below imports is to use relative imports whenever there aren't so many that they don't span multiple lines in `black` (so like, 4 specific function imports).

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -20,16 +22,21 @@
 INSTALL_MESSAGE = "Please install lxml to use this interface!"
 
 
-def subset_shank_channels(recording_extractor: se.RecordingExtractor, xml_file_path: str) -> se.SubRecordingExtractor:
+def subset_shank_channels(
+    recording_extractor: Union[se.RecordingExtractor, si.BaseRecording], xml_file_path: str
+) -> si.BaseRecording:
     """Attempt to create a SubRecordingExtractor containing only channels related to neural data."""
+    if isinstance(recording_extractor, se.RecordingExtractor):
+        recording = si.core.old_api_utils.OldToNewRecording(recording_extractor)
```
CodyCBakerPhD, line 30:
```suggestion
        recording = OldToNewRecording(oldapi_recording_extractor=recording_extractor)
```
Though it may seem silly for a function that only has a single input argument, using keywords is never a bad idea to prevent future errors in the off-chance that a future or past version of that function has a different argument added to the first position; we've been caught by this before.

Certain built-in functions forbid keywords entirely, like `some_dict.get()`, but most functions you find from our own ecosystems will support it.

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -175,20 +179,33 @@ class TestConverter(NWBConverter):
...
-        check_recordings_equal(RX1=recording, RX2=nwb_recording, check_times=False, return_scaled=False)
-        check_recordings_equal(RX1=recording, RX2=nwb_recording, check_times=False, return_scaled=True)
-        # Technically, check_recordings_equal only tests a snippet of data. Above tests are for metadata mostly.
-        # For GIN test data, sizes should be OK to load all into RAM even on CI
-        npt.assert_array_equal(
-            x=recording.get_traces(return_scaled=False), y=nwb_recording.get_traces(return_scaled=False)
-        )
+
+        if isinstance(recording, si.BaseRecording):
+            nwb_recording = si.core.old_api_utils.OldToNewRecording(nwb_recording)
```
CodyCBakerPhD, line 192:
```suggestion
            nwb_recording = OldToNewRecording(oldapi_recording_extractor=nwb_recording)
```

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -175,20 +179,33 @@ class TestConverter(NWBConverter):
...
+            # For GIN test data, sizes should be OK to load all into RAM even on CI
+            npt.assert_array_equal(
+                x=recording.get_traces(return_scaled=False), y=nwb_recording.get_traces(return_scaled=False)
+            )
+
+        else:
+            check_recordings_equal_si(RX1=recording, RX2=nwb_recording, return_scaled=False)
+            # This can only be tested if both gain and offest are present
+            if recording.has_scaled_traces() and nwb_recording.has_scaled_traces():
+                check_recordings_equal_si(RX1=recording, RX2=nwb_recording, return_scaled=True)
```
CodyCBakerPhD, line 208: Very nice!

CodyCBakerPhD:
@h-mayorquin Actual code looks fine to me, just minor format suggestions.

nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
h-mayorquin: Right, this is way more specific. Reading from the list I think this should be a type error though.

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
h-mayorquin: This makes sense to me.

nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py
h-mayorquin: I added this in a different commit.

h-mayorquin:
@CodyCBakerPhD Great. What do you think of merging as it is so I can start a second pull request? 
I already have most of the other modifications for full usage of neurosocope but I still feel that it will be too large of a PR and not specific enough (is addressing both the error in the issue and the change).

CodyCBakerPhD:
> @CodyCBakerPhD Great. What do you think of merging as it is so I can start a second pull request?

The GitHub repo is configured to not allow PR merges whenever tests are failing; that said, while you're working on fixing those, you can start on a second PR that branches from this one if you point its merge destination to this one.

>I already have most of the other modifications for full usage of neurosocope but I still feel that it will be too large of a PR and not specific enough (is addressing both the error in the issue and the change).

Great instincts! I always prefer to keep PRs as small and modular as possible in their changes

CodyCBakerPhD:
> That said I expect PR https://github.com/catalystneuro/nwb-conversion-tools/pull/391 ought to fix the test issues; if so I'll approve and merge

@h-mayorquin Never mind, while that did fix some installation problems it does still look as if there's some import issues remaining

h-mayorquin:
@CodyCBakerPhD 
Yes, it was me removing the `import spikeinterface as si` from your code suggestions without being mindful that it was used somewhere else. Now it should be working.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin Great, thank you!

PR #384, opened 2022-01-25
Add tests for Neuroscope dtype

CodyCBakerPhD:
## Motivation

Recently discovered that when gains are set to an old RecordingExtractor, the newly refactored code for writing ElectricalSeries sometimes upcasts the data type to `float` due to a bug in the `OldToNewRecording` functions on spikeinterface. This PR adds explicit tests for this behavior and will point to a fix on SI to ensure it works.

Also attached some style suggestions for previous test written by @h-mayorquin 

## Checklist

### DIFF
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -10,6 +10,6 @@ jsonschema>=3.2.0
 psutil>=5.8.0
 lxml>=4.6.5
 spikeextractors>=0.9.7
-spikeinterface @ git+https://github.com/catalystneuro/spikeinterface.git@Fix-old-to-new-recording-dtype
+spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@master
 neo>=0.9.0
 roiextractors>=0.3.1
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -10,6 +10,6 @@ jsonschema>=3.2.0
 psutil>=5.8.0
 lxml>=4.6.5
 spikeextractors>=0.9.7
-spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git
+spikeinterface @ git+https://github.com/catalystneuro/spikeinterface.git@Fix-old-to-new-recording-dtype
 neo>=0.9.0
 roiextractors>=0.3.1
--- a/tests/test_on_data/test_gin_ecephys.py
+++ b/tests/test_on_data/test_gin_ecephys.py
@@ -276,7 +276,7 @@ class TestEcephysNwbConversions(unittest.TestCase):
         converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, conversion_options=conversion_options)
 
         with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
-            nwbfile_in = io.read()
+            nwbfile = io.read()
             output_dtype = nwbfile_in.acquisition["ElectricalSeries_raw"].data.dtype
             self.assertEqual(first=output_dtype, second=np.dtype("int16"))
 
--- a/tests/test_on_data/test_gin_ecephys.py
+++ b/tests/test_on_data/test_gin_ecephys.py
@@ -277,7 +277,7 @@ class TestEcephysNwbConversions(unittest.TestCase):
 
         with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
             nwbfile = io.read()
-            output_dtype = nwbfile_in.acquisition["ElectricalSeries_raw"].data.dtype
+            output_dtype = nwbfile.acquisition["ElectricalSeries_raw"].data.dtype
             self.assertEqual(first=output_dtype, second=np.dtype("int16"))
 
     def test_neuroscope_starting_time(self):
--- a/tests/test_on_data/test_gin_ecephys.py
+++ b/tests/test_on_data/test_gin_ecephys.py
@@ -235,34 +235,50 @@ class TestEcephysNwbConversions(unittest.TestCase):
     )
     def test_neuroscope_gains(self, name, conversion_options):
         input_gain = 2.0
-        data_interface = NeuroscopeRecordingInterface
         interface_kwargs = dict(file_path=str(DATA_PATH / "neuroscope" / "test1" / "test1.dat"), gain=input_gain)
 
-        nwbfile_path = str(self.savedir / f"{data_interface.__name__}-{name}.nwb")
+        nwbfile_path = str(self.savedir / f"test_neuroscope_gains_{name}.nwb")
 
         class TestConverter(NWBConverter):
-            data_interface_classes = dict(TestRecording=data_interface)
+            data_interface_classes = dict(TestRecording=NeuroscopeRecordingInterface)
 
         converter = TestConverter(source_data=dict(TestRecording=interface_kwargs))
         converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, conversion_options=conversion_options)
 
-        # nwb file check-test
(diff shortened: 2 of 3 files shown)


tests/test_on_data/test_gin_ecephys.py
```diff
@@ -235,34 +235,50 @@ class TestConverter(NWBConverter):
...
-        io = NWBHDF5IO(nwbfile_path, "r")
-        nwbfile_in = io.read()
-        output_conversion = nwbfile_in.acquisition["ElectricalSeries_raw"].conversion
-        output_gain = output_conversion * 1e6
-        assert input_gain == pytest.approx(output_gain)
+        with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
+            nwbfile_in = io.read()
+            output_conversion = nwbfile_in.acquisition["ElectricalSeries_raw"].conversion
+            output_gain = output_conversion * 1e6
+            self.assertEqual(first=input_gain, second=output_gain)
```
CodyCBakerPhD, line 252: @h-mayorquin It's good practice to use context managers whenever possible when dealing with file I/O, to ensure safe opening/closing/access

CodyCBakerPhD, line 252: @h-mayorquin When inside a `unittest.TestCase`, it's best to use the large collection of `self.assert{do_something}` functions that are available. Only some array based ones have to be done through `numpy.testing`, like below

h-mayorquin, line 252:
```suggestion
            self.assertAlmostEqual(first=input_gain, second=output_gain)
```

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -235,34 +235,50 @@ class TestConverter(NWBConverter):
...
+        class TestConverter(NWBConverter):
+            data_interface_classes = dict(TestRecording=NeuroscopeRecordingInterface)
 
-        # A subrecording extractor is a very simple class, input output (you can use the recorder above 'recording in 229')
-        # [The two things that you can do is to subest time [that is frames] or channels so maybe a combination of those uses.
+        converter = TestConverter(source_data=dict(TestRecording=interface_kwargs))
+        converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, conversion_options=conversion_options)
+
+        with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
+            nwbfile_in = io.read()
```
h-mayorquin, line 279:
```suggestion
            nwbfile = io.read()
```

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -235,34 +235,50 @@ class TestConverter(NWBConverter):
...
+            output_dtype = nwbfile_in.acquisition["ElectricalSeries_raw"].data.dtype
```
h-mayorquin, line 280:
```suggestion
            output_dtype = nwbfile.acquisition["ElectricalSeries_raw"].data.dtype
```

h-mayorquin (commented):
I think that both in this and the other test that I wrote we should use `nwbfile` instead of `nwbfile_in` to keep consistency with the previous tests above. I could not remember what I was thinking.

requirements-minimal.txt
```diff
@@ -10,6 +10,6 @@ jsonschema>=3.2.0
 psutil>=5.8.0
 lxml>=4.6.5
 spikeextractors>=0.9.7
-spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git
+spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@master
```
h-mayorquin, line 13: .

h-mayorquin:
Thanks for the advice on how to test better. Looks great to me!

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: Yeah there's many different devOps packages with their own ways of doing things. If you're in a testing file that is using `pytest` functions (no `unittest`) then direct `assert` is the right way to go, or any of the other package-specific methods.

But here, since we're already inside one, we might as well use things like `self.assertEqual(first=object_1, second=object_2)` instead of `assert object_1==object_2`. For simple objects this difference is minimal but for more complicated objects there are a host of other `self.assert{something}` functions specific to handling those cases. Examples: lists, dictionary subsets, error catching, etc.

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -235,34 +235,50 @@ class TestConverter(NWBConverter):
...
         input_gain = 2.0
-        data_interface = NeuroscopeRecordingInterface
         interface_kwargs = dict(file_path=str(DATA_PATH / "neuroscope" / "test1" / "test1.dat"), gain=input_gain)
 
-        nwbfile_path = str(self.savedir / f"{data_interface.__name__}-{name}.nwb")
+        nwbfile_path = str(self.savedir / f"test_neuroscope_gains_{name}.nwb")
 
         class TestConverter(NWBConverter):
-            data_interface_classes = dict(TestRecording=data_interface)
+            data_interface_classes = dict(TestRecording=NeuroscopeRecordingInterface)
```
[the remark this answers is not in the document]
CodyCBakerPhD: Sure, we can go back to doing that when/if we want to extend this test to other interfaces. Will need to change the actual test function name too in that case, but for now just reducing code to what's immediately necessary.

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: Is exact equal failing for some reason? I can't think of a reason they would actually change at the floating point level, and the tests seem to be passing as is

In general, stronger statements are better when you can get away with them. For certain array testing (like the trace calls), we use approximate testing because array equality is calculated on a matrix norm, which can easily incur very minor epsilon's from floating point errors accolated across all the elements.

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: Indicating that the NWBFile is being read in as a part of roundtrip write is fine by me, it's extra contextual information; I leave this change up to whatever you prefer.

requirements-minimal.txt
CodyCBakerPhD: Yeah, this doesn't actually change functionality. Just makes it more explicit (especially to distinguish 'main' (new repo's) vs. 'master' (old repo's).

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: I have this idea that floats should not be compared with stric equality for floating point representation issues.

h-mayorquin: I have this idea that as a general practice float equality should be tested almost equal (epsilon difference) because of floating point issues.

CodyCBakerPhD: For data quality, like trace scaling through SpikeInterface readers, I'd completely agree. But for metadata fields (attributes, column values, etc) I do think exact representation matching ought to be the ideal whenever it's achievable.

Explicit example: in Neuroscope, when writing gains, if the input `gain_to_uV=0.195`, the NWBFile ought to store `nwbfile.acquisition["ElectricalSeries_raw"].conversion` as `0.195E-6` instead of `0.1944444449E-6`, unless there's no way around it. But here it seems to be passing exactness anyhow.

h-mayorquin: All right, that makes sense.

h-mayorquin (approved):
Approving the changes.

PR #385, opened 2022-01-27
removed unnecessary argument in add_device

h-mayorquin:
## Motivation
Quick bug?

Some of the `add_device` functions in the utils for `spikeinterface` contain the recording argument that was removed before (discussed in the closed pull request #377). It seems that some of those were still there.


CodyCBakerPhD:
@h-mayorquin Thanks! Merging on test completion.

CodyCBakerPhD (approved):

PR #392, opened 2022-02-03
videocontext manager with tests

Saksham20:
## Motivation

This is a PR isolating the creation of a Context Manager for video files and its corresponding tests. 
This would then be integrated with MovieDataInterface in a separate pr #372


CodyCBakerPhD:
@Saksham20 For the tests/CI, I recommend using lazy import/skip for the `cv2` dependency

Saksham20:
> @Saksham20 For the tests/CI, I recommend using lazy import/skip for the `cv2` dependency

I prefer to add cv2 installation in the tests' workflows

CodyCBakerPhD:
> Its better to add cv2 installation in the tests' workflows then

No; those are built fundamentally to mimic the environment of fellow developers to avoid issues with the minimal install. Note that the cv2 is already in the full requirements: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/requirements-full.txt#L12

So this is purely a minimal install issue; I'd remove those lines you changed to the workflows, and add the lazy type of import that the existing movie tests have so that these tests are only triggered in the second half of the workflow. I'd recommend utilizing `@pytest.mark.skip(logical_triggger, reason="OpenCV is not installed!")` for this.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+        assert self.frame_count > 0, "movie contains no frames"
+        self.fps = self.get_movie_fps()
+        self.frame = self.get_movie_frame(0)
+        assert self.frame is not None, "unable to read the movie file provided"
+
+    def get_movie_timestamps(self):
+        """
+        Return numpy array of the timestamps for a movie file.
+
+        """
```
CodyCBakerPhD, line 35:
```suggestion
        """Return numpy array of the timestamps for a movie file."""
```
And ditto for the others; `pydocstyle D200`

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+        self.vc = cv2.VideoCapture(*args, **kwargs)
+        self._args = args
+        self._kwargs = kwargs
+        self.stub = stub
+        self._current_frame = 0
+        self.frame_count = self.get_movie_frame_count()
+        assert self.frame_count > 0, "movie contains no frames"
+        self.fps = self.get_movie_fps()
+        self.frame = self.get_movie_frame(0)
+        assert self.frame is not None, "unable to read the movie file provided"
```
CodyCBakerPhD, line 29:
```suggestion
    def __init__(self, file_path: FilePathType, **kwargs):
        self.file_path = str(file_path)

        self.fps = self.get_movie_fps()
        self.frame_count = self.get_movie_frame_count()
        assert self.frame_count > 0, "The movie contains no frames!"
        self._current_frame = 0

        self.frame = self.get_movie_frame(self._current_frame)
        assert self.frame is not None, "Unable to read the movie file provided."

        self.vc = cv2.VideoCapture(filename=self.file_path)
```

It is probably in our best interest to avoid future issues to assume full control over the input arguments to the `VideoCapture`.

Also, in the current version of `nwb-conversion-tools`, we try to always refer to file based input arguments as `file_path`, and these lines would be the ideal place to perform the aliasing.

Also, I think the `stub` option would be more appropriate for the `run_conversion` step of the `MovieDataInterface`, so we can wait to perform it there and just have the context be in charge of general I/O to movie files.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+        return np.array(ts)
+
+    def get_movie_fps(self):
+        """
+        Return the internal frames per second (fps) for a movie file.
+
+        """
+        if int(cv2.__version__.split(".")[0]) < 3:
+            return self.vc.get(cv2.cv.CV_CAP_PROP_FPS)
+        return self.vc.get(cv2.CAP_PROP_FPS)
```
CodyCBakerPhD, line 52:
```suggestion
    def get_movie_fps(self):
        """Return the internal frames per second (fps) for a movie file."""
        if self.vc.isOpened():
            if int(cv2.__version__.split(".")[0]) < 3:
                return self.vc.get(cv2.cv.CV_CAP_PROP_FPS)
            return self.vc.get(cv2.CAP_PROP_FPS)
```

This could make access to this field even safer.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+        self.frame = self.get_movie_frame(0)
+        assert self.frame is not None, "unable to read the movie file provided"
+
+    def get_movie_timestamps(self):
+        """
+        Return numpy array of the timestamps for a movie file.
+
+        """
+        if not self.vc.isOpened():
+            raise ValueError("movie file is not open")
```
CodyCBakerPhD, line 37:
```suggestion
            raise ValueError("The movie file is not open!")
```
Minor punctuation suggestions.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+    def current_frame(self, frame_no):
+        if int(cv2.__version__.split(".")[0]) < 3:
+            set_arg = cv2.cv.CV_CAP_PROP_POS_FRAMES
+        else:
+            set_arg = cv2.CAP_PROP_POS_FRAMES
+        set_value = self.vc.set(set_arg, frame_no)
+        if set_value:
+            self._current_frame = frame_no
+        else:
+            raise ValueError(f"could not set frame no {frame_no}")
```
CodyCBakerPhD, line 88: How does this kind of frame scanning interact with differenced-based codecs? Are there any perform problems it might cause?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+        return self.frame.shape
+
+    def get_movie_frame_count(self):
+        """
+        Return the total number of frames for a movie file.
+
+        """
+        if self.stub:
+            # if stub the assume a max frame count of 10
+            return 10
```
CodyCBakerPhD, line 67:
```suggestion
        if self.stub:
            # if stub the assume a max frame count of 10
            return 10
```

See comment above; probably for the best if stubbing is outside the purview of this class.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+        if int(cv2.__version__.split(".")[0]) < 3:
+            count = self.vc.get(cv2.cv.CV_CAP_PROP_FRAME_COUNT)
+        else:
+            count = self.vc.get(cv2.CAP_PROP_FRAME_COUNT)
+        return int(count)
```
CodyCBakerPhD, line 72:
```suggestion
        if self.vc.isOpened():
            if int(cv2.__version__.split(".")[0]) < 3:
                count = self.vc.get(cv2.cv.CV_CAP_PROP_FRAME_COUNT)
            else:
                count = self.vc.get(cv2.CAP_PROP_FRAME_COUNT)
            return int(count)
```

This could make access to this field even safer.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+
+        """
+        if not self.vc.isOpened():
+            raise ValueError("movie file is not open")
+        ts = [self.vc.get(cv2.CAP_PROP_POS_MSEC)]
+        for i in tqdm(range(1, self.get_movie_frame_count()), desc="retrieving video timestamps"):
+            self.current_frame = i
+            ts.append(self.vc.get(cv2.CAP_PROP_POS_MSEC))
+        self.current_frame = 0
+        return np.array(ts)
```
CodyCBakerPhD, line 43: Do you know if this iteration is done lazily with respect to the reading of video frames? That is, whenever it tries to retrieve the `CAP_PROP_POS_MSEC` for a frame, does it also have to decompress/load the frame data?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+        Return the dtype for frame in a movie file.
+        """
+        return self.frame.dtype
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if not self.vc.isOpened():
+            self.vc = cv2.VideoCapture(*self._args, **self._kwargs)
```
CodyCBakerPhD, line 116: What is the intent of forcing the file opening here? Might be best to leave definitions of the `self.vc` for the `__enter__` and `__init__` stages, and assume that the `next` is only intentionally called after doing one of those two entry points.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+                if success:
+                    return frame
+                else:
+                    return np.nan * np.ones(self.get_frame_shape())
+            else:
+                self._current_frame = 0
+                self.vc.release()
+                raise StopIteration
+        except Exception:
+            raise StopIteration
```
CodyCBakerPhD, line 132: Exceptions in iteration contexts are usually very well defined - is there a particular non-`StopIteration` that gets raised unpredictably here?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,6 +15,134 @@
...
+                self._current_frame += 1
+                self.vc.release()
+                if success:
+                    return frame
+                else:
+                    return np.nan * np.ones(self.get_frame_shape())
+            else:
+                self._current_frame = 0
+                self.vc.release()
+                raise StopIteration
```
CodyCBakerPhD, line 130: Usually, for iterators, the effective 'generator' being called becomes exhausted. This is currrently being done via the `current_frame` property, but the idea is the same, hence it resetting to zero is a tad odd.

tests/test_internals/test_movie_utils.py
```diff
@@ -0,0 +1,78 @@
...
+                frames.append(vcc.get_movie_frame(no))
+        assert_array_equal(frames, self.movie_frames)
+
+    def test_iterable(self):
+        frames = []
+        with VideoCaptureContext(self.movie_loc) as vcc:
+            pass
+        for frame in vcc:
+            frames.append(frame)
+        assert_array_equal(np.array(frames), self.movie_frames)
```
CodyCBakerPhD, line 77:
```suggestion
        with VideoCaptureContext(self.movie_loc) as vcc:
            pass
            for frame in vcc:
                frames.append(frame)
        assert_array_equal(np.array(frames), self.movie_frames)
```

Did you intend this kind of nesting here? It's otherwise rather odd to alias the context to a variable which then gets referenced outside the context closure.

CodyCBakerPhD: Or maybe this is meant to just be

```python
        for frame in VideoCaptureContext(file_path=self.movie_loc):
            frames.append(frame)
        assert_array_equal(np.array(frames), self.movie_frames)
```

tests/test_internals/test_movie_utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_frame_shape(self):
+        with VideoCaptureContext(self.movie_loc) as vcc:
+            frame_shape = vcc.get_frame_shape()
+        assert_array_equal(frame_shape, self.frame_shape)
+
+    def test_frame_value(self):
+        frames = []
+        with VideoCaptureContext(self.movie_loc) as vcc:
+            for no in range(self.no_frames):
```
CodyCBakerPhD, line 67: Would be nice if this line indicated how the number of frames might be pulled from the `vcc` rather than the testing class.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: sounds good

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: yes, it makes sense to control the input arguments to the context with only `file_path` being allowed. I have made changes https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/4536888fedb0caed486a19ddb8b392491da03d95

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: yes, thanks for catching that. I think, better than returning `None` if not `self.vc.isOpened()`is to raise an Exception. 
In fact I am now asserting `isOpened()` for all these methods https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/783a2a84331272de0b29da93f862c6661d068046

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: setting stub internally helps to correctly iterate over when its used as an utterable:
```python
vcc = VideoCaptureContext("movie_file_path.avi", stub=True)
frames=[]
for frame in vcc:
   frames.append(frame)
len(frame) # this would be 10 
```
If we don't have the stub option then we would have to iterate over all frames (I/O for all frames) and then select only the stub from that. It would be wasteful. 
the `__next__` raises `StopIteration` if the `frame_no` is greater than that allowed by stub.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: I think, based on [this](https://stackoverflow.com/questions/47743246/getting-timestamp-of-each-frame-in-a-video), OpenCV assumes the frame rate to be constant and the timestamps retrieved using `cv2.CAP_PROP_POS_MSEC` is generated from the frame_rate. So it might be better that we give this up all together and just build from the frame rate https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/4536888fedb0caed486a19ddb8b392491da03d95

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: There are two possible cases, that we should take care of when `self.vc.read()` can fail: when the video is run out of frames or if the video is not being read like if its being streamed. both of these cases result in returning a `success=False` after `success, frame = self.vc.read()`. I've made it neater here: https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/5658f5801cbfac6c0d9c4a048c6b3f8e1bc773fd
Any other reason for the frame not being read, we should allow it to be thrown.

tests/test_internals/test_movie_utils.py
Saksham20: I want the Context class to be an iterable even if the `VideoCapture` object is closed. This is what this line is doing: it opening and then closing the `VideoCapture` object. Irrespective, we can still iterate over it. 
But I have reduced it here https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/42141abd2d89a53904f663d0f9bc94e3a3002893

tests/test_internals/test_movie_utils.py
Saksham20: yes, thanks for catching that.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: I wanted this be an iterable even if the object is closed: reinstantiating the `VideoWriter` attribute. If its open (we are iterating after instantiating the class) then we can skip this re-instantiation. 
Moreover, `__next__` is releasing it each time so we would need to instantiate it. 
But now that I think of it, it might be expensive to reinstantiate after every call to `__next__` since it releases it. The `StopIteration` would hit before releasing it would be necessary. 
It might be better to not release it after every call and do so only before `StopIteration`. Then we can also just enforce the `VideoWriter` to be open before iterating over it. https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/48db3f541b95b4fc41679a1e31f66eefa97290ed

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,9 +14,112 @@
...
+
+    def __enter__(self):
+        self.vc = cv2.VideoCapture(self.file_path)
+        return self
+
+    def __exit__(self, *args):
+        self.vc.release()
+
+    def __del__(self):
+        self.vc.release()
```
CodyCBakerPhD, line 118:
```suggestion
    def release(self):
        self.vc.release()

    def __exit__(self, *args):
        self.vc.release()

    def __del__(self):
        self.vc.release()
```

I'd suggest adding an additional command to release the internal attribute at user discretion. Both the other methods act more or less automatically, typically hidden to the user.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
CodyCBakerPhD: I would agree. By design, this class should intend to handle file opening in `__enter__` or `__init__`, and no where else. Likewise, file closing in a manual `release`, `__del__`, or `__exit__`. All other functionalities acting on the attribute as if it were open, and for safety, checking that it is.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
CodyCBakerPhD: That discussion indicates specificity to FFMPEG; I recall trying this out on a variety of actual real-world video files and in contrast found them to be highly irregular to the sampling rate

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
CodyCBakerPhD: Not necessarily, we would simply override the `my_video_capture.frame_count = stub_amount` after `my_video_capture = VideoCaptureContext(...)`; hence https://github.com/catalystneuro/nwb-conversion-tools/pull/392/files#diff-4482e8014d6730ce58954d23f828a997e8af7bd61b2d287c7b562f484e5e8733R97 would behave exactly the same way, but with full outside control of stubbing, and not the responsibility of this class.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -13,9 +14,112 @@
 PathType = Union[str, Path]
 
 
+class VideoCaptureContext:
+    """Retrieving video metadata and frames using a context manager"""
+
+    def __init__(self, file_path: FilePathType, stub=False):
+        self.vc = cv2.VideoCapture(file_path)
```
CodyCBakerPhD, line 21:
```suggestion
        self.vc = cv2.VideoCapture(filename=file_path)
```

I always strongly suggest using keywords when possible; we've had errors from the old days that were caused by non-transparent changes to positional orders and they were a complete pain to debug.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: yeah this is more user friendly. I've made the changes

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: Its a very core functionality of `VideoCapture`, its hard to imagine any other arg taking precedence over file path as first arg. Did you come across such opencv api changes?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: yeah, I like the idea of making `self.frame_count` a property and allow the user to set it to override the number of frames. Indeed it allows more flexibility over fixing a value if `stub=True`. I have made changes https://github.com/catalystneuro/nwb-conversion-tools/pull/392/files#diff-4482e8014d6730ce58954d23f828a997e8af7bd61b2d287c7b562f484e5e8733R45

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: There is a huge difference in time when retrieving timestamps for the two cases below:
```python
import cv2
from time import time
from tqdm import tqdm
file_path = "path_to_file.mp4"

# retrieve frames:
vc = cv2.VideoCapture(file_path)
no_frames = int(vc.get(cv2.CAP_PROP_FRAME_COUNT))
fps = vc.get(cv2.CAP_PROP_FPS)
print(f"no frames: {no_frames}, fps: {fps}")


start = time()
ts1 = []
for no in tqdm(range(no_frames//3000), desc="setting frames"):
    vc.set(cv2.CAP_PROP_POS_FRAMES, no)
    ts1.append(vc.get(cv2.CAP_PROP_POS_MSEC))

print(f"using set total time: {time()-start}")
vc.release()


vc = cv2.VideoCapture(file_path)
start = time()
ts2 = []
for no in tqdm(range(no_frames//3000),desc="retrieve frames"):
    success, frame = vc.read()
    ts2.append(vc.get(cv2.CAP_PROP_POS_MSEC))
    if not success:
        break

print(f"using set total time: {time()-start}")
vc.release()
```
Moreover, the ts1 and ts2 values differ. ts2 is accurate (checking against frame rate)
Can you try it on your end for a large video file?
The one I tried was with over 100k frames with 25 fps.

Saksham20: I have made the changes here: https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/75fbe1adaafddf41a4abd08b60da014722b121c0

CodyCBakerPhD: Good to know these kinds of things; very interesting that there is not only a speed difference but a value difference, which indicates need for some investigation. I'll run some on my end as well, then.

CodyCBakerPhD:
PR is really starting to take shape... still To-Do

> @Saksham20 You can and should also reference the full coverage report for this class to identify gaps in coverage: https://app.codecov.io/gh/catalystneuro/nwb-conversion-tools/blob/add_context_class_pre_pr/nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py

- [ ] There are a lot of lines that would be hard to cover due to `opencv` versions; what you can do for all those lines that work on non-`v3`, you can add a comment `# pragma: no cover` which will manually exclude them from coverage reports


- [ ] Variable names and other references using `no` to mean `number`, would be much more verbose to replace all these with the full English `number`.

Saksham20:
> There are a lot of lines that would be hard to cover due to opencv versions; what you can do for all those lines that work on non-v3, you can add a comment # pragma: no cover which will manually exclude them from coverage reports

It might be neater to include a configuration file, so we don't have to write anything in the code https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/e1aa9f61e6419f550d1dba62501a0d7c46f02f5b.
I think

Saksham20:
>> There are a lot of lines that would be hard to cover due to opencv versions; what you can do for all those lines that work on non-v3, you can add a comment # pragma: no cover which will manually exclude them from coverage reports

> @CodyCBakerPhD It might be neater to include a configuration file, so we don't have to write anything in the code https://github.com/catalystneuro/nwb-conversion-tools/commit/e1aa9f61e6419f550d1dba62501a0d7c46f02f5b.

Discussed with ben and made a few changes https://github.com/catalystneuro/nwb-conversion-tools/pull/392/commits/1fc8322754474fbdf3a5cb554d108f0a0e3cf85c 
`# pragma: no cover` is needed only on one line instead of three. 
I think, for now it does not make sense to have a separate config file just to tweak the codecov report for this file. 
@CodyCBakerPhD do you see other places we may need to change to codecov report?

CodyCBakerPhD:
> @CodyCBakerPhD do you see other places we may need to change to codecov report?

There's a couple but they are pretty minor so they can be added later. The static method for general attribute retrieval was a good idea.

Last two things to do with this PR are 
- [ ] Investigate timestamp value return to figure out the correct method; also with multiple different video formats (I'm working on this as well)
- [ ] Investigate how the frame 'seeking' method performs on various formats; the original idea was that for difference-based codecs (that base their compression/decompression off of differenced value of consecutive frames) seeking might require them to "un-difference" the frames up to the seek point, thus leading to sub-optimal performance. Would just be good for us to know when recommending or utilizing this feature. Could you work on that while I dig further into the timestamps?

CodyCBakerPhD:
@Saksham20 Also, this is at a good enough point to where you can update #372 to rely on this. You can also develop a PR branching from this that introduces the movie data chink iterator as a stand-alone feature (again with unit tests) prior to integration with the data interface.

Saksham20:
> @Saksham20 Also, this is at a good enough point to where you can update #372 to rely on this. You can also develop a PR branching from this that introduces the movie data chink iterator as a stand-alone feature (again with unit tests) prior to integration with the data interface.

I have updated #372 to be a PR to this branch instead. Merged this into that and the tests of MovieDataInterface pass.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
CodyCBakerPhD: OK here's some reading...

Unreliability of FFMPEG documentation, and how opencv relies on it a lot more than we would like underneath: https://stackoverflow.com/questions/70227392/opencvs-cap-setcv2-cap-prop-pos-msec-changes-to-unreachable-timestamp

Official maintainer comments on the unreliability of frame seeking and how frame-specific properties ought to really only be referred to on grabbed frames: https://github.com/opencv/opencv/issues/15749

Very acquisition-specific (non-python) approaches showing irregular timestamps can indeed be encoded, albeit explicitly: https://stackoverflow.com/questions/28086775/can-i-create-a-vfr-video-from-timestamped-images

More background reading on VFR (Variable Frame Rate): http://avisynth.nl/index.php/VFR#encoding_to_vfr_.28mkv.29

Most common use case in neuroscience I think I've seen is encoding of 'dropped' frames (produces irregular timestamps because some of the normally regular timestamps go missing). Also pretty common in general whenever a frame gets corrupted and we need to essentially skip it; I've run into cases of this during some Buzsaki conversion, but it will take me a while to find the exist files again...

CodyCBakerPhD: Now, what I've found from your script on a random `.mpg` file with 9100 frames at 30 fps: `ts2` is indeed accurate and regular w.r.t. the fps. The `ts1` field is equal to `ts2 / (fps * 100)`, indicating a kind of double-scaling bug where the frame number to time conversion is applied twice or something like that.

There is also the massive difference in time to extract these values, which related to the general seeking behavior for these codecs (almost answer two birds in one stone here); the set frame -> get timestamp (supposedly lazy method) takes me around 8 times longer than just reading each frame in succession and throwing it away (the accurate `ts2` method).

So, given these results, for our implementation here I really have to

(a) ask if we need to support allowing users to 'seek' to a particular frame for some reason. It seems very cost ineffective and potentially dangerous depending on how well the user knows these sorts of details that we are exploring

(b) it would be great if we could extract frames to be written to `data` and timestamps simultaneously, but that is only possible when the ultimate write method (outside the scope of this PR) is to not use iterative write (I.e., a data buffer). Otherwise, we have to fully extract the timestamps to determine if they are regular, in order to determine if we want to set the timestamps for the `ImageSeries`... So I guess we really should just implement and use the `ts2` method for all timestamp extraction by default.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+
+    def get_movie_timestamps(self):
+        """Return numpy array of the timestamps(s) for a movie file."""
+        ts2 = []
+        for no in tqdm(range(self.get_movie_frame_count()), desc="retrieving timestamps"):
+            success, frame = self.vc.read()
+            ts2.append(self.vc.get(cv2.CAP_PROP_POS_MSEC) / 1000)
+            if not success:
+                break
+        return np.array(ts2)
```
CodyCBakerPhD, line 40:
```suggestion
        timestamps= []
        for _ in tqdm(range(self.get_movie_frame_count()), desc="retrieving timestamps"):
            success, frame = self.vc.read()
            timestamps.append(self.vc.get(cv2.CAP_PROP_POS_MSEC))
            if not success:
                break
        return np.array(timestamps) / 1000
```

CodyCBakerPhD: Simple changes to code readability.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+
+    def get_movie_fps(self):
+        """Return the internal frames per second (fps) for a movie file"""
```
CodyCBakerPhD, line 43:
```suggestion
        """Return the internal frames per second (fps) for a movie file."""
```
`pydocstyle D400`. I'd suggest setting up some kind of linting in your IDE to make these easier to catch.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
     HAVE_OPENCV = True
 except ImportError:
     HAVE_OPENCV = False
+INSTALL_MESSAGE = "Please install opencv to use this interface! (pip install opencv-python)"
 
 PathType = Union[str, Path]
 
 
+class VideoCaptureContext:
+    """Retrieving video metadata and frames using a context manager"""
```
CodyCBakerPhD, line 22:
```suggestion
    """Retrieving video metadata and frames using a context manager."""
```
`pydocstyle D400`. I'd suggest setting up some kind of linting in your IDE to make these easier to catch.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+            raise ValueError(f"could not set frame no {frame_number}")
+
+    def get_movie_frame(self, frame_number: int):
+        """Return the specific frame from a movie."""
+        assert self.isOpened(), self._movie_open_msg
+        assert frame_number < self.get_movie_frame_count(), "frame number is greater than length of movie"
+        self.current_frame = frame_number
+        success, frame = self.vc.read()
+        self.current_frame = 0
+        return frame
```
CodyCBakerPhD, line 100:
```suggestion
    def get_movie_frame(self, frame_number: int):
        """Return the specific frame from a movie."""
        assert self.isOpened(), self._movie_open_msg
        assert frame_number < self.get_movie_frame_count(), "frame number is greater than length of movie"
        initial_frame_number = self.current_frame
        self.current_frame = frame_number
        success, frame = self.vc.read()
        self.current_frame = initial_frame_number 
        return frame
```
Let's say you partially iterate (or explicitly set the current frame to some value), then explicitly ask for some other frame (before or after the point you stopped iterating); if you reset the current frame to zero, you erase the point at which the iteration stopped (or whatever value you were otherwise saved at). This instead restores it to act more like an in-place operation.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+        assert self.isOpened(), self._movie_open_msg
+        prop = self.get_cv_attribute("CAP_PROP_FRAME_COUNT")
+        return int(self.vc.get(prop))
+
+    @property
+    def current_frame(self):
+        return self._current_frame
+
+    @current_frame.setter
+    def current_frame(self, frame_number):
```
CodyCBakerPhD, line 84:
```suggestion
    def current_frame(self, frame_number: int):
```

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+
+    @frame_count.setter
+    def frame_count(self, val):
+        self._frame_count = val
+
+    def get_movie_frame_count(self):
+        return self.frame_count
+
+    @staticmethod
+    def get_cv_attribute(attribute_name):
```
CodyCBakerPhD, line 68:
```suggestion
    def get_cv_attribute(attribute_name: str):
```

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+        if self._frame_count is None:
+            self._frame_count = self._movie_frame_count()
+        return self._frame_count
+
+    @frame_count.setter
+    def frame_count(self, val):
+        self._frame_count = val
+
+    def get_movie_frame_count(self):
+        return self.frame_count
```
CodyCBakerPhD, line 65:
```suggestion
    def get_movie_frame_count(self):
        return self.frame_count
```
The point to using the `@property` decorator is so that the function you apply the decorator to is defined to be the 'getter' method for that property. Was there some other reason you had this additional method that does exactly the same thing as `self.frame_count`?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+
+    @property
+    def frame_count(self):
+        if self._frame_count is None:
+            self._frame_count = self._movie_frame_count()
+        return self._frame_count
+
+    @frame_count.setter
+    def frame_count(self, val):
+        self._frame_count = val
```
CodyCBakerPhD, line 62:
```suggestion
    @frame_count.setter
    def frame_count(self, val: int):
        assert val > 0, "You must set a positive frame_count (received {val})."
        assert val <= self._movie_frame_count(), "Cannot set manual frame_count beyond length of video ({val})."
        self._frame_count = val
```
One main use of `@property.setter` functions is to provide some type of input validation for manually overriding things that are otherwise internal/private values.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+    def get_cv_attribute(attribute_name):
+        if int(cv2.__version__.split(".")[0]) < 3:  # pragma: no cover
+            return getattr(cv2.cv, "CV_" + attribute_name)
+        return getattr(cv2, attribute_name)
+
+    def _movie_frame_count(self):
+        """Return the total number of frames for a movie file."""
+        assert self.isOpened(), self._movie_open_msg
+        prop = self.get_cv_attribute("CAP_PROP_FRAME_COUNT")
+        return int(self.vc.get(prop))
```
CodyCBakerPhD, line 77: Recommend organizing this function so that it immediately follows the `frame_count` properties where it is used, to make this more readable.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+
+    @current_frame.setter
+    def current_frame(self, frame_number):
+        assert self.isOpened(), self._movie_open_msg
+        set_arg = self.get_cv_attribute("CAP_PROP_POS_FRAMES")
+        set_value = self.vc.set(set_arg, frame_number)
+        if set_value:
+            self._current_frame = frame_number
+        else:
+            raise ValueError(f"could not set frame no {frame_number}")
```
CodyCBakerPhD, line 91:
```suggestion
            raise ValueError(f"Could not set frame number (received {frame_number}).")
```

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
+    def isOpened(self):
+        return self.vc.isOpened()
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        assert self.isOpened(), self._movie_open_msg
+        if self._current_frame < self.frame_count:
+            self.current_frame = self._current_frame
```
CodyCBakerPhD, line 120: There's going to be an unnecessary performance issue here by calling the setter for `current_frame` (which forcibly sets the frame position) each time we increment the count; as this is a magic method, it should be sufficient to simply increment the private reference directly and rely on the `vc.read()` to set the frame position in the capture object itself.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -1,21 +1,146 @@
...
 
+class VideoCaptureContext:
+    """Retrieving video metadata and frames using a context manager"""
+
+    def __init__(self, file_path: FilePathType):
+        assert HAVE_OPENCV, INSTALL_MESSAGE
+        self.vc = cv2.VideoCapture(filename=file_path)
+        self.file_path = file_path
+        self._current_frame = 0
+        self._frame_count = None
```
CodyCBakerPhD, line 29:
```suggestion
        self._frame_count = self._movie_frame_count()
```
Would this make sense to do?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: yes, that's right, its safer to do this internally than expecting to set `self.current_frame=no` to the previous value.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: yes, I know this is redundant but I wanted to keep this method to be consistent with the other methods : `get_movie_timestamps()`, `get_movie_fps()`, `get_frame_shape()` which are similar to how we had before when we had independent functions.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: I wanted to keep the `__init__` as light as possible, not retrieving anything from the movie other than instantiating the `VideoWriter` object.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@Saksham20 Thanks for working hard on this - I'm being extra particular about having this kind of stuff polished because it's likely to be used not just for writing to NWB but also, in conjunction with the new external link capability on DANDI, for all side-by-side video management.

PR #394, opened 2022-02-04
Modify neuroscope data interface to use spikeextractors

h-mayorquin:
## Motivation
This follow the update of the Neuroscope data interface to fully use `spikeinterface` objects.

This is a draft that I made to see the results on the tests on github actions. I will modify this to a more complete description soon.

## How to test the behavior?
All the tests should pass when the change is fully implemented.

## Checklist

- [] Have you ensured the PR description clearly describes the problem and solutions?


nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -52,22 +53,36 @@ def add_recording_extractor_properties(recording_extractor: se.RecordingExtracto
...
+            )
+    else:
+        # Channel groups
+
+        # Group names
+        group_names = [group_names[channel_map[channel_id]] for channel_id in channel_id_list]
+        recording_extractor.set_property(key="group_name", ids=channel_id_list, values=group_names)
+        # Shank electrode number
+        group_electrode_numbers = [group_electrode_numbers[channel_map[channel_id]] for channel_id in channel_id_list]
+        recording_extractor.set_property(key="shank_electrode_number", ids=channel_id_list, values=group_names)
```
CodyCBakerPhD, line 76: For blocks like this I'd rather just coerce the object to a `OldtoNew` instead of maintaining the entire logic compatible with the old version. So the logic ought only reflect `BaseRecording` objects, and we obtain compatability with previous version purely through SI and their `old_api_utils`

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -52,22 +53,36 @@ def add_recording_extractor_properties(recording_extractor: se.RecordingExtracto
...
 
 
 class NeuroscopeRecordingInterface(BaseRecordingExtractorInterface):
     """Primary data interface class for converting a NeuroscopeRecordingExtractor."""
 
-    RX = se.NeuroscopeRecordingExtractor
+    # RX = se.NeuroscopeRecordingExtractor
+    from spikeinterface.extractors import NeuroScopeRecordingExtractor
+
+    RX = NeuroScopeRecordingExtractor
```
CodyCBakerPhD, line 85: Changing this one line is actually the biggest change here; I'd actually appreciate having an entirely separate PR just for this one line so that we can do a performance comparison. I've also toyed with the idea of having the back end extractor be specifiable as an option, what do you think @h-mayorquin?

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -406,10 +406,11 @@ def add_electrodes(
...
             # checked_recording.get_channel_locations defaults to np.nan if there are none
-            location = checked_recording.get_channel_locations(channel_ids=[channel_id])[0]
-            if all([not np.isnan(loc) for loc in location]):
-                # property 'location' of RX channels corresponds to rel_x and rel_ y of NWB electrodes
-                electrode_kwargs.update(dict(rel_x=float(location[0]), rel_y=float(location[1])))
+            if checked_recording.get_property("location") is not None:
+                location = checked_recording.get_channel_locations(channel_ids=[channel_id])[0]
+                if all([not np.isnan(loc) for loc in location]):
+                    # property 'location' of RX channels corresponds to rel_x and rel_ y of NWB electrodes
+                    electrode_kwargs.update(dict(rel_x=float(location[0]), rel_y=float(location[1])))
```
CodyCBakerPhD, line 413: These look like more general changes, could you move them to a different PR?

Also this logic could likely be improved since I assume `checked_recording.get_channel_locations()` returns vector with their missing values handled automatically, so we could do it outside a loop in fewer lines.

PR #396, opened 2022-02-07
Python 3.9 compatibility

CodyCBakerPhD:
## Motivation
Python 3.9 requires a couple version bumps to install correctly.

Also disabling the windows codecov uploads since there aren't any windows specific tests to capture; this should reduce the codecov spam a bit.


bendichter:
lgtm in principle, but it looks like we are having an issue with scipy 1.8 in python 3.7

CodyCBakerPhD:
Yup, I noticed that. We're probably going to have to have python-version specific `requirements-full.txt` files, unless you can think of a better way to handle that?

Basically, I would propose a `requirements-full.txt` file that excludes exact `scipy` and `h5py` version pinning, then a `requirements-full-3-8.txt` that pins to <=3.8 compatabible versions, then `requirements-full-3-9.txt` that pins to these versions

CodyCBakerPhD:
As much as I prefer to have the `requirements-full.txt` pin exact dependencies so there isn't any unexpected changes when they release new versions of those packages, it doesn't seem possible to have a single requirements file with exact pinning that works across all the python versions.

What I've done here is set lower bounds for those problem packages in the `requirements-full.txt`, and it seems to do the trick because the pip installer will auto-detect the latest version of the package available for that python version.

The only alternative I can think of would be to add separate `requirements-full-{python_version}.txt` files, each containing different version pins corresponding to each python version, and ask the user to install from those separately. But that seems a bit inefficient...

bendichter (approved):

Issue #403, opened 2022-02-24
Support non-int channel_ids


alejoe91:
## Description

SpikeInterface allows channel_ids to be of any type, but currently the `write_recording` function expects `int`

## Steps to Reproduce

```
import spikeinterface.full as si
from nwb_conversion_tools.utils.spike_interface import write_recording
from datetime import datetime

rec, sort = si.toy_example(num_segments=1)
rec_str = rec.channel_slice(channel_ids=rec.get_channel_ids(), 
                            renamed_channel_ids=[str(ch) for ch in rec.get_channel_ids()])
print(rec_str.get_channel_ids())

start_time = datetime.now()
metadata = {"NWBFile": {'session_start_time': start_time}}
write_recording(rec_str, save_path="test_nwb_str.nwb", metadata=metadata)
```

### Error

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/var/folders/vl/0vmqjsjn0dg8ycnxdx_mqrdw0000gr/T/ipykernel_71428/3145529863.py in <module>
----> 1 write_recording(rec_str, save_path="test_nwb_str1.nwb", metadata=metadata)

~/Documents/Codes/catalyst/nwb-conversion-tools/nwb_conversion_tools/utils/spike_interface.py in write_recording(recording, save_path, overwrite, nwbfile, starting_time, use_times, metadata, write_as, es_key, write_electrical_series, write_scaled, compression, compression_opts, iterator_type, iterator_opts)
    977                 nwbfile = make_nwbfile_from_metadata(metadata=metadata)
    978 
--> 979             add_all_to_nwbfile(
    980                 recording=recording,
    981                 nwbfile=nwbfile,

~/Documents/Codes/catalyst/nwb-conversion-tools/nwb_conversion_tools/utils/spike_interface.py in add_all_to_nwbfile(recording, nwbfile, starting_time, use_times, metadata, write_as, es_key, write_electrical_series, write_scaled, compression, compression_opts, iterator_type, iterator_opts)
    825         assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
    826 
--> 827     add_electrodes_info(recording=recording, nwbfile=nwbfile, metadata=metadata)
    828 
    829     if write_electrical_series:

[...]

PR #404, opened 2022-02-24
Si channel ids

alejoe91:
Fix https://github.com/catalystneuro/nwb-conversion-tools/issues/403

## Motivation

Support non-int channel_ids form spikeinterface

## How to test the behavior?
```
import spikeinterface.full as si
from nwb_conversion_tools.utils.spike_interface import write_recording
from datetime import datetime

rec, sort = si.toy_example(num_segments=1)
rec_str = rec.channel_slice(channel_ids=rec.get_channel_ids(), 
                            renamed_channel_ids=[str(ch) for ch in rec.get_channel_ids()])
print(rec_str.get_channel_ids())

start_time = datetime.now()
metadata = {"NWBFile": {'session_start_time': start_time}}
write_recording(rec_str, save_path="test_nwb_str.nwb", metadata=metadata)
```

## Checklist


CodyCBakerPhD:
@alejoe91 Can you add some tests for it?

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -398,10 +398,14 @@ def add_electrodes(
     for name in elec_columns_append:
         _ = elec_columns.pop(name)
 
+    # add channel_id column
+    nwbfile.add_electrode_column(name="channel_id", description="channel_id", index=False)
```
h-mayorquin, line 402: I was playing with this and it gives problems in the case were the same `nwb` file is re-written by another extractor.
Maybe?
```suggestion
    if 'channel_id' not in nwbfile.electrodes.colnames:
        nwbfile.add_electrode_column(name="channel_id", description="channel_id", index=False)
```

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -398,10 +398,14 @@ def add_electrodes(
...
+
     for j, channel_id in enumerate(checked_recording.get_channel_ids()):
         if channel_id not in nwb_elec_ids:
```
h-mayorquin, line 405: Note that `nwb_elec_ids=nwbfile.electrodes.id.data[:]`.

That is, maybe we should have?
```suggestion
        if checked_recording.id_to_index(channel_id) not in nwb_elec_ids:
```

h-mayorquin (commented):
I went through the discussion of this in the slack. Wasn't @CodyCBakerPhD suggestion there to create a new property called 'channel_name' that would be the string version of `channel_id`?

Looking at the testing it seems that we have some code that makes the assumption that the ids of the electrodes table in the `nwbfile` is equivalent to the `channel_ids` in the extractors:

https://github.com/catalystneuro/nwb-conversion-tools/blob/99d483794f17be7285637c9047f0fda6fec4dd93/tests/test_internals/test_si.py#L522

tests/test_internals/test_si.py
```diff
@@ -577,6 +576,23 @@ def test_group_set_custom_description(self):
                 else:
                     assert nwb.electrodes["group_name"][i] == "M1"
                     assert nwb.electrodes["group"][i].description == "M1 description"
+                    
+    def test_non_ints_as_channel_ids(self):
+        channel_ids = self.RX.get_channel_ids()
+        renamed_channel_ids = [str(id) for id in channel_ids]
+        self.RX3 = se.subrecordingextractor.SubRecordingExtractor(
+            parent_recording=self.RX, channel_ids=channel_ids, renamed_channel_ids=renamed_channel_ids,
```
alejoe91: @h-mayorquin note that non-int channels are not formally supported by the old SI API. I think it would be great to substitute the extractor with a new-API extractor for the test.

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: > I went through the discussion of this in the slack. Wasn't @CodyCBakerPhD suggestion there to create a new property called 'channel_name' that would be the string version of channel_id?

Indeed; the electrode table also has a primary ID for each electrode, the field `electrode.id`, a unique integer. Naming this new column `channel_id` could then get confusing if the two differ, whereas an aliasing column `channel_name` would be the arbitrary string version of the associated ID.

One question is whether to bother writing the column `channel_name` in the case where they correspond one to one (`int(electrode.channel_name) == electrode.id`). A point in favor of this would be to allow better representation of existing electrodes when appending new electrodes that do have differing names from the IDs.

CodyCBakerPhD: Also as per below, it would be nice for us to internally rely on the existence of the `channel_name` column so we can always form an index map between `channel_name` and `id`.

tests/test_internals/test_si.py
h-mayorquin: Ok. It should be like this now.

CodyCBakerPhD:
Closing in lieu of #415

PR #405, opened 2022-02-24
adding movie data chunk iterator

Saksham20:
adding data chunk iterator for movie frames. Inherits from hdmf's `GenericDatachunkIterator`. 
Continuation of PR: #372 

## Checklist


nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +142,51 @@ def __exit__(self, *args):
...
     def __del__(self):
         self.vc.release()
+
+
+class MovieDataChunkIterator(GenericDataChunkIterator):
+    """DataChunkIterator specifically for use on RecordingExtractor objects."""
+
+    def __init__(
+        self,
+        movie_file: PathType,
```
CodyCBakerPhD, line 152:
```suggestion
        movie_file: FilePathType,
```

We use the TypeVar's from `utils.json_schema` these days; you should be able to delete the `PathType` definition above.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +142,51 @@ def __exit__(self, *args):
...
+        chunk_mb: float = None,
+        chunk_shape: tuple = None,
+        stub: bool = False,
+    ):
+        self.video_capture_ob = VideoCaptureContext(movie_file)
+        if stub:
+            self.video_capture_ob.frame_count = 10
+        self._default_chunk_shape = False
+        if chunk_shape is None:
+            chunk_shape = (1, *self.video_capture_ob.get_frame_shape())
```
CodyCBakerPhD, line 164: 👍

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +142,51 @@ def __exit__(self, *args):
...
+    """DataChunkIterator specifically for use on RecordingExtractor objects."""
+
+    def __init__(
+        self,
+        movie_file: PathType,
+        buffer_gb: float = None,
+        buffer_shape: tuple = None,
+        chunk_mb: float = None,
+        chunk_shape: tuple = None,
+        stub: bool = False,
```
CodyCBakerPhD, line 157:
```suggestion
        stub_test: bool = False,
```

What would you think about swapping from `stub` to `stub_test` for consistency with the rest of the usage across the repo?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +142,51 @@ def __exit__(self, *args):
...
+
+    def _get_data(self, selection: Tuple[slice]) -> Iterable:
+        if self._default_chunk_shape:
+            self._current_chunk += 1
+            frame = next(self.video_capture_ob)
+            return frame[np.newaxis, :]
+        frames_return = []
+        step = selection[0].step if selection[0].step is not None else 1
+        for frame_no in range(selection[0].start, selection[0].stop, step):
+            frame = self.video_capture_ob.get_movie_frame(frame_no)
```
CodyCBakerPhD, line 183:
```suggestion
        for frame_number in range(selection[0].start, selection[0].stop, step):
            frame = self.video_capture_ob.get_movie_frame(frame_number)
```

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +142,51 @@ def __exit__(self, *args):
...
+            frames_return.append(frame[selection[1:]])
+            self._current_chunk += 1
+        return np.concatenate(frames_return, axis=0)
```
CodyCBakerPhD, line 186: Since this is a private function we can make some stronger (and safer) assumptions about how the data is accessed and returned than we could if this was an all-purpose retrieval (like `recording.get_traces` is).

As such here are a couple suggestions:

i) The selections that will always feed into this come from https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/data_utils.py#L237 - as such, the stuff about step sizes is unnecessary and worst case could allow the iterator to skip frames (data loss in writing into the final `h5py.Dataset`)

ii) Similar to how we did it in the integration of the context class with the movie interface, I would pre-allocate the numpy array so it already has a block in memory designated. Otherwise the `np.concatenate` at the end (or any new array casting) will double the literal amount of RAM taken up by the buffer (it will clear it quickly after filling and returning, but still could cap system usage).

iii) What is the `_current_chunk` counter for? I don't see it being utilized anywhere

iv) Similar to (i) we can make stronger assumptions about order of selections (namely that they are consecutive), and repeated calls to `self.video_capture_ob.get_movie_frame` result in unnecessary overhead via the calls to the `frame_count` setter (https://github.com/catalystneuro/nwb-conversion-tools/blob/6ad1d4211eeb0c4955130c9beb46bac40b8747df/nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py#L102). I'd recommend just doing what you did for the default chunk logic and repeat the calling of `next` on the capture iterable, filling those frames into the corresponding part of the pre-allocated numpy array.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +142,51 @@ def __exit__(self, *args):
...
+        chunk_shape: tuple = None,
+        stub: bool = False,
+    ):
+        self.video_capture_ob = VideoCaptureContext(movie_file)
+        if stub:
+            self.video_capture_ob.frame_count = 10
+        self._default_chunk_shape = False
+        if chunk_shape is None:
+            chunk_shape = (1, *self.video_capture_ob.get_frame_shape())
+            self._default_chunk_shape = True
```
CodyCBakerPhD, line 165:
```suggestion
```
I do apologize for any confusion on this, it's the most frustrating part of any DataChunkIterator class.

The `chunk_shape` and `chunk_size` refer to the actual chunking method of the `h5py.Dataset`, so this shape is mostly about streamlining read/write into and out of that object.

The `buffer_shape` and `buffer_size` arguments refer to how we actually load data from the file into RAM; the selections passed into `_get_data` are precisely these.

For this particular Iterator, I would strongly recommend overriding the default `buffer_shape` with one that fits as many full frames into RAM as possible up to the `buffer_size`. If a manual `buffer_shape` is specified it should also assert that the shape does not subset the frame size (which basically leaves the user with one free parameter - the number of frames).

Actually, that means we probably shouldn't even accept `buffer_shape` as an argument here and instead maybe just accept `buffer_size` instead since it would then be one-to-one with the number of frames.

Now, as for default chunk shape... chunk sizes should be right around or below 1MB in size. That's a roughly a single frame of a 577 x 577 RGB video and while that's in the ballpark for the historical data we prototype, on modern cameras could easily go over that. So I would add some logic for dividing up the frame shape in that situation so that each chunk stays below 1MB.

CodyCBakerPhD:
May also want to add some more tests to ensu

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: > So I would add some logic for dividing up the frame shape in that situation so that each chunk stays below 1MB.

What is the reason to keep chunks below 1MB? 
Buffer size I understand, we want to default to a reasonable value to be RAM friendly.

Saksham20:
currently,
The moviedatachunkierator allows either `chunk_shape` or `buffer_gb` as args.  
Since buffer should be filled with individual frames (limiting to the buffer_gb), `buffer_shape` is useless. 
Ideally, even the chunks will be individual frames in case of movies. However, if the user prefers to access only a segment of the frame then `chunk_shape` can come handy. If not specified, then we assume the chunk_size to be 1.0 MB and build a shape depending upon the frame size but pulling in all the channels (last dim) of each frame.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
CodyCBakerPhD: See [this part](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/data_utils.py#L189-L191) of the documentation for the `GenericDataChunkIterator` and the report by the HDF5 group.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,73 @@ def __exit__(self, *args):
...
+    @staticmethod
+    def _get_default_shape_movie(size_mb, shape, size, max_shape):
+        """Given the shape and size of array, return shape that will fit size_mb."""
+        k = np.floor((size_mb / size) ** (1 / len(shape)))
+        return tuple([min(max(int(x), shape[j]), max_shape[j]) for j, x in enumerate(k * np.array(shape))])
+
+    def _get_frame_details(self):
+        """Get frame shape and size in MB"""
+        frame_shape = (1, *self.video_capture_ob.get_frame_shape())
+        min_frame_size = (np.prod(frame_shape) * self._get_dtype().itemsize) / 10e6
```
CodyCBakerPhD, line 189:
```suggestion
        min_frame_size = (np.prod(frame_shape) * self._get_dtype().itemsize) / 1e6
```
Did you mean 1MB here?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,73 @@ def __exit__(self, *args):
...
+        movie_file: FilePathType,
+        buffer_gb: float = None,
+        chunk_shape: tuple = None,
+        stub_test: bool = False,
+    ):
+        self.video_capture_ob = VideoCaptureContext(movie_file)
+        buffer_gb = 1.0 if buffer_gb is None else buffer_gb
+        self._maxshape = self._get_maxshape()
+        self._full_frame_size_mb, self._full_frame_shape = self._get_frame_details()
+        self._dtype = self._get_dtype()
```
CodyCBakerPhD, line 158:
```suggestion
        self._full_frame_size_mb, self._full_frame_shape = self._get_frame_details()
```

These are covered in the `super().__init()` call: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/data_utils.py#L208-L209

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,73 @@ def __exit__(self, *args):
...
+        return self._get_default_shape_movie(
+            size_gb * 1e3, self._full_frame_shape, self._full_frame_size_mb, self._maxshape
+        )
+
+    def _get_data(self, selection: Tuple[slice]) -> Iterable:
+        start_frame = selection[0].start
+        end_frame = selection[0].stop
+        frames = np.empty(shape=[end_frame - start_frame, *self._maxshape[1:]])
+        for frame_no in range(end_frame - start_frame):
+            frames[frame_no] = self.video_capture_ob.__next__()
```
CodyCBakerPhD, line 204:
```suggestion
            frames[frame_no] = next(self.video_capture_ob)
```

Is there a difference between these two ways of calling it?

CodyCBakerPhD:
@Saksham20 Really coming together now; I'll take a deeper dive over the weekend.

In the meantime looks like you need a specific test for the `else` condition of `_get_default_chunk_shape_movie`.

Saksham20:
> See [this part](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/data_utils.py#L189-L191) of the documentation for the GenericDataChunkIterator and the report by the HDF5 group.

From what I understand from the documentation above, the reading speed depends a lot on the access pattern of the user. In the case of movies, its fair to assume that the access pattern will be each frame completely (unless the user specifies a custom chunk shape). 
If its a complete frame in one shot, then even if the chunk cache is smaller than the uncompressed frame size, it does not matter as it will be read in one go. 
Caching the chunk adds to read speed only when it needs to be accessed multiple times (so it avoids the IO + decompress operation). This would be valid if the user were to read the frame rows/columns wise. 

Thus we can hard code the `chunk_shape` as `frame_shape`. If the frame size happens to be smaller than 1MB then the function checks if we can fit multiple frames in 1MB else it defaults to 1 frame. https://github.com/catalystneuro/nwb-conversion-tools/pull/405/commits/82a5d15478b612e2bd6dbbc657c6db0543fd2554

Saksham20:
TODO: 
- [ ] Tests update.

CodyCBakerPhD:
> its fair to assume that the access pattern will be each frame completely (unless the user specifies a custom chunk shape).

It's probably fine, but here's 2 points about that. (i) most users will not know nor care to learn the low level details of optimal chunking based on expected access usage, (ii) that's true if their only use for storing the movie as a `h5py.Dataset` is to view each frame in its entirety individually; there are cases, however, where position inference (say, via LED tracking) is highly localized to a section of all frames. In this case the user or an external tool may attempt to make slice requests that span multiple frames but only a small subset of each frame. Refer to page 20 of the HDF5 document to see how (in a worst case scenario) this can affect read speed.

> If its a complete frame in one shot, then even if the chunk cache is smaller than the uncompressed frame size, it does not matter as it will be read in one go. Caching the chunk adds to read speed only when it needs to be accessed multiple times (so it avoids the IO + decompress operation). This would be valid if the user were to read the frame rows/columns wise.

There are several aspects to the speed resultant of choice of chunk shape; the caching itself, yes, exists for multiple access to compressed data.

But optimal compression efficiency also depends on chunking over sections of data with high commonality, which usually means a local patch over time with respect to natural movie data. This is important not just for saving as much storage space as possible, but the speed of reading/writing the chunks depends on how efficient the compression was.

This is also why I request sub-sectioning a single frame if it's too large; if it's a modern high-res camera, and the user has any potential to want to retrieve a subsection of every frame over time, then a hard-enforced rule of per-frame compression will result in a lot of wasted retrieval with or without compression (the HDF5 doc also illustrates chunking with and without compression).

For this first PR it's probably fine just to go with it as is; I can always come back and tune the defaults more later.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,64 @@ def __exit__(self, *args):
...
+            buffer_gb=buffer_gb,
+            chunk_shape=chunk_shape,
+            display_progress=True,
+        )
+
+    def _get_default_chunk_shape(self, chunk_mb):
+        """Shape is either one frame or a subset: scaled frame size but with all pixel colors"""
+        no_frames = int(chunk_mb // self._full_frame_size_mb)
+        no_frames = 1 if no_frames is 0 else no_frames
+        return (no_frames,) + self._full_frame_shape[1:]
```
CodyCBakerPhD, line 168:
```suggestion
        num_frames = int(chunk_mb // self._full_frame_size_mb)
        num_frames = 1 if no_frames is 0 else no_frames
        return (num_frames ,) + self._full_frame_shape[1:]
```
'no frames' has a different meaning than 'number of frames'. It's also very rare to see the № symbol used these days

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,64 @@ def __exit__(self, *args):
...
+        k = np.floor((size_mb / size) ** (1 / len(shape)))
+        return tuple([min(max(int(x), shape[j]), max_shape[j]) for j, x in enumerate(k * np.array(shape))])
+
+    def _get_frame_details(self):
+        """Get frame shape and size in MB"""
+        frame_shape = (1, *self.video_capture_ob.get_frame_shape())
+        min_frame_size_mb = (np.prod(frame_shape) * self._get_dtype().itemsize) / 1e6
+        return min_frame_size_mb, frame_shape
+
+    def _get_data(self, selection: Tuple[slice]) -> Iterable:
```
CodyCBakerPhD, line 190:
```suggestion
    def _get_data(self, selection: Tuple[slice]) -> np.ndarray:
```
This private method actually does expect to return a buffered numpy array of data to be passed BY the primary Iterable, which is the class itself. The array itself is not iterated over.

Saksham20:
> But optimal compression efficiency also depends on chunking over sections of data with high commonality, which usually means a local patch over time with respect to natural movie data. This is important not just for saving as much storage space as possible, but the speed of reading/writing the chunks depends on how efficient the compression was.

Assuming, what you say is true: pixels will tend to be more similar across time than space, thus better compression, we would need to get IO speed metrics before concluding that this will definitely be true. 
Is it true that for higher compression ratios, the read/write speed is more?

> there are cases, however, where position inference (say, via LED tracking) is highly localized to a section of all frames. In this case the user or an external tool may attempt to make slice requests that span multiple frames but only a small subset of each frame.
The user writing the nwbfile ("writer") has to make the best guess on how a downstream user ("reader") may read the data. If its a video with LED tracking etc, IMO then its the "writers" responsibility to provide a custom chunk shape. 
For us as tool developers, should we not default to the most common case (reading the whole frame)? 
Yes indeed, if the "reader" wants a segment of the frame across time, and the data is chunked in frames, it will be slower. 

>There are several aspects to the speed resultant of choice of chunk shape; the caching itself, yes, exists for multiple access to compressed data.
is it that some chunk shapes are preferable for faster IO?

CodyCBakerPhD:
> pixels will tend to be more similar across time than space, thus better compression, we would need to get IO speed metrics before concluding that this will definitely be true.

It's true that for a _particular_ video this depends, and over the space of _all_ videos is can vary quite a bit. That's why we try to go for an 'average' expected access pattern so that it's not fully optimized for any particular one but is also not super non-optimized for others.

We haven't bothered going to the level of actual performance tuning during write procedures of such things. I'd expect the overhead for that might be a bit excessive to actually provide reliably improvements.

> Is it true that for higher compression ratios, the read/write speed is more?

Higher compression ratio means more data commonality and more efficient data reduction, thus less expensive to compress and to read. Lower compression ratio alternatively would mean a deeper encoding tree (see https://en.wikipedia.org/wiki/Deflate) with longer bit expressions for the larger percentage of less common elements, hence taking longer to span the tree when reconstructing each data element.

Worst case scenario, imagine your compression ratio was 1 - you've taken the time to write an entire alternative representation of the data. Another exaggeration of this would be writing white noise vs. deterministically structured data in line with known chunk design.

> The user writing the nwbfile ("writer") has to make the best guess on how a downstream user ("reader") may read the data. If its a video with LED tracking etc, IMO then its the "writers" responsibility to provide a custom chunk shape.

Exactly.

> Yes indeed, if the "reader" wants a segment of the frame across time, and the data is chunked in frames, it will be slower.

Correct.

> For us as tool developers, should we not default to the most common case (reading the whole frame)?

Are you sure that's the most common case?

> is it that some chunk shapes are preferable for faster IO?

Only with respect to the expected access patterns (see above discussion). I don't think it matters beyond that; we aren't talking about binary blobs where access is  faster when things are consecutive in memory, the location of these chunks gets hashed all over the space. That's why the size is what matters more due to the total cost of calculating where in memory all the requested pieces are that correspond to a particular slice.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,64 @@ def __exit__(self, *args):
...
+        super().__init__(
+            buffer_gb=buffer_gb,
+            chunk_shape=chunk_shape,
+            display_progress=True,
+        )
+
+    def _get_default_chunk_shape(self, chunk_mb):
+        """Shape is either one frame or a subset: scaled frame size but with all pixel colors"""
+        num_frames = int(chunk_mb // self._full_frame_size_mb)
+        num_frames = 1 if num_frames is 0 else num_frames
```
CodyCBakerPhD, line 167:
```suggestion
        num_frames = int(np.ceil(chunk_mb / self._full_frame_size_mb))
```
Would this work?

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,64 @@ def __exit__(self, *args):
...
+        return (num_frames,) + self._full_frame_shape[1:]
+
+    def _get_default_buffer_shape(self, buffer_gb):
+        """Buffer shape is a multiple of frame shape along the frame dimension."""
+        assert buffer_gb >= self._full_frame_size_mb / 1e3, f"provide buffer size >= {self._full_frame_size_mb/1e3} GB"
+        frames_count = self._get_default_shape_movie(
+            buffer_gb * 1e3, self._full_frame_shape[:1], self._full_frame_size_mb, self._maxshape[:1]
+        )
+        return tuple([frames_count[0], *self._maxshape[1:]])
```
CodyCBakerPhD, line 176: Why is the inner list wrapping needed here? I like what you did above with `(num_frames,) + self._full_frame_shape[1:]`

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
```diff
@@ -141,3 +139,64 @@ def __exit__(self, *args):
...
+
+    @staticmethod
+    def _get_default_shape_movie(size_mb, shape, size, max_shape):
+        """Given the shape and size of array, return shape that will fit size_mb."""
+        k = np.floor((size_mb / size) ** (1 / len(shape)))
+        return tuple([min(max(int(x), shape[j]), max_shape[j]) for j, x in enumerate(k * np.array(shape))])
```
CodyCBakerPhD, line 182: Just to get some perspective (I never tested this method on 4D data), can you report some example outputs returned by this for common video sizes you've worked with? What about extreme values like very high res pixels or very long movies?

tests/test_internals/test_movie_utils.py
```diff
@@ -110,3 +115,99 @@ def test_isopened_assertions(self):
...
+            params=None,
+        )
+        for k in range(number_of_frames):
+            writer.write(movie_frames[k, :, :, :])
+        writer.release()
+        return movie_file
+
+    def test_iterator_general(self):
+        movie_file = self.create_movie(self.fps, self.frame_shape, self.number_of_frames)
+        it = H5DataIO(MovieDataChunkIterator(movie_file), compression="gzip")
```
CodyCBakerPhD, line 150:
```suggestion
        it = H5DataIO(MovieDataChunkIterator(movie_file))
```
I don't think we need to use compression when running the tests here, probably just makes them take longer. Chunking should be handled automatically though within the `H5DataIO`

CodyCBakerPhD: Can also be changed in the rest of the testing file.

tests/test_internals/test_movie_utils.py
```diff
@@ -110,3 +115,99 @@ def test_isopened_assertions(self):
...
+        img_srs = ImageSeries(name="imageseries", data=it, unit="na", starting_time=None, rate=1.0)
+        self.nwbfile.add_acquisition(img_srs)
+        with NWBHDF5IO(path=self.nwbfile_path, mode="w") as io:
+            io.write(self.nwbfile)
+            print(self.nwbfile_path)
+        with NWBHDF5IO(path=self.nwbfile_path, mode="r") as io:
+            nwbfile = io.read()
+            assert nwbfile.acquisition["imageseries"].data.chunks == (1,) + self.frame_shape
```
CodyCBakerPhD, line 158: I would highly recommend adding a full array assertion to make sure the frames out are equal to the frames in.

tests/test_internals/test_movie_utils.py
```diff
@@ -110,3 +115,99 @@ def test_isopened_assertions(self):
...
+
+    def test_iterator_stub(self):
+        movie_file = self.create_movie(self.fps, self.frame_shape, self.number_of_frames)
+        it = H5DataIO(MovieDataChunkIterator(movie_file, stub_test=True), compression="gzip")
+        img_srs = ImageSeries(name="imageseries", data=it, unit="na", starting_time=None, rate=1.0)
+        self.nwbfile.add_acquisition(img_srs)
+        with NWBHDF5IO(path=self.nwbfile_path, mode="w") as io:
+            io.write(self.nwbfile)
+            print(self.nwbfile_path)
```
CodyCBakerPhD, line 167: Are these print statements necessary?

CodyCBakerPhD:
@Saksham20 Could you also run some performance comparisons to the old iterator? Basically run a self-contained snippet like you do to write the movies in the tests here, but try running them on some decently sized real movie files and report the write speeds of this `Iterator` vs the old `DataChunkIterator` (i.e., the current `MovieDataInterface`). Feel free to vary the buffer size for the new iterator, too, but it's a little tricky to do that with the old one so it's OK if you just use the default as a baseline.

tests/test_internals/test_movie_utils.py
Saksham20: no, I had added them during debug, have removed them

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: this won't because I want `np.ceil` only if the value is between 0 and 1. Else I want `np.floor`.

tests/test_internals/test_movie_utils.py
Saksham20: do we not want to test the fact that given the custom `MovieDataChunkIterator`, `H5DataIO` is able to compress + write data correctly so that it can be read. 
I mean, is there a possibility if things run fine without compression but not with.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
Saksham20: because `self._full_frame_shape` is a tuple (which I define in this code) but `self._maxshape` is defined in the parent class elsewhere which we have no control over. I dont want to assume it to be a tuple. 
I can do: `frames_count + tuple(self._maxshape[1:])` instead

tests/test_internals/test_movie_utils.py
Saksham20: for tuples containing ints, I dont think `assert a==b` and `assert all([a[i]==j for i,j in enumerate(b)])` are any different

CodyCBakerPhD: Sorry for any confusion. I mean full array assertions on the `data` itself not just the chunks. Similar to how we examine full traces in the GIN `ecephys` testing data.

CodyCBakerPhD: The original way of testing the `.data.chunks` field is fine

tests/test_internals/test_movie_utils.py
CodyCBakerPhD: That would be the responsibility of unit tests of the `H5DataIO` on the HDMF side, which is very thoroughly done. Our only job is to make sure the data itself reads and writes through the IO (namely the interaction with chunking).

One time we do test compression setting is on the Interface level, where we are passing it as a higher-level input argument on our own side. But that's still not really for testing that the actual compression works, just testing that the call gets passed down.

nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py
CodyCBakerPhD: 👍

PR #374, opened 2022-01-17
Verion 0.10.3 changelog: spikeinterface.add_electrodes - Not for direct merge

CodyCBakerPhD:
## Motivation

A limited restoration of @Saksham20 and @alejoe91 proposed refactor of the `utils.spike_interface.add_electrodes` method. I'll look through everything while the remaining To-Do's are finished by those most familiar with the code and see if I can make any suggestions.

## To-Do's

- [ ] Fix tests
- [ ] Add docstring for new `conversion_tools` helper function `add_properties_to_dynamictable`
- [ ] unit tests for both new `conversion_tools` helper functions
- [ ] If any major changes to behavior to new `add_electrodes` functionality, addition of new tests covering that behavior and some documentation of the changes posted to this PR so we can reference it going forward

PR #406, opened 2022-02-25
Refactor add electrodes

h-mayorquin:
## Motivation
This is a small part of issue #374:
1) Basically, the properties that are set a the individual channel id level are set in bulk in line with the new capabilities of `spikeinterface`. 

## To-do
- [x] Remove the channel-by-channel reading in the extraction of the properties from the recording.
- [x] Check that all tests are passed.

## Checklist

-  [x] Have you checked our [Contributing](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/docs/contribute.rst) document?


h-mayorquin:
@CodyCBakerPhD 
The PR is ready for review:
From our discussion today there were two things that were missing:
1) The logic for handling the default data of previously used channels for which non-data is available in this recording. 
2) Casting of data to float.

I think I have a good idea of for a larger refactor of this function. But maybe we can close this PR as it is an implement the larger change in a different PR.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -322,77 +322,68 @@ def add_electrodes(
...
         if name not in default_updated:
             if nwbfile.electrodes is None:
                 nwbfile.add_electrode_column(name=name, description=des_args["description"], index=des_args["index"])
             else:
-                combine_data = [channel_property_defaults[found_property_types[name]]] * len(nwbfile.electrodes.id)
-                des_args["data"] = combine_data + des_args["data"]
+                samp_data = des_dict["data"][0]
+                data_type_found = [proptype for proptype in property_default_types if isinstance(samp_data, proptype)][
+                    0
+                ]
```
CodyCBakerPhD, line 380:
```suggestion
                data_type_found = [proptype for proptype in property_default_types if isinstance(des_dict["data"][0], proptype)][
                    0
                ]
```
`samp_data` wasn't being used again after this, right?

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -322,77 +322,68 @@ def add_electrodes(
...
+                extended_data = [property_default_types[data_type_found]] * len(nwbfile.electrodes.id)
+                previous_data = des_args["data"].tolist()
+                # Cast data to float
+                if data_type_found == Real:
+                    previous_data = [float(x) for x in previous_data]
```
CodyCBakerPhD, line 385:
```suggestion
                if data_type_found == Real:
                    previous_data = [float(x) for x in previous_data]
```
Most of the comment insertions above are somewhat useful to enhancing readability, but this is self-apparent.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -322,77 +322,68 @@ def add_electrodes(
...
                 nwbfile.add_electrode_column(name=name, description=des_args["description"], index=des_args["index"])
             else:
-                combine_data = [channel_property_defaults[found_property_types[name]]] * len(nwbfile.electrodes.id)
-                des_args["data"] = combine_data + des_args["data"]
+                samp_data = des_dict["data"][0]
+                data_type_found = [proptype for proptype in property_default_types if isinstance(samp_data, proptype)][
+                    0
+                ]
+                extended_data = [property_default_types[data_type_found]] * len(nwbfile.electrodes.id)
+                previous_data = des_args["data"].tolist()
```
CodyCBakerPhD, line 382: What makes this data 'previous'? Is it not what is 'currently' being added (for example, in append mode to an existing table)?

nwb_conversion_tools/utils/spike_interface.py
h-mayorquin: No.
The reason to define a name for it is that I found that many things were done inside the list comprehension and I wanted to make it more clear through naming in the variables that the operation that is happening there is that "the data type (name of the final variable) is found by the sample data the default property types to the sample data" 

The same mechanism is used in the loop before that: 
https://github.com/catalystneuro/nwb-conversion-tools/blob/86a2bfb32236c6aab0215b35cdb7ceceafa10df1/nwb_conversion_tools/utils/spike_interface.py#L358-L368

nwb_conversion_tools/utils/spike_interface.py
h-mayorquin: So, what I had in mind was (the data that was there *previous* to the extension with default types in `extended_data`). But after your comment I think it is confusing because there is also some data in the electrode table already and it might be confused with that. 

A better name might be `data_to_add` or just `data`. 

Btw, this is one of the lines that I think we can write way more clear once #404 is done. If we have a `channel_names` property we can explicitly fill with default types only those already existing channels for which no channel_name exists on this recorder (or something similar with the mapping between `channel_names` and `electrode.id` that we intend to make available).

nwb_conversion_tools/utils/spike_interface.py
h-mayorquin: Agree on this one.

nwb_conversion_tools/utils/spike_interface.py
```diff
@@ -273,8 +273,8 @@ def add_electrodes(
     else:
         checked_recording = recording
     if nwbfile.electrodes is not None:
-        ids_absent = [id not in nwbfile.electrodes.id for id in checked_recording.get_channel_ids()]
-        if not all(ids_absent):
+        ids_already_in_electrode_table = [id in nwbfile.electrodes.id for id in checked_recording.get_channel_ids()]
+        if all(ids_already_in_electrode_table):
             warnings.warn("cannot create electrodes for this recording as ids already exist")
             return
```
h-mayorquin, line 279: I was thinking about this.
@CodyCBakerPhD @Saksham20 

In the previous version in the diff we **only write the electrodes table if ALL the channels are new**. That is, we only write the electrodes table if NONE of the channels in the recording object were previously available in the electrodes table. In other words, if there is only one or two channels that are not in the electrodes table this not enough for us to add them to the table.

Am I misunderstanding? 
Is this the desired behavior? 
My naive understanding is that if there are channels whose properties were not written before maybe we should add them to the table? 

The version that I implemented in this change does that: **the only way you don't write the electrode table is if ALL the channels you are adding were previously available in the electrode table**. That is, there is no new channel in the recording object.

h-mayorquin (commented):
I added some discussion about the condition for writing the electrode table. I will note that the change in the diff does not affect our tests. -They were passing before- .

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: Yeah I hear you, we run into this from time to time when we're doing a list search that only returns a single element. I'd leave as is and refactor later; for reference the better way to do it is an explicit `for` loop like

```python
for proptype in property_default_types:
    if isinstance(nwbfile.electrodes[colname].data[0] , proptype)]:
        default_datatype = proptype
        break
```
The break command will then exit the `for` loop as soon as the `proptype` is found, so more efficient too.

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: I see now, it makes sense from that perspective. Should be fine. Looking forward to the next refactor!

nwb_conversion_tools/utils/spike_interface.py
CodyCBakerPhD: > The version that I implemented in this change does that: the only way you don't write the electrode table is if ALL the channels you are adding were previously available in the electrode table. That is, there is no new channel in the recording object.

This should indeed be the desired behavior. I do agree the previous logic would have falsely terminated if _some_ channels were and _some_ channels weren't present in the existing table.

CodyCBakerPhD: Good catch!

CodyCBakerPhD (approved):

PR #408, opened 2022-02-27
Provide a pip installable testing suite 

h-mayorquin:
## Motivation

Right now the default installation `pip install nwb_conversion_tools` uses the dependencies in the `requirements-minimal.txt`. However, using our complete testing suite needs the additional installation of the dependencies listed in `requirements-full.txt`. The goal of this pull request is to provide an easy and convenient way of installing all the necessary dependencies to run our complete test suit in an isolated environment.

More precisely, the following commands:
```bash
pip install nwb-conversion-tools[test_full]
pip install -e .[test_full]
pip install .[test_full]
```
Will automatically install both the requirements in `requirements-minimal.txt` and `requirements-full.txt` allowing us or other users to build rapidly an environment for trying our complete test suit.

I am also providing a companion file that automatically builds a `conda` environment with the same dependencies.

I also added `dandi` as a dependency that seems to be required by some of our tests.

## Checklist


setup.py
```diff
@@ -8,6 +8,12 @@
     long_description = f.read()
 with open(os.path.join(path, "requirements-minimal.txt")) as f:
     install_requires = f.read().strip().split("\n")
+with open(os.path.join(path, "requirements-full.txt")) as f:
+    test_requires = f.read().strip().split("\n")
+test_requires += ["pytest", "pytest-cov"]
```
bendichter, line 13: why not have these in requirements-full.txt?

bendichter:
this is cool, I didn't know about this feature of `setup`!

setup.py
h-mayorquin: I am not really sure. I wanted to do that but I was afraid it would break some of the CI stuff. I am hoping that can @CodyCBakerPhD comment on it.

requirements-full.txt
```diff
@@ -20,5 +21,5 @@ ndx-events==0.2.0
 parameterized==0.8.1
 pyintan==0.3.0
 pyopenephys==1.1.2
-sonpy
+sonpy>=1.7.1
```
bendichter, line 24: usually requirements-full specifies an exact version. Is that possible here?

h-mayorquin: I was getting some dependency errors in my local installation when I tried pinning this to an exact version.

The problem is that is that the pip dependency resolver (or what it looks like it?) requires a different versions of `sonpy`  for python 3.7 and 3.9. 

I did not dig too deep into it but it might come from the flexibility allowed in `scipy` and `h5py`.

I could dig deeper into this if you think is worth it.

bendichter: oh yes I remember that being an issue. OK this is good

CodyCBakerPhD: I'm sure Ben remembers now but for for @h-mayorquin: `pip install sonpy` installs a different version depending on platform and Python version, and most of the cross-platform and cross-Python versions are not intercompatible. Attempts to pin to an exact version inevitably break the non-compatible regions of the test matrix, so we are forced to leave this as loose (at least on the upper bound). So this is fine.

setup.py
CodyCBakerPhD: `requirements-full.txt` is the list of legit external dependencies a user might need for utilizing the `DataInterfaces`. `pytest` is only needed for developers looking to run the full testing suite. `pytest-cov` is only for the CI to compile reports for upload.

So I'd prefer `pytest` and `pytest-cov` to be developer-specific installations outside requirement files. Or a separate `requirements-dev.txt` but that seems excessive.

requirements-full.txt
h-mayorquin: Yes. I realized when I was testing this. Thanks for the explanation!

h-mayorquin:
@CodyCBakerPhD 
> 1. If you are inside a folder containing the nwb-conversion-tools project, I assume `pip install nwb-conversion-tools[test_full]`  (no -e flag) knows to call setup on THAT project and NOT the official PyPI version?

I just tested this and it calls the PyPI version by default if you happen to provide the package name. In order for pip to call the local version it seems that the dot notation is required: `pip install .[test_full]`

> 2. How does this work with PyPI releases? If you are NOT inside a folder containing the nwb-conversion-tools project and you try that command, I assume it works analogously without disrupting a normal minimal install of `pip install nwb-conversion-tools`? Are there any examples of PyPI released projects that utilize this so we can double check that?

I think so. In my understanding this feature is designed to call the normal dependencies plus the ones defined as extra. The python-neo project uses extra to add the `nixio` and `tiffio` dependencies:

https://neo.readthedocs.io/en/stable/install.html

CodyCBakerPhD:
@h-mayorquin Cool, thanks for looking into it!

Looking forward, would be good to keep in mind we might eventually utilize this for intention-specific installs depending on modality and interfaces, such as `pip install nwb-conversion-tools[ecephys, movies]` or more specific like `pip install nwb-conversion-tools[intan, phy, movies]`. This could help us keep the base installation even lazier (I notice it takes a while actually to do a full bottom-up fresh install of `pynwb` + `spikeinterface`).

CodyCBakerPhD (approved):

PR #411, opened 2022-03-01
Reorganize tools

### LINKED ISSUE #400: v0.11 util vs. tool nesting
It's been requested for a long time by @bendichter to use a more proper `utils` designation applying only to functions that contain no external dependencies beyond the standard python library. Anything that depends on stuff like `pynwb` or `spikeinterface` should really, then, be classified as a `tools` function instead. Below are some proposals for how this ought to be nested, and I wanted to get everyone's feedback on which looks nicest.

@alejoe91 has also requested that commonly used tool functions for SI be exposed at a higher level (ideally `from nwb_conversion_tools import write_recording`), and while implementing this is as easy as updating the `__init__.py` for the package independent of the `utils`/`tools` discrepancy, I have some ambiguity concerns with respect to the naming schemes there, which I highlight at the end of each proposal.

(a)

```
nwb_conversion_tools
  - data_interfaces
    -- All the various interfaces we use
  - schemas
    -- All the various JSON schemas we use
  - utils
    -- schemas
      --- Basically, all the internal schematic retrieval utilities we use like get_schema_from_hdmf_class
    -- dict
      --- Dictionary-specific utils, of which we have several but the most important being dict_deep_update
  - write_tools
    -- nwbfile.py
    -- yaml_conversion_specification.py
    -- spikeinterface
      --- write functions as well as generic iterator
    -- roiextractors
      --- write functions as well as generic iterator (TODO)
    -- neo
      --- the ABF reader and other future icephys stuff
```

@alejoe91 - For unexposed imports, this would look like `from nwb_conversion_tools.write_tools.spikeinterface import write_recording`. My concern here is that, when exposed at the highest level, `from nwb_conversion_tools import write_recording` doesn't indicate any specificity to SpikeInterface; so when we eventually add neo support or anything similar, it could be ambiguous what kind of 'recording' we are writing; whereas when unexposed, the way it is imported clearly indicates the type of recording being references. Re-naming the function to `write_spike_interface_recording` could otherwise resolve this ambiguity, but then a direct unexposed import of `from nwb_conversion_tools.write_tools.spikeinterface import write_spike_interface_recording` seems excessively verbose.

(b)

```
nwb_conversion_tools
  - data_interfaces
    -- ecephys
      --- baserecordinginterface.py, ...
      --- write_tools
        ---- SpikeInterface write functions as well as generic iterator. The idea here is that these tools
        ---- are so specific to these interfaces, they could live next to each other.
    -- ophys
      --- baseimaginginterface.py, ...
      --- write_tools
        ---- roiextractor write functions as well as generic iterator. The idea here is that these tools
        ---- are so specific to these interfaces, they could live next to each other.
    -- icephys
      --- baseicephysinterface.py
      --- write_tools
        ---- the ABF reader and other future icephys/neo stuff. The idea here is that these tools
        ---- are so specific to these interfaces, they could live next to each other.
  - schemas
    -- All the various JSON schemas we use
  - utils
    -- schemas
      --- Basically, all the internal schematic retrieval utilities we use like get_schema_from_hdmf_class
    -- dict
      --- Dictionary-specific utils, of which we have several but the most important being dict_deep_update
  - tools
    -- nwbfile.py
    -- yaml_conversion_specification.py
```

@alejoe91 - unexposed:  `from nwb_conversion_tools.data_interfaces.ecephys.write_tools.spikeinterface import write_recording` I'll admit, that is so utterly nested it just needs to be exposed regardless. In this case I'd just insert `_spike_interface_` into all the function names to make it obvious that's the type of recording being referred to.

(c)

```
nwb_conversion_tools
  - data_interfaces
    -- All the various interfaces we use
  - schemas
    -- All the various JSON schemas we use
  - utils
    -- schemas
      --- Basically, all the internal schematic retrieval utilities we use like get_schema_from_hdmf_class
    -- dict
      --- Dictionary-specific utils, of which we have several but the most important being dict_deep_update
  - tools
    -- nwbfile.py
    -- yaml_conversion_specification.py
    -- writing
      --- spikeinterface
        ---- write functions as well as generic iterator
      --- roiextractors
        ---- write functions as well as generic iterator (TODO)
      --- neo
        ---- the ABF reader and other future icephys stuff
```

@alejoe91 - similar to (b), nesting becomes undeniably complicated.

Open to all other suggestions as well; trying to make the structure and clean, organized, and intuitive as possible.

CodyCBakerPhD:
fix #400 

## Summary of Changes

Big one-time reorganization of import and file structure for `utils` and `tools`.

### Utils

`utils` functions are now split into 3 file categories based on application: 
1. `dict` (deep update and its related functions, loading proper dict structures from YAML/JSON), 
2. `schema` (little schema helpers and the complex functions for schema inference and retrieval from method signatures and HDMF structures), and 
3. `types` (standalone file containing our custom defined `TypeVar`s and other central references).

All `utils` functions and classes are now imported as: `from nwb_conversion_tools.utils import deep_dict_update, FilePathType`. The new file structure is purely for ease of organization and external introspection or navigation.

The NWBInspector utils are now used for the general logic for checking regularity in timestamps.

### Tools

In accordance with option (a) from the corresponding issue, each externally supported package now has its own folder within the `tools` directory.

References to `spike_interface` are now renamed to `spikeinterface` to better match the official package and for consistency with `roiextractors`.

Each of these are now importable from the following alias method:

```python
from nwb_conversion_tools import spikeinterface, roiextractors
spikeinterface.write_recording?
roiextractors.write_imaging?
```

or

```python
import nwb_conversion_tools
nwb_conversion_tools.spikeinterface.write_recording?
nwb_conversion_tools.roiextractors.write_imaging?
```

which is syntactically more analogous to the previous static method calls, but instead of belonging to a class they belong to a module. The requested alias method of `from nwb_conversion_tools.spikeinterface import write_recording` proved quite difficult and hacky to implement, so this is the best we could do.

@bendichter Possibly relevant to your DataLad work: another change is to simplify a filename from what was previously `nwb_conversion_tools.utils.nwbfile_tools` to `nwb_conversion_tools.tools.nwb_helpers`. This is the submodule that contains NWBFile related things like `get_module`. The name `nwb_helpers` was chosen to avoid conflicts with the common names for instances of NWBFile objects (typically `nwbfile` and `nwb`).

The YAML conversion specification feature is now directly importable from the highest level: `from nwb_conversion_tools import run_conversion_from_yaml`. It now has its own folder as well in anticipation of many future features such globbing and CLI.

## Checklist


bendichter:
I worry about a module called nwbfile, because we often use `nwbfile` as the variable name for an `NWBFile` object. I also think we might have NWB tools that don't operate on `NWBFile` objects per se, so it might be better to just call this module `nwb.py`.

CodyCBakerPhD:
> I worry about a module called nwbfile, because we often use nwbfile as the variable name for an NWBFile object. I also think we might have NWB tools that don't operate on NWBFile objects per se, so it might be better to just call this module nwb.py.

That's a fair point... but I've also seen other people use `nwb` as the name of their `NWBFile` object, lol.

Hmmm.... maybe a slightly redundant name `nwb_conversion_tools.tools.nwb_tools` then?

bendichter:
that has "tool" in it 3 times..

CodyCBakerPhD:
> that has "tools" in it 3 times..

'slightly'™ redundant 😄

The issue at hand is this is a module that **can** be imported as `from nwb_conversion_tools.tools import nwbfile` (or `nwb`) and you would then have an object in the environment that could have the same name as an actual instance of an NWBFile.

Otherwise if anyone imports specific functions from it like `from nwb_conversion_tools.tools.nwbfile import get_module` I think it's clear enough that it's a module, not an instance of an NWBFile.

Any other ideas? `nwb_helpers`? `nwb_funcs`? `nwb_extra`?

bendichter:
I like nwb_helpers

CodyCBakerPhD:
`nwb_helpers` it is - see anything else?

bendichter:
why change "json_schema" to "schema"? We are specifically using the [json schema](https://json-schema.org/) protocol

CodyCBakerPhD:
> why change "json_schema" to "schema"? We are specifically using the [json schema](https://json-schema.org/) protocol

Thought was so that it then matches the name of the outermost level `schemas` (the folder where we keep hard-coded JSON schemas centralized).

Would you prefer these both to be changed to

`nwb_conversion_tools/schemas/` -> `nwb_conversion_tools/json_schemas/`

and

`nwb_conversion_tools/utils/schema.py` -> `nwb_conversion_tools/utils/json_schema.py`

to further emphasize the fact they are explicitly following the JSON protocol?

bendichter:
I don't see why they need to match. In fact I think it's better that they don't so they don't get confused with each other.

CodyCBakerPhD:
Restored util submodule to previous `json_schema`.

In the release notes for this major version increase, I will include the following simple instructions for updating any scripts that are broken in going from `v0.9.9` to `>=v0.11.x`.

`from nwb_conversion_tools.utils.json_schema import xyz` becomes `from nwb_conversion_tools.utils import xyz`
`from nwb_conversion_tools.utils.nwbfile_tools import xyz` becomes `from nwb_conversion_tools.tools.nwb_helpers`
`from nwb_conversion_tools.utils.spike_interface import xyz` becomes `from nwb_conversion_tools import spikeinterface` + `spikeinterface.xyz`

nwb_conversion_tools/basedatainterface.py
```diff
@@ -1,7 +1,7 @@
 """Authors: Cody Baker and Ben Dichter."""
 from abc import abstractmethod, ABC
 
-from .utils.json_schema import get_base_schema, get_schema_from_method_signature
+from .utils import get_base_schema, get_schema_from_method_signature
```
bendichter, line 4: What is the purpose of this change?

CodyCBakerPhD: Much easier to import utils when you don't have to think about which sub-module they belong to. Also prevents future additions/changes from breaking anything; tbh we probably always should have done it that way.

bendichter (approved):
This looks good to me

CodyCBakerPhD:
> This looks good to me

Just let me know if you want me to go into existing code somewhere that requires these updates. Happy to save other people (and you) the pain if you just point me to the place that needs it

I'll give the other PRs a look through to see if anything else can get merged first to reduce conflicts

PR #413, opened 2022-03-04
Adding parameterized as a dependency to the setup

h-mayorquin:
## Motivation

Small PR requested by @CodyCBakerPhD  plus some minor modification to variable naming in the `setup.py` for readability


CodyCBakerPhD:
@h-mayorquin Thanks, next step is to adjust the workflows to call it this way. We can do that in our meeting if you want.

requirements-minimal.txt
```diff
@@ -1,18 +1,17 @@
...
+dandi>=0.34.1
+nwbinspector>=0.2.3
 spikeextractors>=0.9.9
 spikeinterface>=0.93.0
 neo>=0.9.0
 roiextractors>=0.4.0
-dandi>=0.34.1
-nwbinspector>=0.2.3
+lxml>=4.6.5
+scipy>=1.4.1
```
CodyCBakerPhD, line 17: Eventually, goal is to have these removed as well for super minimal install.

Justification being, if someone just wants to use nwbct for ophys, why require them to install spikeinterface and neo as well?

CodyCBakerPhD (approved):

bendichter:
@h-mayorquin now that `parameterized` is an official testing requirement, could you remove the soft import logic [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/612b52cb151c909a183be4c13f7cebd1dc8f1f83/tests/test_on_data/test_gin_ecephys.py#L37-L42)?

PR #414, opened 2022-03-04
Interface for DLC data

Saksham20:
## Motivation

dlc2nwb module converts DLC output to nwb format. This PR leverages that and packages that as a behavior data interface. 

We cannot simply use the `convert_h5_to_nwb` offered by the `dlc2nwb` module since datainterfaces' `run_conversion()` method only needs to add relevant data to an existing nwbfile and not write one on disk (their convert method has no option to not write on disk). Thus much of the code from the library is ported into this `run_conversion()` method. 

## Checklist


Saksham20:
TODO:

- [ ] Tests

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        Conversion from DLC output files to nwb. Derived from dlc2nwb library.
+
+        Parameters
+        ----------
+        nwbfile: pynwb.NWBFile
+        metadata: dict
+        individual_name : str
+            Name of the subject (whose pose is predicted) for single-animal DLC project.
+            For multi-animal projects, the names from the DLC project will be used directly.
```
CodyCBakerPhD, line 75: How do the multiple subjects get added to the NWBFile? A single NWBFile generally only supports data for one subject at a time

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,143 @@
...
+        scorer = "DLC" + os.path.splitext(scorer)[0]
+        return dict(vidname=vidname, scorer=scorer)
+
+    def get_metadata(self):
+        metadata = dict(
+            NWBFile=dict(
+                session_description=self._config_file["Task"],
+                experimenter=self._config_file["scorer"],
+                identifier=self._derived_metadata["scorer"],
+                session_start_time=datetime.datetime.now(datetime.timezone.utc),
```
bendichter, line 60: please don't add `session_start_time` like this anymore. The correct session_start_time needs to be handled by the user

```suggestion
```

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,143 @@
...
+        dlc_file_path = self.source_data["dlc_file_path"]
+        vidname, scorer = os.path.split(dlc_file_path)[-1].split("DLC")
+        scorer = "DLC" + os.path.splitext(scorer)[0]
+        return dict(vidname=vidname, scorer=scorer)
+
+    def get_metadata(self):
+        metadata = dict(
+            NWBFile=dict(
+                session_description=self._config_file["Task"],
+                experimenter=self._config_file["scorer"],
```
CodyCBakerPhD, line 58: I also feel like this, out of all the interfaces we support, ought to be the first one where we actually require (and can hopefully extract) `Subject` info, since if someone is using DLC at all it would imply there is a Subject being tracked.

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,143 @@
...
+                original_videos=[video[0]],
+                dimensions=[list(map(int, video[1].split(",")))[1::2]],
+                scorer=self._derived_metadata["scorer"],
+                source_software="DeepLabCut",
+                source_software_version=__version__,
+                nodes=[pes.name for pes in pose_estimation_series],
+                edges=paf_graph,
+            )
+
+            get_module(nwbfile=nwbfile, name="behavior", description="processed behavioral data").add(pe)
```
bendichter, line 143: It looks like what you've done here is copy/paste [a large section of code in DLC2NWB](https://github.com/DeepLabCut/DLC2NWB/blob/5c704446a41f7fd6dfe9bcf7ceeedde69ec03717/dlc2nwb/utils.py#L89-L175). Please instead make the necessary modifications to DLC2NWB in a PR and import those functions here.

CodyCBakerPhD:
@Saksham20 Also, when you write tests on data for this (which are stored on the actual `dlc2nwb` repo: https://github.com/DeepLabCut/DLC2NWB/tree/main/examples) if you can set them up with the config options like the ecephys/ophys tests (example: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/tests/test_on_data/test_gin_ecephys.py#L47-L62) so we can view and compare examples of what these NWBFiles look like, that would be a huge plus.

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
CodyCBakerPhD: Agreed, a very simple issue+PR would simply request they allow an existing NWBFile object to be passed in, and if so prevent the lines https://github.com/DeepLabCut/DLC2NWB/blob/main/dlc2nwb/utils.py#L157-L162 and https://github.com/DeepLabCut/DLC2NWB/blob/main/dlc2nwb/utils.py#L169-L172 from getting called, instead returning the `nwbfile` object post-modification.

Would make the extension even better as well that way, so I can't see them objecting.

Saksham20: yeah makes sense, I've created a PR https://github.com/DeepLabCut/DLC2NWB/pull/10

CodyCBakerPhD:
Something I just noticed while reading their main DLC code: https://github.com/DeepLabCut/DeepLabCut/blob/master/deeplabcut/utils/auxfun_videos.py#L127

Does OpenCV not return color channels as RGB? If not can you fix that in our movie utils and add it to the documentation for the interface and utils @Saksham20?

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
CodyCBakerPhD: In the meantime, feel free to pin to that dev branch and update the code here correspondingly so we can see what it would look like

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,54 @@
...
+        Conversion from DLC output files to nwb. Derived from dlc2nwb library.
+
+        Parameters
+        ----------
+        nwbfile: pynwb.NWBFile
+        metadata: dict
+
+        """
+        self._config_file["Task"] = metadata["NWBFile"]["session_description"]
+        self._config_file["scorer"] = metadata["NWBFile"]["experimenter"]
```
bendichter, line 52: I don't know if this is going to work. Why are these needed in `write_subecjt_to_nwb`? These values may be overwritten by another interface.

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,54 @@
...
+        if "DLC" not in dlc_file_path or not dlc_file_path.endswith(".h5"):
+            raise IOError("The file passed in is not a DeepLabCut h5 data file.")
+        assert HAVE_DLC2NWB, "to use this datainterface: 'pip install dlc2nwb'"
+        self._config_file = auxiliaryfunctions.read_config(config_file_path)
+        super().__init__(dlc_file_path=dlc_file_path, config_file_path=config_file_path)
+
+    def get_metadata(self):
+        metadata = dict(
+            NWBFile=dict(session_description=self._config_file["Task"], experimenter=[self._config_file["scorer"]]),
+            Subject=dict(subject_id="ind1"),
```
bendichter, line 37: where are you gettinb "ind1" from?

bendichter: If we use this, every subject for a dandiset will have the same name, "ind1"

Saksham20: currently, its expected of the user to change the metadata(add `session_start_time`, and modify `subject_id`).
But yes, I agree that if we have the `subject_name` as an optional argument to the `init`, it prompts the user to care to specify the subject name. 
As for the default `subject_id`, I would prefer it to not be a constant ("ind1") but construct it from the video name: `subject_id=f"subject_{video_name}"`. Since every video file will be unique to a subject, if the user does not specify the subject name for a bunch of nwb files each will have a unique subject name.

nwb_conversion_tools/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
Saksham20: yes thanks for catching that, its unnecessary, as its only used in creating the nwbfile which happens by the `NWBConverter` using the info in `get_metadata()`. Now it looks like : https://github.com/catalystneuro/nwb-conversion-tools/pull/414/files#diff-0b2add275d370a025b0cc183dd63f4fb168efd54f56fb58fca1ecb76901c8f57R56

Saksham20:
@CodyCBakerPhD why are we not doing full testing ( GIN data conversions) for Mac and windows.
These lines in the workflow are commented out https://github.com/catalystneuro/nwb-conversion-tools/blob/5ca7b852366e23f3a3c110df8c3ab9a61f2f9a3f/.github/workflows/testing-mac.yml#L37

bendichter:
There were problems with installing datalad if I recall correctly

CodyCBakerPhD:
> @CodyCBakerPhD why are we not doing full testing ( GIN data conversions) for Mac and windows. These lines in the workflow are commented out
> 
> https://github.com/catalystneuro/nwb-conversion-tools/blob/5ca7b852366e23f3a3c110df8c3ab9a61f2f9a3f/.github/workflows/testing-mac.yml#L37

(a) Windows was never able to unpack caches or artifacts created from Linux, and never been able to get datalad to function on Windows

(b) If the Mac CI finds the cache (not a guarantee until they improve their interface, which is something they've discussed adding with the community) it will run full tests; however, if it does NOT find the cache, it would need to force an install of the data, which we've never gotten working through datalad on Mac either. If you don't force the install, it will save an empty cache with that name which will screw up the Linux side

If you know of another way to download the data efficiently without datalad within the workflow environments, especially a cross-platform way, I'd welcome it (in a separate PR)

CodyCBakerPhD:
Marking as draft until upstream PR is merged to their main.

h-mayorquin:
Now that https://github.com/catalystneuro/neuroconv/pull/24 is merged I guess we can finally close this one.

PR #415, opened 2022-03-04
Handle non int channel ids in spikeinterface

### LINKED ISSUE #403: Support non-int channel_ids

h-mayorquin:
## Motivation

This is a pull request fulfilling the same role as #404 and should close fix #403.

I opened a second PR because the organization blocked the merge in the other one and I thought it would be cleaner to open a new one feeding in the other PR discussion.

I started by adding some tests and now I will work on the code for passing them. It would be great if either of you @alejoe91  and @CodyCBakerPhD were to have other testing scenarios in mind.

## Checklist


h-mayorquin:
The tests now are passing and the draft is ready for review.

Two important issues:
1. Right now I am creating this extra property in the electrodes table call `channel_name` that holds the string reference for the channels. Maybe there is a better name for this? `channel_id_reference`, `channel_identifier` or maybe we can even use the `reference` property that is [documented](https://pynwb.readthedocs.io/en/stable/_modules/pynwb/file.html#NWBFile.add_electrode_column) in `pynwb`. See also the latest [discussion](https://nwb-users.slack.com/archives/C5XKC14L9/p1646424876546959) in the slack channel about this. 
2. To pass the tests on data, it was necessary to add a mechanism to reproduce the old behavior of assigning `recorder.channel_ids` to `nwb_file.electrode.table.ids` when the channel_ids are integers. The reason for this is that some of our tests like `test_convert_recording_extractor_to_nwb` use a `NwBRecordingExtractor` which extracts the `nwb_file.electrodes.ids` as channel ids and then the comparison (`check_recordings_equal`) fails.  A possible workaround would be to extract the `channel_name` from the `nwb_file.electrodes` table as channels_ids if the column exist but this is a responsibility of `spikeextractors` and `spikeinterface` Maybe there is a middle way?

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -334,6 +329,14 @@ def add_electrodes(
                 index = isinstance(data[0], (list, np.ndarray))
                 elec_columns[prop].update(description=prop, data=data, index=index)
 
+    # Add the property of channel name
+    channel_name_array = checked_recording.get_channel_ids()
+    elec_columns["channel_name"].update(
+        description="a string named referencefor the channel",
```
CodyCBakerPhD, line 335:
```suggestion
        description="Text-based reference of the electrode ID.",
```

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -334,6 +329,14 @@ def add_electrodes(
...
+        data=channel_name_array,
+        index=False,
+    )
```
CodyCBakerPhD, line 338: One thing to check here would be that `channel_name` is unique within the table. The `id` field does this naturally via PyNWB using an `enforce_unique_values` type of argument which complains if you try to add an `id` that already exists. This will also keep the mapping of `channel_name` -> `id` one-to-one.

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -359,31 +362,30 @@ def add_electrodes(
...
     for name in elec_columns_append:
         _ = elec_columns.pop(name)
 
-    for j, channel_id in enumerate(checked_recording.get_channel_ids()):
-        if channel_id not in nwb_elec_ids:
+    channel_name_array = checked_recording.get_channel_ids()
+    # If the channel ids are integer keep the old behavior of asigning electrodes.ids equal to channel_ids
+    if np.issubdtype(channel_name_array.dtype, np.integer):
+        elec_columns["id"].update(data=channel_name_array)
+
```
CodyCBakerPhD, line 374: I would have imagined a coerced RecordingExtractor -> BaseRecording object would have its IDs upcast to strings. Is this not the case? If so, I do think we should probably adjust that in the `OldtoNew` functions. @alejoe91

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -400,17 +402,31 @@ def add_electrodes(
...
 
             nwbfile.add_electrode(**electrode_kwargs)
 
     # add columns for existing electrodes:
+    channels_not_available_in_recorder = set(channel_names_in_electrodes_table).difference(set(channel_name_array))
     for col_name, cols_args in elec_columns_append.items():
+        data = cols_args["data"]
+        samp_data = data[0]
+        data_type_found = [proptype for proptype in property_default_types if isinstance(samp_data, proptype)][0]
+
```
CodyCBakerPhD, line 419:
```suggestion
        for proptype in property_default_types:
            if isinstance(data[0], proptype):
                data_type_found = proptype
                break
```
Good opportunity here to do what I suggested in previous PR.

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -400,17 +402,31 @@ def add_electrodes(
...
+        default_value = property_default_types[data_type_found]
```
CodyCBakerPhD, line 420:
```suggestion
        for proptype in property_default_types:
            if isinstance(data[0], proptype):
                default_value = property_default_types[proptype]
                break
```
Good opportunity here to do what I suggested in previous PR.

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -400,17 +402,31 @@ def add_electrodes(
...
+        column_size = len(nwbfile.electrodes.id[:])
+        extended_data = [default_value for _ in range(column_size)]
+        extended_data[len(channels_not_available_in_recorder) :] = data
+        if data_type_found == Real:
+            extended_data = [float(_) for _ in extended_data]
```
CodyCBakerPhD, line 425: It would be simpler and cleaner here to define `extended_data` as a numpy array and use `.astype(float)` to re-cast the entire array. Can  this be done or does it interfere with other functionality somewhere?

tests/test_internals/test_si.py
```diff
@@ -576,5 +579,74 @@ def test_group_set_custom_description(self):
...
+        # Re-write
+        write_recording(
+            recording=self.RX_non_int_channels2, nwbfile=self.nwbfile1, metadata=self.metadata_list[1], es_key="es2"
+        )
+
+        with NWBHDF5IO(str(self.path1), "w") as io:
+            io.write(self.nwbfile1)
+        with NWBHDF5IO(str(self.path1), "r") as io:
+            nwb = io.read()
+
```
CodyCBakerPhD, line 634: Couple minor comments about the tests...

(a) I love the use of `TestCase`, it will eventually make this file much easier to look at.

(b) Another habit we really should get into here is using the simplest types and functions possible for evaluating the functionality of targetted changes to specific functions. That is, many of these tests could be designed to operate using only `add_electrodes` and make their assertions directly on the electrode of the `nwbfile` object in memory (no need for explicit `NWBHDF5IO`.

tests/test_internals/test_si.py
```diff
@@ -576,5 +579,74 @@ def test_group_set_custom_description(self):
...
+            # Test extension of the same property for the new channels
+            for id in [0, 1, 2, 3]:
+                assert nwb.electrodes["property"][id] == "value_before_rewrite"
+            for id in [4, 5]:
+                assert nwb.electrodes["property"][id] == "value_after_rewrite"
+
+            # Test addition of new property for the second recording
+            for id in [2, 3, 4, 5]:
+                assert nwb.electrodes["property2"][id] == "second_recoder_property"
```
CodyCBakerPhD, line 648: (c) Yet another  habit to get into would be making a separate test function for each case and naming the functions appropriately to the purpose of the test. So each of these parts indicated by comments could be their own tests, and the common parts of the `setUp` could be propagated into the actual class setup.

tests/test_internals/test_si.py
```diff
@@ -576,5 +579,74 @@ def test_group_set_custom_description(self):
...
+        self.path1 = self.test_dir + "/test_electrodes1.nwb"
+
+        self.nwbfile1 = NWBFile("session_description1", "file_id1", datetime.now())
+
+        self.RX1 = generate_recording(num_channels=4, durations=[3])
+        self.metadata_list = [dict(Ecephys={i: dict(name=i, description="desc")}) for i in ["es1", "es2"]]
+
+    def tearDown(self):
+        del self.RX1
+        shutil.rmtree(self.test_dir)
```
CodyCBakerPhD, line 595: One thing to learn about `TestCase` is the difference between available setup methods. `setUp` and `tearDown` run each time a test function in the class is run. Using

```python
    @classmethod
    def setUpClass(cls):
        ...

    @classmethod
    def tearDownClass(cls):
        ...
```

instead only run once total for the testing class. Try to think about what setup operations can be called fewer times using this distinction vs. what 'new'/'fresh' objects you need to have going into each test for custom modification leading to assertion.

tests/test_internals/test_si.py
```diff
@@ -576,5 +579,74 @@ def test_group_set_custom_description(self):
...
+
+        with NWBHDF5IO(str(self.path1), "w") as io:
+            io.write(self.nwbfile1)
+        with NWBHDF5IO(str(self.path1), "r") as io:
+            nwb = io.read()
+
+            # First we test the channel names are assigned in the written order
+            expected_channel_names = ["a", "b", "c", "d", "e", "f"]
+            for id in nwb.electrodes.id[:]:
+                assert nwb.electrodes["channel_name"][id] == expected_channel_names[id]
```
CodyCBakerPhD, line 638:
```suggestion
                assert self.assertEqual(nwb.electrodes["channel_name"][id], expected_channel_names[id])
```

When using `TestCase` it's probably better to use the built-in assertion methods whenever possible.

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: What should we do if the strings channel_ids are not unique? 
The current implementation that I have here will always produce unique electrode.table ids but not unique channel_name. Should we throw an error?

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: This is not the case and if we do this some of our data tests will fail with the current implementation that I have on this PR.

The reason being is that when the channel_ids are strings I assigned nwb_file.elctrode.ids to be `range(number_of_electrodes)` but one of our test in data (I don't remember which one) has channel_ids that are like this [1, 2, 3, 4, 125, 126] (integers). So if you cast to string the current implementation assigns nwb_file.electrode.ids as [0, 1, 2, 3, 4, 5, 6] and the comparison of channel_ids of the original recorder to the channel_ids of the nwb_recorder fails. 

One way around would be make to use a function like it `isdigit()` to with the strings for the cases where the strings channel_ids can be easily assigned as the nwb_file.electrode.ids but then I am not sure we are gaining something. What do you think?

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: I still don't feel completely comfortable with this because of the number of for loops. I am trying to reduce https://en.wikipedia.org/wiki/Cyclomatic_complexity of this function as it really bogged me down the first time I attempted to understand it and I want this to be easier for a future developer. 

Check the accepeted answer here that I will implement in a refactor:
https://stackoverflow.com/questions/9868653/find-first-sequence-item-that-matches-a-criterion
If you still don't like it we go with the loop:

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Done.

tests/test_internals/test_si.py
h-mayorquin: This is good, I was not aware of this.

tests/test_internals/test_si.py
h-mayorquin: I agree with the spirit but in this case we are not only testing `add_electrodes` but also `add_electrical_series` were some changes also had to be made.

 I think it is better to keep the test this general test until we have more specific tests for the other parts of  `writing_recording` as future changes in other part of `writing_recording` might try to rely or assume that `nwb_file.electrodes.id` is equivalent to `channel_ids` and this test will fail alerting them that they should not. Now that said, I would agree with you that the solution for that is to write a specific test for all the different parts of `write_recording` so we can decouple testing but right now I don't think that I have enough knowledge to do that. What do you think?

h-mayorquin: With respect to using `NWBHDF5IO`, that is, testing in memory vs read from disk, I think it makes sense. Thinking about this however I am wondering, why do we this in other tests? which kind of operations would grant a test with writing and reading to/from disk? a quick glance at the other test concerning the electrode table do not reveal evident pattern to me.

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
alejoe91: The `channel_ids` should be already enforced to be unique in spikeinterface

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
alejoe91: @CodyCBakerPhD We don't enforce any specific dtype on the SI side (and actually most of the extractors still use int). So we have to deal with both str and ints here

tests/test_internals/test_si.py
h-mayorquin: I gave this a thought and I am not sure what do we gain. I think that all these test make sense together as they test writing non int channel ids to the electrode table. 

If I were to partition there would be three functions instead of one with the following names:
* test_non_ints_as_channel_ids_in_electrode_table_channel_name.
* test_non_ints_as_channel_ids_in_electrode_table_property_extension.
* test_non_ints_as_channel_ids_in_electrode_table_new_property_addition.

This increases a little bit the boilerplate and the future user of the test needs to jump back and forth between the actual test and the `setUp` definition which is not too bad here as both things are small but what do we gain? this is not clear to me.

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Is this new, the following code does not throw an error?
```python
from spikeinterface.core.testing_tools import generate_recording
rec = generate_recording(num_channels=4, durations=[10])
channel_ids = rec.get_channel_ids()
non_int_channel_ids = ['a', 'b', 'c', 'a']
rec_non_int_channels = rec.channel_slice(channel_ids=channel_ids, renamed_channel_ids=non_int_channel_ids)
rec_non_int_channels.get_channel_ids()

output:
array(['a', 'b', 'c', 'a'], dtype='<U1')
```

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Might as well force the upcasting ourselves then. (a) always include `channel_name`, (b) always have it be `str(channel_id)`

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Great job, a generator approach is even better than what I suggested!

tests/test_internals/test_si.py
CodyCBakerPhD: >which kind of operations would grant a test with writing and reading to/from disk?

Tests involving data attributes; one of the biggest changes whenever you do that is the file on disk has everything written to `h5py.Dataset` objects so that's when compression/chunking actually becomes callable if you had previously specified the `data` as a list or `array`.

tests/test_internals/test_si.py
h-mayorquin: This makes sense.

tests/test_internals/test_si.py
```diff
@@ -576,5 +585,68 @@ def test_group_set_custom_description(self):
...
+        property_value = ["second_recorder_property" for _ in range(num_channels)]
+        cls.RX_non_int_channels2.set_property("property2", values=property_value)
+
+        # Write the two electrodes to a file
+        add_electrodes(recording=cls.RX_non_int_channels1, nwbfile=cls.nwbfile1)
+        add_electrodes(recording=cls.RX_non_int_channels2, nwbfile=cls.nwbfile1)
+
+    @classmethod
+    def tearDownClass(cls):
+        del cls.RX1, cls.RX_non_int_channels1, cls.RX_non_int_channels2
```
CodyCBakerPhD, line 630: Can you look into if this manual deletion of variables is necessary? I'm pretty sure it should happen automatically, whereas the temporary folder directory cleanup is something we definitely are responsible for.

h-mayorquin: It is actually not necessary, I removed it.

CodyCBakerPhD:
@alejoe91 @h-mayorquin Sorry, so what was the status on enforcing uniqueness of `channel_id`'s in the new spikeinterface? And our reliability on it for the `channel_name` here?

h-mayorquin:
> @alejoe91 @h-mayorquin Sorry, so what was the status on enforcing uniqueness of `channel_id`'s in the new spikeinterface? And our reliability on it for the `channel_name` here?

So according to @samuelgarcia it should be enforced everywhere and the toy example above is a bug. So we have uniquness.

Of your two suggestions  above( (a) always include channel_name, (b) always have it be str(channel_id)) the only missing is b.

Shall I implement that one?
Is there anything else missing besides merging the test improvements?

h-mayorquin:
> > @alejoe91 @h-mayorquin Sorry, so what was the status on enforcing uniqueness of `channel_id`'s in the new spikeinterface? And our reliability on it for the `channel_name` here?
> 
> So according to @samuelgarcia it should be enforced everywhere and the toy example above is a bug. So we have uniquness.
> 
> Of your two suggestions above( (a) always include channel_name, (b) always have it be str(channel_id)) the only missing is b.
> 
> Shall I implement that one? Is there anything else missing besides merging the test improvements?

h-mayorquin:
> > Shall I implement that one? Is there anything else missing besides merging the test improvements?
> 
> After (b) it should be good. Main reason we want that is so that if we write to recordings, first one being all-int, second one all-strings, we want the `channel_name` column to be present for our reference purposes.

@CodyCBakerPhD 
Channel name was already present in the version of the last commit but it was not always a string. The new commit does ensures that channel_name is always a string. I added another test for this behavior.

CodyCBakerPhD:
> > > Shall I implement that one? Is there anything else missing besides merging the test improvements?
> > 
> > 
> > After (b) it should be good. Main reason we want that is so that if we write to recordings, first one being all-int, second one all-strings, we want the `channel_name` column to be present for our reference purposes.
> 
> @CodyCBakerPhD Channel name was already present in the version of the last commit but it was not always a string. The new commit does ensures that channel_name is always a string. I added another test for this behavior.

Do you think we should we think about what would happen if someone manually added their own electrodes prior to using `add_electrodes(recording=recording, ...) with a `recording` object? Can you add a test for that and see if the expected behavior is appropriate? I would think in that case it would be that situation where you add the column but fill with defaults, only thing is I do think the default in this case should be `str(previous_electrode_ids)` instead of a blank string or something.

tests/test_internals/test_si.py
```diff
@@ -576,5 +585,115 @@ def test_group_set_custom_description(self):
...
+        # TODO, fill in any other required values to add_electrode
+        self.nwbfile.add_electrode(id=123)
+        self.nwbfile.add_electrode(id=124)
+
+        add_electrodes(recording=self.recording_1, nwbfile=self.nwbfile)
+
+        expected_ids = []
+        expected_names = []
+        self.assertListEqual(list(self.nwbfile.electrodes.id.data), expected_ids)
+        self.assertListEqual(list(self.nwbfile.electrodes["channel_name"].data), expected_names)
```
CodyCBakerPhD, line 666: @h-mayorquin Here is the base use-case logic we discussed.

Also, it's fine to have multiple assertions in a single function if they correspond to different expected outputs, its just best practice to keep the number of such assertions relatively small and/or vectorized as much as possible where appropriate.

tests/test_internals/test_si.py
```diff
@@ -576,5 +585,173 @@ def test_group_set_custom_description(self):
...
+        values_dic = self.defaults
+
+        # Previous properties
+        values_dic.update(rel_x=0.0, rel_y=0.0, id=123, channel_name=str(123))
+        self.nwbfile.add_electrode(**values_dic)
+
+        values_dic.update(rel_x=0.0, rel_y=0.0, id=124, channel_name=str(124))
+        self.nwbfile.add_electrode(**values_dic)
+
+        values_dic.update(rel_x=0.0, rel_y=0.0, id=None, channel_name="6")  # automatic ID set
```
h-mayorquin, line 700: Note here that the if we have use our `add_electrodes` functio before, the user has to provide the `channel_name` afterwards.

tests/test_internals/test_si.py
```diff
@@ -576,5 +585,173 @@ def test_group_set_custom_description(self):
...
+        expected_ids = [20, 21, 2, 3]
+        expected_names = ["c", "d", "a", "b"]
+        self.assertListEqual(list(self.nwbfile.electrodes.id.data), expected_ids)
+        self.assertListEqual(list(self.nwbfile.electrodes["channel_name"].data), expected_names)
+
+        expected_property_values1 = ["value1_c", "value1_d", "", ""]
+        expected_property_values2 = ["value2_c", "value2_d", "value2_a", "value2_b"]
+        self.assertListEqual(list(self.nwbfile.electrodes["property1"].data), expected_property_values1)
+        self.assertListEqual(list(self.nwbfile.electrodes["property2"].data), expected_property_values2)
+
```
h-mayorquin, line 738: If the user has provided some of the channel_names before the `add_electrodes` function will add the properties by matching with channel_name.

tests/test_internals/test_si.py
```diff
@@ -576,5 +585,173 @@ def test_group_set_custom_description(self):
...
+        values_dic.update(id=0)
+        self.nwbfile.add_electrode(**values_dic)
+
+        values_dic.update(id=1)
+        self.nwbfile.add_electrode(**values_dic)
+
+        self.assertRaises(ValueError, add_electrodes, recording=self.base_recording, nwbfile=self.nwbfile)
+
+
```
h-mayorquin, line 755: This throws a key error if the user tries to use `add_electrodes` with a recording that has channel_ids with ids that were added previously to the table.

h-mayorquin (commented):
@CodyCBakerPhD 
I filled an extended your tests. I had to modify the code but the tests are now passing for the new proposing scenarios in my machine. 

Let me know if there is something missing or to discuss.

tests/test_internals/test_si.py
CodyCBakerPhD: Yep, and it will thankfully warn them they need to do so

tests/test_internals/test_si.py
```diff
@@ -576,5 +585,173 @@ def test_group_set_custom_description(self):
...
+        values_dic = self.defaults
+
+        values_dic.update(id=0)
+        self.nwbfile.add_electrode(**values_dic)
+
+        values_dic.update(id=1)
+        self.nwbfile.add_electrode(**values_dic)
+
+        self.assertRaises(ValueError, add_electrodes, recording=self.base_recording, nwbfile=self.nwbfile)
```
CodyCBakerPhD, line 753:
```suggestion
        self.assertRaisesWith(ValueError, "insert the exact error message here"):
            recording=self.base_recording, nwbfile=self.nwbfile)
```

Minor suggestion for enhanced specificity. Also always prefer to pass the callables after the context to make the usage more apparent.

CodyCBakerPhD: Note that `self.assertRaisesWith` is only present in the `from hdmf.testing import TestCase` library.

CodyCBakerPhD:
@h-mayorquin These look fantastic! Thanks for all the great work on this.

CodyCBakerPhD (approved):

tests/test_internals/test_si.py
h-mayorquin: What about:
```python
        self.assertRaises(ValueError, lambda: add_electrodes(recording=self.base_recording, nwbfile=self.nwbfile))
```
It is expressive without requiring tools outside of unittest. I don't have a specific message designed here, but if you want that it can be done with `assertRaisesRegex`. If you do want that, do you have any suggestions for the message? : O

CodyCBakerPhD: The `hdmf` `TestCase` is a child of `unittest`, and we use it elsewhere in the testing suite so you don't have to worry about that.

Will the text of the message itself change for this specific test function? If not you can simply run it once, grab the message, and fill in the text value. This adds specificity that it isn't some other `ValueError` causing a problem (fairly common error type, but not as common as `AssertionError`.)

PR #417, opened 2022-03-09
Channel name testing suggestions

CodyCBakerPhD:
## Motivation

@h-mayorquin Here are those suggestions I said I'd send your way with regards to improving the design of the new test class for appending electrode tables, and eventually what all those SI tests ought to look like for optimal performance and readability.

Let me know if you have any questions about anything.

## Checklist

### DIFF
--- a/tests/test_internals/test_si.py
+++ b/tests/test_internals/test_si.py
@@ -588,21 +588,21 @@ class TestWriteElectrodes(unittest.TestCase):
 class TestAddElectrodes(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        cls.nwbfile = NWBFile(
+        """Use common recording objects and values."""
+        cls.num_channels = 4
+        cls.base_recording = generate_recording(num_channels=cls.num_channels, durations=[3])
+
+    def setUp(self):
+        """Start with a fresh NWBFile, ElectrodeTable, and remapped BaseRecordings each time."""
+        self.nwbfile = NWBFile(
             session_description="session_description1", identifier="file_id1", session_start_time=datetime.now()
         )
-
-        cls.num_channels = 4
-        base_recording = generate_recording(num_channels=cls.num_channels, durations=[3])
-        channel_ids = base_recording.get_channel_ids()
-
-        non_int_channel_ids_1 = ["a", "b", "c", "d"]
-        non_int_channel_ids_2 = ["c", "d", "e", "f"]
-        cls.recording_1 = base_recording.channel_slice(
-            channel_ids=channel_ids, renamed_channel_ids=non_int_channel_ids_1
+        channel_ids = self.base_recording.get_channel_ids()
+        self.recording_1 = self.base_recording.channel_slice(
+            channel_ids=channel_ids, renamed_channel_ids=["a", "b", "c", "d"]
         )
-        cls.recording_2 = base_recording.channel_slice(
-            channel_ids=channel_ids, renamed_channel_ids=non_int_channel_ids_2
+        self.recording_2 = self.base_recording.channel_slice(
+            channel_ids=channel_ids, renamed_channel_ids=["c", "d", "e", "f"]
         )
 
     def test_channel_names(self):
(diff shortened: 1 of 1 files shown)


tests/test_internals/test_si.py
```diff
@@ -585,66 +585,56 @@ def test_group_set_custom_description(self):
                     assert nwb.electrodes["group"][i].description == "M1 description"
 
 
-class TestSpikeInterfaceRecorders(unittest.TestCase):
+class TestAddElectrodes(unittest.TestCase):
```
h-mayorquin, line 588: I think this is way more descriptive of what the test is doing right now. I was thinking that we would like to add more `spikeinterface` specific functionality that we would like to test in the future and decided to leave it more generally. Maybe this is an instance of early optimization (early generalization?). Note also that there is a test called `TestWriteElectrodes` above leading to some semantic overlap.

tests/test_internals/test_si.py
```diff
@@ -585,66 +585,56 @@ def test_group_set_custom_description(self):
...
+    def test_channel_names(self):
+        """Ensure channel names merge correctly after appending."""
+        add_electrodes(recording=self.recording_1, nwbfile=self.nwbfile)
+        add_electrodes(recording=self.recording_2, nwbfile=self.nwbfile)
 
-        channel_ids = cls.RX1.get_channel_ids()
-        num_channels = cls.RX1.get_num_channels()
-        non_int_channel_ids1 = ["a", "b", "c", "d"]
-        non_int_channel_ids2 = ["c", "d", "e", "f"]
+        self.assertListEqual(list(self.nwbfile.electrodes["channel_name"].data), ["a", "b", "c", "d", "e", "f"])
```
h-mayorquin, line 613: Out of curiosity, what about casting the `["a", "b", ...]` to array and use numpy testing tools?

tests/test_internals/test_si.py
```diff
@@ -585,66 +585,56 @@ def test_group_set_custom_description(self):
...
-        cls.RX_non_int_channels1 = cls.RX1.channel_slice(
-            channel_ids=channel_ids, renamed_channel_ids=non_int_channel_ids1
-        )
-        cls.RX_non_int_channels2 = cls.RX1.channel_slice(
-            channel_ids=channel_ids, renamed_channel_ids=non_int_channel_ids2
-        )
+    def test_common_property_extension(self):
+        """Add a property for a first recording that is then extended by a second recording."""
+        self.recording_1.set_property(key="property", values=["value_1"] * self.num_channels)
+        self.recording_2.set_property(key="property", values=["value_2"] * self.num_channels)
```
h-mayorquin, line 618: My leaning here is to change "property" to "common_property" and "property2" to "added_property" for even more expressiveness. Likewise, I would have the following map for their values:
* "value_1" -> "common_property_value_in_recording_1"
* "value_2" -> "common_property_value_in_recording_2"
* "value_3" -> "value_of_added_property"  

I do think that we are discussing aesthetics at this level, or what do you think? can you cut this one with arguments for one side or the other?

h-mayorquin (commented):
Hey, thanks a lot for doing this. With this I have a better idea of what makes sense or how does a good unittesting looks like. Points to note for me are:
* You re-initialize the nwbfile. I guess this provides a cleaner file while debugging which isolates what is tested in line with the philosophy of unitesting. I guess this is also the reason why you slice the channels again in order to erase their properties? 
* I like the names that you chose for the unitest better. I guess I was hanged up in the fact that we were testing non int channel ids that did not think in the more general properties that we are testing (which implicity test for non int channels) and the resulting names I had were and felt too cumbersome. 
* Documentation of unnitest by giving a brief comment of the intent.

Also, by the end the `TearDown` method for cleaning the directory was not necessary?

h-mayorquin:
Also, I prefer the specificity of `recorder` to the RX convention that we use here and in other places of our library (e.g. as a class attribute of the data interfaces).

tests/test_internals/test_si.py
h-mayorquin: I will take care of this now, then.

CodyCBakerPhD:
> Also, I prefer the specificity of `recorder` to the RX convention that we use here and in other places of our library (e.g. as a class attribute of the data interfaces).

Exactly right; the original issue with that (and still present elsewhere in the file) is a name like `RX` implies that its a **class**, not an **instance** of the class. Instances should be all lowercase with underscore separation whereas classes generally follow camel case.

PR #419, opened 2022-03-10
Sorting extractor interfaces' allow new spike interface sorting objects

Saksham20:
## Motivation

This PR makes changes that allow the `BaseSortingExtractorInterface` to allow new spiketerface sortingextractor objects. 
## Checklist


CodyCBakerPhD:
#379 and #394 are already open for this, taking it one interface at a time. #379 is the best one to start with, but will require thorough analysis of behavior and performance to ensure the `neo` backend does not have any unexpected differences from the `spikeextractors`.

Also, I always thought it would be a nice idea to not lose the work we had done previously by making the backend `neo` (`spikeinterface`) or `spikeextractors` optional and eventually making `neo` the default.

Saksham20:
> #379 and #394 are already open for this, taking it one interface at a time. #379 is the best one to start with, but will require thorough analysis of behavior and performance to ensure the `neo` backend does not have any unexpected differences from the `spikeextractors`.

Thanks for pointing these two existing PRs.
> Also, I always thought it would be a nice idea to not lose the work we had done previously by making the backend `neo` (`spikeinterface`) or `spikeextractors` optional and eventually making `neo` the default.
Lets keep this PR to only for the `BaseSortingExtractorInterface` to allow both old and new sorting extractor objects when creating a stub. Just like `BaseRecordingExtractorInterface`

CodyCBakerPhD:
Can you add a test that it works? Something like

```python
si_rec = generate_recording(...)
TestInterface(BaseSortingInterface):
    __init__(self, recording):
        self.recording_extractor = recording
        self.subset_channels = None

test_interface = TestInterface(recording=si_rec)
test_interface.run_conversion(nwbfile=minimal_nwbfile, metadata=dict(), stub_test=True)

# test output is as expected
```

Saksham20:
@CodyCBakerPhD I have added a test for sorting interface using the stub option and without. 
Currently, I have added these tests in `test_internals/test_converter.py`

CodyCBakerPhD (approved):

PR #422, opened 2022-03-10
add css for arg formatting

bendichter:
before:
Untitled

after:
Untitled 2


CodyCBakerPhD:
That is lovely, I'll copy it over to the NWBInspector as well!

CodyCBakerPhD (approved):

PR #425, opened 2022-03-14
fix css

bendichter:
## Motivation

fix #424


CodyCBakerPhD:
@bendichter Thanks, that's much better!

CodyCBakerPhD (approved):

PR #426, opened 2022-03-14
II Refactor of add electrodes

h-mayorquin:
## Motivation
I stooped midway with the refactor of our `add_electrodes` function in #406 because we were working on adding support for non integer channels #415 and I wanted to see how the final design would look like. This PR is the final refactor. 

What has been changed?
* Separate the logic of the function in three parts: 1) [data extraction](https://github.com/catalystneuro/nwb-conversion-tools/blob/f2bb36e5d8f17747d5fcb11dc1533f0a4e8b976b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L295-L302), 2) [data preparation](https://github.com/catalystneuro/nwb-conversion-tools/blob/f2bb36e5d8f17747d5fcb11dc1533f0a4e8b976b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L311-L359) and 3) electrode table construction by [rows](https://github.com/catalystneuro/nwb-conversion-tools/blob/f2bb36e5d8f17747d5fcb11dc1533f0a4e8b976b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L401-L406) and [columns](https://github.com/catalystneuro/nwb-conversion-tools/blob/f2bb36e5d8f17747d5fcb11dc1533f0a4e8b976b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L433-L449) clearly (separation of concerns). The logic of those processes were mixed. Two examples of this are:
 1) The creation of missing groups was previously done during the addition of rows to the electrodes table: https://github.com/catalystneuro/nwb-conversion-tools/blob/c3a1713958cdd3c756c5ddc9e5aadfd339a5ee0b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L380-L409
2)  Processing of extracted properties was previously done during the extraction of the data from the recorder:
https://github.com/catalystneuro/nwb-conversion-tools/blob/c3a1713958cdd3c756c5ddc9e5aadfd339a5ee0b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L312-L330

In these two examples, the mixing of the logic resulted in a complicated control flow structure as you can see by the depth of the control flow structure (meaning if and for directives here) in those two previous examples. As you can see in the diff the resulting structure is way flatter. 
* For adding information to the electrodes table, some information is added as rows and some information is added as columns. Before this refactor, which properties went to each of the streams was decided by popping information from a giant dic data structure that contain all the extracted properties from the spikeinterface recorder. This flow is now implemented with set data structure logic that should make it clear which properties are going to be added in what way:
https://github.com/catalystneuro/nwb-conversion-tools/blob/f2bb36e5d8f17747d5fcb11dc1533f0a4e8b976b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L376-L380
* Modified a test case to include handling of missing property.
* Unified and simplified variable naming through the function for consistency and clarity.

## Checklist

### DIFF
--- a/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
+++ b/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
@@ -289,9 +289,6 @@ def add_electrodes(
         [electrode["name"] != "group" for electrode in metadata["Ecephys"]["Electrodes"]]
     ), "Passing metadata field 'group' is deprecated; pass group_name instead!"
 
-    if nwbfile.electrode_groups is None or len(nwbfile.electrode_groups) == 0:
-        add_electrode_groups(recording=recording, nwbfile=nwbfile, metadata=metadata)
-
     # 1. Build columns details from extractor properties: dict(name: dict(description='',data=data, index=False))
     data_to_add = defaultdict(dict)
 
@@ -380,7 +377,7 @@ def add_electrodes(
     required_properties = set(required_property_to_default_value)
     extracted_properties = set(data_to_add)
     properties_to_add_by_rows = electrode_table_previous_properties | required_properties
-    properties_to_add_as_columns = extracted_properties - properties_to_add_by_rows
+    properties_to_add_by_columns = extracted_properties - properties_to_add_by_rows
 
     # Find default values for properties / columns already in the electrode table 
     type_to_default_value = {list: [], np.ndarray: np.array(np.nan), str: "", Real: np.nan}
@@ -399,9 +396,9 @@ def add_electrodes(
 
     properties_with_data = [property for property in properties_to_add_by_rows if "data" in data_to_add[property]]
     rows_in_data = [index for index in range(recording.get_num_channels())]
-    rows_to_add = [index for index in rows_in_data if channel_name_array[index] not in channel_names_used_previously]
+    rows_to_add_to_table = [index for index in rows_in_data if channel_name_array[index] not in channel_names_used_previously]
 
-    for row_index in rows_to_add:
+    for row_index in rows_to_add_to_table:
         electrode_kwargs = dict(property_to_default_values)
         for property in properties_with_data:
             electrode_kwargs[property] = data_to_add[property]["data"][row_index]
@@ -411,7 +408,7 @@ def add_electrodes(
     # Add channel_name as a column and fill previously existing rows with channel_name equal to str(ids)
     previous_table_size = len(nwbfile.electrodes.id[:]) - len(channel_name_array)
 
-    if "channel_name" in properties_to_add_as_columns:
+    if "channel_name" in properties_to_add_by_columns:
         cols_args = data_to_add["channel_name"]
         data = cols_args["data"]
 
@@ -433,7 +430,7 @@ def add_electrodes(
     indexes_for_default_values = electrodes_df.index.difference(indexes_for_new_data).values
     
     # Add properties as columns
-    for property in properties_to_add_as_columns - {"channel_name"}:
+    for property in properties_to_add_by_columns - {"channel_name"}:
         cols_args = data_to_add[property]
         data = cols_args["data"]
         if np.issubdtype(data.dtype, np.integer):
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@h-mayorquin Thanks a bunch, it'll be a couple days before I can thoroughly review this, but will try to do so before our meeting on Friday.

Can you also post that lovely figure you made the highlights the steps/conditions of the current algorithm?

CodyCBakerPhD:
> The only possible case when there is no electrodes table after this function is if somehow a recorder with no channels was passed. Is that even possible? Otherwise, I think we should throw a warning if no data whatsoever was added but I think that's it.

Agreed; I think this was for previous versions of the code that had the possibility of failing to add electrodes for various reasons. We can remove it now.

Even in the case of an empty recording, we should still be able to run `add_electrodes` on it without it failing, it's just that it shouldn't add an `ElectrodeTable` to the NWBFile in that edge case.

Really, we could make this clearer by just having a condition at the start of the function (and other functions as well) like

```python
if len(recording.get_channel_ids()) == 0)
    return nwbfile
```

for an early exit.


> My understanding is that this is to provide descriptions of the electrode tables properties, am I right? That's the only reason we have metadata here, correct?

Indeed; these are the rich `text` descriptions describing those custom columns that are added via properties of the `recording`. `spikeextractors` did not have a way really to set this internal to the recording object itself or I would use that. Perhaps something nice and new has been added in the latest SI, though?

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -253,36 +253,23 @@ def add_electrodes(
         object to ignore when writing to the NWBFile.
     """
     assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
+
     if isinstance(recording, RecordingExtractor):
-        checked_recording = OldToNewRecording(oldapi_recording_extractor=recording)
-    else:
-        checked_recording = recording
+        recording = OldToNewRecording(oldapi_recording_extractor=recording)
```
CodyCBakerPhD, line 258: Just so we're 100% on this behavior, what will happen now is that any time anyone uses a `recording` that is a `RecordingExtractor` (old type) in conjunction with `add_electrode` or `write_recording`, the `recording` they passed into the function will be modified in-place to become a `BaseRecording` mapped via the `OldToNew` class.

Personally not a fan of behavior like that because it can produce very unexpected results when/if the user isn't expecting it; _e.g._, if they had any lines after calling `add_electrodes` or `write_recording` in their script that assumed the recording to still be a `RecordingExtractor`, those lines will now fail due to the API differences.

I'd suggest restoring the `checked_recording`, but also feel free to make it a general function used across all the methods here to reduce amount of code.

CodyCBakerPhD:
@h-mayorquin One comment on the coercion of `recording` object, but otherwise looks great to me! Great documentation on all the steps, makes it much easier to follow and comprehend.

nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Aha!
This was my omission of understanding. Python passes most stuff by reference so this means that the outer-scope recording object will be modified. You are 100 % correct Cody.

h-mayorquin:
> Even in the case of an empty recording, we should still be able to run `add_electrodes` on it without it failing, it's just that it shouldn't add an `ElectrodeTable` to the NWBFile in that edge case.
> 
> Really, we could make this clearer by just having a condition at the start of the function (and other functions as well) like

I can think on the the following cases:
1) No new rows are added but some columns are. (no new channel_ids but yes new properties)
2) No new rows or columns are added because all the properties are already there. (no new channel_ids or properties)
3) No new rows are added because the recorder contains no new channels. (no channel_ids and therefore no-properties)

As it stand, the function works correctly for all of those cases without any need to catch exceptions or extra code. In fact, I realize that we should remove the following lines that we change recently because it gets on the way of 2:

https://github.com/catalystneuro/nwb-conversion-tools/blob/a4f6882588732d4510e1c1071e9a246191b1f956/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L260-L265

My view, is that we should not do any checks or early termination but just throw a warning in the end if no new information was added to the electrodes table after calling the function.

What do you think?

CodyCBakerPhD:
@h-mayorquin Ahh, yes I can see those other conditions now. Good catch!

> My view, is that we should not do any checks or early termination but just throw a warning in the end if no new information was added to the electrodes table after calling the function.

A warning in that event sounds fine, then.

CodyCBakerPhD (approved):

Issue #429, opened 2022-03-15
[Bug]: `check_recordings_equal` from `spikeinterface` fails when using non integer channel ids. 


h-mayorquin:
@bendichter @alejoe91 @samuelgarcia @CodyCBakerPhD 

## Problem and context
Hi, I wanted to discuss the following issue. We recently added capabilities for `spikeinterface` to handle non integer channels #415. During the discussion of that aforementioned PR the following difficulty ensued. Many of our round trip tests use `NwbRecordingExtractor` for testing that we indeed wrote the nwb file correctly. Here is an example:

https://github.com/catalystneuro/nwb-conversion-tools/blob/712c775a4660171fd8c1467eaf28e6d604538183/tests/test_internals/test_si.py#L289-L297

The problem with this approach is that, as it stands, the **`NwbRecordingExtractor` assumes that the `channel_ids` are equivalent to the `electrode.table.ids` and thus it fails for recordings with non integer channels** (see example in steps to reproduce below).

## Possible solutions
Some possible solutions that come to my mind are:
Within `spikeinterface`
1) Remove the assumption that channels ids are equivalent to the electrode ids altogether in the `NwbRecordingExtractor`.  
2) In `check_recordings_equal` if the channel_ids of any of the electrodes is of type other than integer, then skip the comparison of channel_ids. 

Within `nwb-conversion-tools` 
1) Remap the ids of `NwbRecordingExtractor` channel_ids to the field `channel_name` (as we implemented it) before we do a full recorder comparison with `check_recordings_equal`
2) Wrap the `NwbRecordingExtractor` so we change the assignation of ids from the `electrodes.table.id` to `electrodes.table.channel_name` as we already have the functionality to write it like this.

### Steps to Reproduce

```python
from nwb_conversion_tools.tools.spikeinterface import spikeinterface
from spikeinterface.core.testing_tools import generate_recording
from spikeinterface.core.testing import check_recordings_equal
from spikeinterface.extractors import NwbRecordingExtractor

from nwb_conversion_tools.tools.spikeinterface import get_nwb_metadata
from datetime import datetime

rec = generate_recording(num_channels=4, durations=[10])
channel_ids = rec.get_channel_ids()
non_int_channel_ids = ['a', 'b', 'c', 'd']
rec_non_int_channels = rec.channel_slice(channel_ids=channel_ids, renamed_channel_ids=non_int_channel_ids)

[...]

PR #430, opened 2022-03-15
Template issue for internal discussion

h-mayorquin:
## Motivation
While filling the issue in #429 I thought that the current templates seem more oriented for normal users. While a more rigid form might grant us more structured feedback from general users, it feels to me that we can benefit from a more flexible template for internal discussion. I see two main benefits

1) Reduced boilerplate which makes for a quicker writing of an issue.
2) No need to fit the problem in the categories of bug or feature request or documentation when they don't fall cleanly into it. 

This PR is for discussing this and I wrote a draft about how such template could look like. 

### DIFF
--- a/.github/ISSUE_TEMPLATE/discussion.yml
+++ b/.github/ISSUE_TEMPLATE/discussion.yml
@@ -1,0 +1,23 @@
+name: Internal Discussion
+description: General internal discussion for developers
+title: "[Discussion]: "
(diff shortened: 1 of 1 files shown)


h-mayorquin:
A solution for this was mentioned by @CodyCBakerPhD  in today's meeting:
image

As a way to open issues with any pre-format. So this can be done.

PR #432, opened 2022-03-18
Update neuroscope interface to use spike interface

h-mayorquin:
## Motivation

This is basically #394 updated after all the modifications that we did to spikeinterface (#426 #415). So, this should close #394 when it is merged. 

## Checklist


h-mayorquin:
Right now the tests are all working in my computer. 

With the CI however, the tests should fail as they require these modifications to spike-interface: https://github.com/SpikeInterface/spikeinterface/pull/550

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -164,7 +170,8 @@ class TestConverter(NWBConverter):
 
     @parameterized.expand(input=parameterized_recording_list, name_func=custom_name_func)
     def test_convert_recording_extractor_to_nwb(self, data_interface, interface_kwargs):
-        nwbfile_path = str(self.savedir / f"{data_interface.__name__}.nwb")
+        x = np.random.randint(10000)
```
h-mayorquin, line 173: So, I was having problems with the test because both neuroscope tests (differentiated by backend) have the same nwb_file name. Do you a better way? I have the feeling that there is something with the name_function above that we could adapt but maybe you remember it better? @CodyCBakerPhD

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -18,32 +19,37 @@
     from .neuroscope_utils import get_xml_file_path, get_channel_groups, get_shank_channels
 
     HAVE_LXML = True
+
 except ImportError:
     HAVE_LXML = False
 INSTALL_MESSAGE = "Please install lxml to use this interface!"
 
+SpikeInterfaceRecording = Union[se.RecordingExtractor, BaseRecording]
```
CodyCBakerPhD, line 27: Might as well import this from here: https://github.com/catalystneuro/nwb-conversion-tools/blob/update_neuroscope_interface_to_spike_interface/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L24

Rather then re-define it

requirements-full.txt
```diff
@@ -14,8 +14,8 @@ psutil==5.8.0
 lxml==4.6.5
 opencv-python==4.5.1.48
 spikeextractors==0.9.9
-spikeinterface==0.93.0
-neo==0.10.0
+spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git
+neo>0.10.0
```
CodyCBakerPhD, line 18: `requirements-full.txt` uses exact versioning; I'd return this to `==` and pin it to the exact version needed, such as `0.10.1`.

If you need to, update minimal requirements in `requirements-minima.txt` with lower bound versioning as `>=` the precise minimal version needed.

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: Yeah, just add `_se_backend_{interface_kwargs.get("spikeextractors_backend", False)}` to the f-string for the filename.

Aside from that, what's the problem? The fact that `overwrite=True` isn't used?

CodyCBakerPhD:
@h-mayorquin Looks good; the last couple of things... Can you

- [ ] Update all of those docstrings with the new argument?
- [ ] Run some basic performance comparisons on some `.dat` files and report; (a) the write speed to NWB in MB/s, and (b) try to note if you observe appropriate memory usage during the write process (will require a somewhat larger test file, at least 10GB with `buffer_gb` parameters set to detectable levels against basic system usage). Trying to make sure of two things; (i) no significant difference in I/O cost during writing, and (ii) no memory leaks (data being read into memory outside of the intended Iterator).

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: The file gets locked for some reason. I was wondering if the tests are run in parallel.

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
h-mayorquin: Great.
I turns out that I no longer need it in this specific instance. For the future though, maybe we should consolidate these imports in the `utils/typing` as we have some other types there?

h-mayorquin:
> @h-mayorquin Looks good; the last couple of things... Can you
> 
> * [x]  Update all of those docstrings with the new argument?
> * [ ]  Run some basic performance comparisons on some `.dat` files and report; (a) the write speed to NWB in MB/s, and (b) try to note if you observe appropriate memory usage during the write process (will require a somewhat larger test file, at least 10GB with `buffer_gb` parameters set to detectable levels against basic system usage). Trying to make sure of two things; (i) no significant difference in I/O cost during writing, and (ii) no memory leaks (data being read into memory outside of the intended Iterator).

Concerning the second point, does accepting this PR depends on this?
If it does not, can we move forwards with this PR and then do the performance checks as a separate PR?

I am thinking that should we have a set of gin-tests for testing performance that we can run when we get big changes. Maybe having a performance suite that we only run in specific PRs.

requirements-full.txt
h-mayorquin: Done.

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
CodyCBakerPhD: They are not, however, `utils` since they require external dependencies. They belong in tools, and the location in the `tools.spikeinterface` seems most appropriate.

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: File locking would indicate a potentially bad context exit, should look into that.

> The files for both backends have the same name for their respective nwbfile

I know, but tests are usually run in sequence to `overwrite=True` generally takes care of it unless there are issues with closing the file.

CodyCBakerPhD:
> Concerning the second point, does accepting this PR depends on this?

I would say so, we've had random major issues in the past with certain interfaces so it's better to be safe than sorry.

It really shouldn't take long, all you need is `from time import time` and insert `time()` value tracking at the right points. Full `cprofiles` wouldn't be a bad idea either for each backend to compare the spread of downstream function passing: https://www.machinelearningplus.com/python/cprofile-how-to-profile-your-python-code/

> If it does not, can we move forwards with this PR and then do the performance checks as a separate PR?

Such a thing does not involve any actual code pushed to the repo; only run locally, and on non-GIN files.

CodyCBakerPhD:
> I am thinking that should we have a set of gin-tests for testing performance that we can run when we get big changes. Maybe having a performance suite that we only run in specific PRs.

It's tough to run such things on GIN-level data, it's really something you want to run on examples of real data magnitudes and therefore only locally.

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: OK.
Where or how should I look at this? Any tips? (bad context exit)

nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
h-mayorquin: Fair enough.

CodyCBakerPhD:
@h-mayorquin

> It really shouldn't take long, all you need is from time import time and insert time() value tracking at the right points. Full cprofiles wouldn't be a bad idea either for each backend to compare the spread of downstream function passing:

Also also, you might be able to just enable the `tqdm` progress bar when writing the recording and that *should* display the total runtime of the actual steps involving only the data write. Probably the easiest way to do that.


And for memory, just watch how your `top`/`htop` looks during the writing process. To do it programmatically is pretty tricky.

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: > Where or how should I look at this? Any tips? (bad context exit)

That can be tricky, all you can do is try to reproduce and isolate the exact line of the issue then step back and examine the most immediate calls that opened/closed the file, which appears like it may be the `get_traces` calls from the previous `NwbRecordingExtractor` opened on the file for round-trip tests, but I don't see anything in the code for that that would cause an issue.

I'd say for now, just name them uniquely to bypass the issue and then see if you can systematically and reliably reproduce the error, then raise a formal issue on it and we can investigate it separately.

CodyCBakerPhD (approved):

PR #433, opened 2022-03-18
Update tutorials and add them to the documentation

h-mayorquin:
## Motivation
Updated the tutorials to ensure that they run and aligned them with their respective tests.

## Checklist

### DIFF
--- a/docs/data_interface.rst
+++ b/docs/data_interface.rst
@@ -13,9 +13,6 @@ Along with these two methods, you can also optionally override the :code:`__init
 The contributed extractors are in the **nwb-conversion-tools/datainterfaces** folder. You can fork the repo and create a new file
 **myformatdatainterface** there. In the folder, create a new file named **myformatdatainterface.py**.
 
-.. code-block:: python
-        ...
-
 
 When you are done we recommend you write a test in the **tests/test_interfaces.py** to ensure it works as expected.
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,20 +17,20 @@ types of neural data to the Neurodata Without Borders (NWB) standard.
     the `GitHub Issues <https://github.com/catalystneuro/nwb-conversion-tools/issues>`_ page.
 
 .. toctree::
-   :maxdepth: 2
-   :caption: User Guide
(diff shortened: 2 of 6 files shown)


h-mayorquin:
I am wondering whether (or if?) we should add the tutorials to the documentation. So far, I am thinking just in the user guide it will be OK. What do you think?

CodyCBakerPhD:
@h-mayorquin We definitely shouldn't have two explicitly different places with the same information, but the documentations is where most people are encouraged to start (many people will never even look at the GitHub code itself). And just linking to the notebooks to download would be fine...

But I wonder if there is an even better way, such as embedding the notebook via `myst_nb`: https://docs.readthedocs.io/en/stable/guides/jupyter.html

That way, there's only one thing that needs to be maintained (the notebooks, whose code is also integrated into the CI to alert us to breakages), but they still show up nice and rendered in the docs. Maybe even interactive? That would work for tutorial #1, but we do plan on having a second set of tutorials that require the user to download a bit of data from GIN.

h-mayorquin:
@CodyCBakerPhD 
I modified a little lit the structure of the documentation in-line with our discussion last Friday. I also added a new page with link to the tutorials as they are in our github page. I am concerned about the permanence of those links, that is, I am not sure that in the way that I wrote them they will always point to the latest version in master. Do you know if there is a way to assure ourselves of that?

Second, I gave it a try to trying to have the notebooks displayed on read the docs but my attempts with both `nbsphinx` and `myst_nb` failed. For both of them what stopped in my tracks were cryptic not clear errors when I was trying to build the errors even if simpler versions of the notebook worked well.  After a couple of hours I think that maybe we can try to do this together in another PR or once I have a better grip of sphinx and its ecosystem.

CodyCBakerPhD:
@h-mayorquin That's fine for now then and we'll explore other options in follow-up PRs.

For working examples, I'd suggest trying https://myst-nb.readthedocs.io/en/latest/examples/basic.html whose `.md` text is found at https://myst-nb.readthedocs.io/en/latest/_sources/examples/basic.md.txt

CodyCBakerPhD (approved):

PR #434, opened 2022-03-20
Solve gin test config tracking

h-mayorquin:
## Motivation

So, we consolidated our configuration for our gin tests in a single file. However, the problem is that if we want to keep that file in github it won't be ignored by version control (i.e. `gitignore`) and therefore we still have local changes to the config file being marked as untracked by git and just plain annoying. Reading around a bit most folks suggest that this problem can be avoiding by having setup create a local copy of a master file that then can be modified by the user -us- without messing with version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,7 +46,8 @@ coverage.xml
 *.cover
 .hypothesis/
 .pytest_cache/
-*.json
+gin_test_config_local.json 
+# *.json
 
 # Translations
 *.mo
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,8 @@
 from setuptools import setup, find_packages
 from codecs import open
 import os
+from pathlib import Path
+from shutil import copy
 
 path = os.path.abspath(os.path.dirname(__file__))
 
@@ -12,6 +14,12 @@ with open(os.path.join(path, "requirements-full.txt")) as f:
     full_dependencies = f.readlines()
 testing_suite_dependencies = ["pytest", "pytest-cov", "ndx-events==0.2.0", "parameterized==0.8.1"]
 extras_require = dict(full=full_dependencies, test=testing_suite_dependencies)
+
+# Create a local-folder for gin data tests configuration
+gin_config_file_base = "./tests/test_on_data/gin_test_config.json"
+gin_config_file_local = "./tests/test_on_data/gin_test_config_local.json"
+copy(src=gin_config_file_base, dst=gin_config_file_local)
+
 setup(
     name="nwb-conversion-tools",
     version="0.11.3",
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -46,7 +46,8 @@ coverage.xml
 *.cover
 .hypothesis/
 .pytest_cache/
-*.json
```
h-mayorquin, line 49: Why do we have this?

CodyCBakerPhD: Why wouldn't we have this? Note that `.gitignore` is only for when you do `git add *` or some other wildcard search. This forces people to type `git add filename` for an explicit change to a filename intended to be pushed.

CodyCBakerPhD:
@h-mayorquin I don't understand, what's the issue and how is this any different? In a typical PR workflow when running full tests, you'd branch off of `main`, update that file locally if needed, then simply don't call `git add gin_test_config_local.json` explicitly and it won't be included in the commit history.

h-mayorquin:
> @h-mayorquin I don't understand, what's the issue and how is this any different? In a typical PR workflow when running full tests, you'd branch off of `main`, update that file locally if needed, then simply don't call `git add gin_test_config_local.json` explicitly and it won't be included in the commit history.

Right now, the file `gin_test_config.json` is in version control. This is mainly so I can reference it on our github page as a reference and so that is downloaded automatically. However, this also implies that when I do local changes to `gin_test_config.json` the file is marked as modified in git because the file its been tracked.

Introducing a local copy that is not in version control `gin_test_config_local.json` but it is created at package installation guarantees that the master file `gin_test_config.json` remains a canonical version not intendend to be modified and designates `gin_test_config_local.json` as the local file intended to be modified.

.gitignore
h-mayorquin: I was wondering if you were tracking some specific files with this? Do json files appear on in some part of our processing (build or something) or is it just for the gin config files? 
That was the sense of the question, not that I think is a bad idea. I should have been more explicit.

CodyCBakerPhD:
Ahh, I see now. Even if it's in `.gitignore`, editing it will still show up under `git status`, it simply won't get added by a wildcard grab. But adding a new file under a `.gitignore` specification will prevent it from showing up at all with `git status`. Hmm...

However, doing this will require updating the developer documentation (right now, in the readme) to inform people to update only their local copy and not the main one, which gets a bit confusing. Maybe store the global config alongside `setup`, with the 'local' version being alongide the tests (where the current config is on main)?

CodyCBakerPhD:
@h-mayorquin Good to go?

CodyCBakerPhD (approved):

PR #438, opened 2022-03-24
get session_id and session_start_time from neuralynx header

### LINKED ISSUE #437: [Feature]: neuralynx session_start_time

pull session_start_time from neuralynx file header

_No response_

pull session_start_time from neuralynx file header

No.

### Code of Conduct

bendichter:
fix #437 

# To-Do
- [x] write tests


CodyCBakerPhD:
Neuralynx has a lot of versions: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/neuralynx

Do all of them support this, or is it an otherwise 'safe' operation to attempt to retrieve it (returns `None` as default if it doesn't find)?

bendichter:
@CodyCBakerPhD that's a good question. I'm having trouble finding documentation on all of the different file format versions. I guess the thing to do would be to run a test over all of the versions on gin

CodyCBakerPhD:
> I guess the thing to do would be to run a test over all of the versions on gin

@bendichter Yes, that is eventually something we wanted to do anyway. But maybe just for now, indicate/document that we currently only 'strongly' support version `v5.7.4`, but can offer 'soft' (not 100% guaranteed) of any others supported by `neo`.

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -16,10 +77,22 @@ class NeuralynxRecordingInterface(BaseRecordingExtractorInterface):
...
         self.recording_extractor = self.RX(extractors)
         self.recording_extractor.set_channel_gains(gains=gains)
+        try:
+            for i, filename in enumerate(nsc_files):
+                self.recording_extractor.set_channel_property(
+                    i,
+                    "filtering",
+                    get_filtering(filename),
+                )
+        except:
```
CodyCBakerPhD, line 94: What specific error occurs if `get_filtering` fails? Do you know?

It's a [E722](https://www.flake8rules.com/rules/E722.html) code style violation to use a bare except but as you know we can always use `BaseException` or `Exception` if we really need to (_i.e._, if we have no clue when/if it would fail)

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,11 +1,72 @@
...
+    index = header.find("TimeCreated") + 12
+    session_start_time = parser.parse(header[index : index + 19])
+
+    index = header.find("SessionUUID") + 12
+    session_id = header[index : index + 36]
+
+    return dict(
+        session_start_time=session_start_time,
+        session_id=session_id,
+        identifier=session_id,
```
CodyCBakerPhD, line 42:
```suggestion
```

Better to let the `identifier` get auto-set by us through `uuid` in the main converter, no?

CodyCBakerPhD: Or at the very least, generate another `uuid` at this line.

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
bendichter: I'm just trying to account for a file that is older and doesn't have the right header form

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
bendichter: why would we do that? a uuid will be generated if this is missing

bendichter: the session_id is guaranteed to be unique for each session. This could I guess lead to problems if you wanted to create two different NWB files for the same session

CodyCBakerPhD: In that case the user would manually override `metadata["NWBFile"]["session_id"]` if they wanted to, I simply don't want it to be possible for two NWBFiles made through this pipeline to have the same `identifier`, but if you're 100% sure that can't happen with the native Neuralynx ID here, then that's perfectly fine.

bendichter, line 42:
```suggestion
```

bendichter:
@CodyCBakerPhD where should tests go for metadata extraction from GIN files?

nwb_conversion_tools/testing/setup_paths.py
```diff
@@ -0,0 +1,38 @@
...
+
+OPHYS_DATA_PATH = LOCAL_PATH / "ophys_testing_data"
+if not OPHYS_DATA_PATH.exists():
+    pytest.fail(f"No folder found in location: {OPHYS_DATA_PATH}!")
+
+if test_config_dict["SAVE_OUTPUTS"]:
+    OUTPUT_PATH = LOCAL_PATH / "example_nwb_output"
+    OUTPUT_PATH.mkdir(exist_ok=True)
+else:
+    OUTPUT_PATH = Path(tempfile.mkdtemp())
```
h-mayorquin, line 38: I like the idea of having all the variables defined in a file so the paths can be imported as a normal python variables. This is the first thing that I tried when I wrote the pull request with the configuration file. A clear advantage here is code reduction as this boilerplate is repeated more or less across the different gin tests file. I remember that @CodyCBakerPhD  was against this but sadly I can' t remember the reason. 

That aside, would it not make more sense semantically to have this somewhere in `/tests` though? Basically we intend for this to be the only place where it is used, correct?

setup.py
```diff
@@ -17,7 +17,7 @@
 
 # Create a local copy for the gin test configuration file based on the master file `base_gin_test_config.json`
 gin_config_file_base = "./base_gin_test_config.json"
-gin_config_file_local = "./tests/test_on_data/gin_test_config.json"
+gin_config_file_local = "gin_test_config.json"
 copy(src=gin_config_file_base, dst=gin_config_file_local)
```
h-mayorquin, line 21: I guess you moved this to the top folder for ease of access after moving the config python file to tests, corrrect? or is there other reason that I am missing for this?

My view, as in the last comment, is that all of this should be on `/tests`.

h-mayorquin (commented):
It seems that the tests are failing in CI but locally the run fine. Only one failed in my systems, see the comments.

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,11 +1,80 @@
...
+    Parameters
+    ----------
+    folder_path: str
+
+    Returns
+    -------
+    dict
+
+    """
+    csc_files = glob(folder_path + "/*.ncs") or glob(folder_path + "/*.Ncs")
```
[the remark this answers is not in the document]
bendichter:
```suggestion
    csc_files = list(Path(folder_path).glob('*.[nN]cs'))
```

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,11 +1,80 @@
...
+    fpath = csc_files[0]
```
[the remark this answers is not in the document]
bendichter: probably because the files are not in a specific order. We could `sort` them to ensure that we are reading the same file.

Though the fact that these files have different start times might be concerning in itself..

nwb_conversion_tools/testing/setup_paths.py
bendichter: tried putting it in tests but wasn't able to get it to import. I'd be fine with putting it there if you can find a way to do it.

setup.py
bendichter: Fine to move this into `tests` if you can find a way to get it to work

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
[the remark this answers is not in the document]
h-mayorquin: Yes, I think sorting by name get us a consistent pipeline. I will take care of this.

We should keep an eye on future files to see how different the extracted times are when we have more than one `.nsc` file.

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,11 +1,80 @@
...
+    Parameters
+    ----------
+    folder_path: str
+
+    Returns
+    -------
+    dict
+
+    """
+    csc_files = list(Path(folder_path).glob("*.[nN]cs"))
```
bendichter, line 39:
```suggestion
    csc_files = sorted(Path(folder_path).glob("*.[nN]cs"))
```

nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,11 +1,80 @@
...
+    ----------
+    folder_path: str
+
+    Returns
+    -------
+    dict
+
+    """
+    csc_files = sorted(Path(folder_path).glob("*.[nN]cs"))
+    csc_files.sort()
```
bendichter, line 40:
```suggestion
```

nwb_conversion_tools/testing/setup_paths.py
bendichter: @h-mayorquin some background on this.

A `testing` module is a module that is imported by tests and can also be imported by other packages. For example, `numpy.testing`, which contains functions like `assert_almost_equal`, which are useful for many different projects and therefore has been packaged within numpy instead of in the tests. Also, sklearn does something similar [here](https://github.com/scikit-learn/scikit-learn/blob/1495f6924/sklearn/utils/testing.py).

I think it's an open question here whether any other package might want to import these gin paths. If we did want to do that, we'd probably want to put the config file in a more accessible location that does not require `git clone`ing NCT, such as in `~/.nwb-conversion-tools/gin_config.json` but I don't want to do that work until there is another library that actually wants to use this.

bendichter:
@CodyCBakerPhD question for when you get back: what is the proper workflow for adding gin data to the tests?

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
```diff
@@ -7,8 +7,7 @@
     NeuralynxRecordingInterface,
 )
 
-from nwb_conversion_tools.testing.setup_paths import ECEPHY_DATA_PATH
-
+from ..setup_paths import ECEPHY_DATA_PATH
```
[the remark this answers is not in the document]
h-mayorquin: This was a mistake push. I will revert. See above.

nwb_conversion_tools/testing/setup_paths.py
bendichter: So you want to add __init__s to the testing module and put all of our source code inside a `src` file? Sounds good to me.

h-mayorquin: Yes.

Maybe we can add the `__init__.py` to get the testing going here and do the `nwb_conversion_tools` to -> `src/nwb_conversion_tools` migration in a specific PR. What do you think?

bendichter: sounds good

h-mayorquin:
I am really confused about this CI errors. 

```python
FileNotFoundError: [Errno 2] No such file or directory: 'ephy_testing_data/neuralynx/Cheetah_v5.4.0/original_data/CSC5_trunc.Ncs'
```

The fact that the correct file is found by `Path(folder_path).glob("*.[nN]cs"))` above but then the statement ` with open(file_path, "r", encoding="latin1")` fails to find them makes me think it is a caching issue in the CI.

h-mayorquin:
All right, so it was a caching issue. Changing the key to the [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) finally solved the issue (see the changes to key in the workflow).

h-mayorquin (approved):
Looks good to me now. 

We should keep an eye when we get some more data to see the issue of multiple differing starting times.

PR #441, opened 2022-03-26
update workflows for windows 

Saksham20:
## Motivation

Problems:

- Testing out some variations in windows workflows to investigate any caching issues. A point noted by @CodyCBakerPhD  in #274 is that cache created by ubuntu is not accessible by windows. 

- Datalad issues with windows ( git-annex issues where in datalad is unable to download the files from GIN because of `annex-ignore=True` setting for the GIN remote). 

### 1. Permutation of installation and download methods:

**Datalad Installation method:**
1. Using `conda`. 
2. Using `conda` with manual `git-annex` re-installation to the latest version(using windows `chocolatey`). 
3. Installing datalad using pip. Instructions for windows are [here](https://handbook.datalad.org/en/latest/intro/installation.html#windows-10)
4. install using pip but manual git-annex re-install.

Data download method:
1. using only datalad (`datalad install <GIN dataset>` AND `datalad get`)
2. Using `datalad install` with `git annex get` 
3. Using `git clone <GIN repo>` and `git annex get` or `datalad get`

### 2. Stylistic changes:
1.  Since the steps for caching, download are common for all the 3 modalities: ephys, ophys, behavior. I have created a template and each of these are put in the build matrix running tests for each modality only and not `tests/test_internals`. 
**Advantage** of this being that each are independently sent to runners **Disadvantage**: we are forcing the download of the full `NeuralEnsamble/ephy` repo from GIN. The data for modalities that we currently have interfaces for are over 300MB but the whole repo is 1.3GB. But given the cache size (limit 10GB) I dont think this would be a problem.
2. Renaming of `test_gin_ecephys.py` to `test_gin_ephy.py`. 

### 3. Caching:
For the issue of windows not being able to access the cache created by linux/Mac, we could just create a different cache for windows? I have changed the key so all the ephys/behavior/ophys are a separate cache. 

## Checklist


Saksham20:
currently, installation of datalad using pip and manual git-annex installation (using chocolatey) and using datalad to install + download data works okay.

PR #444, opened 2022-03-31
Fix yaml conversion script

Saksham20:
## Motivation

The yaml conversion script assumes all the arguments to child data interfaces of `PathType`. So it appends the base data folder location to the `source_data` before instantiating the `CustomNWBConverter`. 
Why are we making this assumption? There could be many other arguments. For one, the new `spikeextractors_backend` argument for all data interfaces eventually. 

Of all the datainterfaces(ephys, ophys, behavior) only `BlackrockRecordingExtractorInterface` has an argument of `FilePathTYpe` which does not have "path" in arg name. We can make it `nsx_override` > `nsx_override_path` to make it more explicit and follow the standard conversion for such arguments. 

## Checklist


CodyCBakerPhD:
Raised as a long-term low-priority issue; this PR has a massive commit history which would need to be rebased/squashes to even consider, breaks backcompatability of an interface, is untested and undocumented, and doesn't really fix the issue at hand. I'll get around to this in the next couple of weeks.

Saksham20:
@CodyCBakerPhD 
> this PR has a massive commit history which would need to be rebased/squashed to even consider,

It has a long commit history because it picked up the commits from the branch of PR #379. I had intended to have this as a sub-PR to that, but I realised this was better as an independent PR so manually reverted all changes to files of #379 except the 2 files. (last two commits: https://github.com/catalystneuro/nwb-conversion-tools/pull/444/commits/c3e2b18844cf7320823b69ba02569b6a18b83f9e and https://github.com/catalystneuro/nwb-conversion-tools/pull/444/commits/c804e8dc1df940824367fa8534e7363cc1cedc56 are for reverting). 
Can't this be squashed and merged if needed? 

>  breaks backcompatability of an interface, is untested and undocumented

Yes, there is only that change to `BlackrockRecordingExtractorInterface`. The api doc [here](https://nwb-conversion-tools.readthedocs.io/en/main/api/interfaces.ecephys.html#module-nwb_conversion_tools.datainterfaces.ecephys.blackrock.blackrockdatainterface) will update after rebuilding. 

> and doesn't really fix the issue at hand

The issue is that all arguments to a data interface are being suffixed with the base path.

PR #445, opened 2022-03-31
removing subset_channels for stub

Saksham20:
## Motivation

I think, stub tests should only slice the recording in frames and not channels to be as realistic as possible. 

Thus the `subset_channels` attribute is unnecessary.
 In addition, the `BaseRecordingExtractorInterface.subset_recording()` is useful only when `stub_test=True` thus its unnecessary to have `stub_test` as argument to that. 

## Checklist


bendichter:
Closing due to too many merge conflicts

PR #446, opened 2022-03-31
update workflow for Mac

Saksham20:
## Motivation

Updating the Mac CI to download ephy/ophys/behavior datasets from GIN using datalad.

### DIFF
--- a/.github/workflows/testing-mac.yml
+++ b/.github/workflows/testing-mac.yml
@@ -34,19 +34,25 @@ jobs:
 
       - name: Install full requirements
         run: pip install .[full]
-      # - name: Get ephy_testing_data current head hash
-      #   id: ephys
-      #   run: echo "::set-output name=HASH_EPHY_DATASET::$(git ls-remote https://gin.g-node.org/NeuralEnsemble/ephy_testing_data.git HEAD | cut -f1)"
-      # - name: Cache ephys dataset - ${{ steps.ephys.outputs.HASH_EPHYS_DATASET }}
-      #   uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/testing-mac.yml
```diff
@@ -34,19 +34,66 @@ jobs:
...
+          datalad get -r ./neuralynx/Cheetah_v5.7.4/original_data/
+          datalad get -r ./neuroscope/test1/
+          datalad get -r ./openephysbinary/v0.4.4.1_with_video_tracking/
+          datalad get -r ./blackrock/
+          datalad get -r ./intan/
+          datalad get -r ./spikegadgets/
+          datalad get -r ./spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/
+          datalad get -r ./phy/phy_example_0/
+          datalad get -r ./axona/
+          datalad get -r ./cellexplorer/
```
bendichter, line 82: Is there a way we can incorporate these calls into the .py files so we don't have to keep updating this every time we add a new format?

CodyCBakerPhD: Incorporating them into the test files simply passes the buck by requiring the extra line to be added somewhere else; an explicit pointer from a format to a GIN location would always have to be specified at some level.

It's on @h-mayorquin low-priority to-do list however to centralize these sets of lines into a single place so it only has to be updated one per format addition

Although, what we can probably start doing, now that we support such a large number of neo formats, is just call a single line of  `datalad -rg name_of_dataset` and that will handle both issues.

PR #448, opened 2022-04-01
[Bug]: Fix Windows YAML schema reference

CodyCBakerPhD:
Apparently windows has a different standard from linux/mac when it comes to URI reference resolvers. This PR fixes it.


bendichter:
lgtm

PR #450, opened 2022-04-04
[DO NOT MERGE] refactor of `add_units` function

h-mayorquin:
## Motivation

This refactor of the `add_units` function should include:

* Build on new `spikeinterface` interface capabilities to refactor the `add_units` function in `tools.spikeinterface`
* Likewise, spikeinterface now allows for non integers units ids. A new property called `unit_name` is to be added to the units table so it serves as an absolute reference across sorting objects.
* Separate waveform data addition from units table creation.

## How to test the behavior?
I added a new collection of unit-tests covering the new uses case of the `add_units` function.


CodyCBakerPhD:
@h-mayorquin Is this ready for review?

h-mayorquin:
> @h-mayorquin Is this ready for review?

Yes, it is!

CodyCBakerPhD:
@h-mayorquin This is a pretty big PR - before I dig deep into details, I'll just ask if there's **any way at all** this can be split into multiple sub-parts, or is it all intrinsically tangled together?

h-mayorquin:
> @h-mayorquin This is a pretty big PR - before I dig deep into details, I'll just ask if there's **any way at all** this can be split into multiple sub-parts, or is it all intrinsically tangled together?

@CodyCBakerPhD 
Yes, I was thinking on this in the context of our current discussions about smaller PRs. 

Conceptually, this PR provides the following features / functionality:
1) Separating waveform addition from standard unit table creation.
2) Adding "unit_name" as a parameter to uniquely identify a unit across sorters.
3) General refactor of the code for clarity and structure (e.g. `spikeinterface` machinery for property extraction, breaking loops, etc.)
4) Correct the bug where the function does not modify at all an already existing units_table:
https://github.com/catalystneuro/nwb-conversion-tools/blob/673410a79402b31a919e0e13e2bf7ed84732d15c/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L1199-L1202

This and the UnitTest addition which are built for the combination of 2-3.

I think separating 1 is totally feasible. I can take a quick stab at it if you think is a good idea.
The code for 2-3 is highly couple and more importantly highly coupled with test addition. 4 could be separated from 2 and 3 but I suspect is going to have bugs if it is not done together with 2 and 3.

Maybe I can take you through the code in our next catch up to speed things up and decide a good course of action?  If you could check the tests that could be great though.

CodyCBakerPhD:
> The code for 2-3 is highly couple and more importantly highly coupled with test addition. 4 could be separated from 2 and 3 but I suspect is going to have bugs if it is not done together with 2 and 3.

2-3 fine to include together then - do you mean that 2+3 would have true 'bugs' without 4, or just less features?

h-mayorquin:
> > The code for 2-3 is highly couple and more importantly highly coupled with test addition. 4 could be separated from 2 and 3 but I suspect is going to have bugs if it is not done together with 2 and 3.
> 
> 2-3 fine to include together then - do you mean that 2+3 would have true 'bugs' without 4, or just less features?

Doing 4 alone might be possible, but I think it will have bugs without 2 and 3 which is where the logic for multiple calls of `add_units_table` is resolved.

CodyCBakerPhD:
> Doing 4 alone might be possible, but I think it will have bugs without 2 and 3 which is where the logic for multiple calls of add_units_table is resolved.

But see, in that case you could just splinter 4 from 2-3 as a follow-up.

I'm not saying you need separate independent PR's, a chain of them would be fine so long as the changelog at each steep is something digestible.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1015,84 +1015,70 @@ def write_recording(
...
-    skip_features: Optional[List[str]] = None,
-    use_times: bool = True,
-    write_as: str = "units",
-    units_name: str = "units",
-    units_description: str = "Autogenerated by nwb_conversion_tools.",
+    property_descriptions: dict = dict(),
+    skip_properties: Optional[List[str]] = list(),
+    units_table_name: str = "units",
+    unit_table_description: str = "Autogenerated by nwb_conversion_tools.",
+    write_units_table_in_processing_module: bool = False,
```
CodyCBakerPhD, line 1025: Instead of `write_as`, which isn't very specific or explanatory, or `write_units_table_in_processing_module` which is quite a bit long for a keyword argument, what would you think about about `write_as_processing_module: bool = False`?

PR #452, opened 2022-04-09
Update repo to src structure

h-mayorquin:
## Motivation

In #438 we introduced package structure for our testing sets. 
However, this introduces some problems, see the relevant section in the [pytest documentation](https://docs.pytest.org/en/latest/explanation/goodpractices.html#tests-outside-application-code
):

>Now pytest will load the modules as tests.foo.test_view and tests.bar.test_view, allowing you to have modules with the same name. But now this introduces a subtle problem: in order to load the test modules from the tests directory, pytest prepends the root of the repository to sys.path, which adds the side-effect that now mypkg is also importable.

>This is problematic if you are using a tool like [tox](https://docs.pytest.org/en/latest/explanation/goodpractices.html#tox) to test your package in a virtual environment, because you want to test the installed version of your package, not the local code from the repository.

This problem can be avoided by transitio nto the [src package structure](https://py-pkgs.org/04-package-structure#the-source-layout) which is also recommended by the [Python Package Authority](https://packaging.python.org/en/latest/tutorials/packaging-projects/) so we are taking an opportunity to jump to the standard.

As you can see from the diff, besides the changes to some yaml tests that load the schema as a file (and therefore require the address of the directory) the changes are minimal. So far, the tests are running correctly in my machine, let's see how they do in the CI.

## Checklist


bendichter (approved):
lgtm. I like the idea of adhering to the python packaging guide best practices. thoughts, @CodyCBakerPhD?

CodyCBakerPhD (approved):

CodyCBakerPhD:
@bendichter @h-mayorquin lgtm

PR #456, opened 2022-04-11
Update spikeglx to use spikeinterface

h-mayorquin:
## Motivation

Should close #379
Updating the spikeglx datainterface to use the `spikeinterface` instead of `spikeextractors`. 

## How to test the behavior?
I added gin_tests for the new back-end. 
## Checklist


src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -1,29 +1,47 @@
...
+    """
 
-    if isinstance(recording, SubRecordingExtractor):
-        n_shanks = int(recording._parent_recording._meta.get("snsShankMap", [1, 1])[1])
-        session_start_time = datetime.fromisoformat(recording._parent_recording._meta["fileCreateTime"]).astimezone()
+    # Support for old spikeextractors objects
+    recording = recording._kwargs.get("oldapi_recording_extractor", recording)
+    if isinstance(recording, RecordingExtractor):
+        if isinstance(recording, SubRecordingExtractor):
+            meta = recording._parent_recording._meta
```
bendichter, line 29: This leading underscore is marking `meta` as private. It means that external libraries are not advised to rely on this variable and that spikeinterface is not guaranteeing backwards compatibility support going forward. If could be OK as a temporary fix, but ideally I'd rather not access private attributes from other repos.

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -1,29 +1,47 @@
...
+
+    # These correspond to single shank channels https://billkarsh.github.io/SpikeGLX/help/imroTables/
+    imDatPrb_type = meta["imDatPrb_type"]
+    if imDatPrb_type not in ["0", "21"]:
         raise NotImplementedError("SpikeGLX metadata for more than a single shank is not yet supported.")
-    metadata["NWBFile"] = dict(session_start_time=session_start_time.strftime("%Y-%m-%dT%H:%M:%S"))
+
+    extracted_start_time = meta.get("fileCreateTime", None)
+    if extracted_start_time:
+        session_start_time = datetime.fromisoformat(extracted_start_time).astimezone().strftime("%Y-%m-%dT%H:%M:%S")
```
bendichter, line 43: Why are you converting this from a string to a datetime back to a string?

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
h-mayorquin: This I inherited from the code before and it is not my solution.

 That said, It is private attribute in the old `spikeextractors` but they are not moving that anymore. As you can see in the new `spikeinterface` (below that) the same meta dictionary (which is just a the metadata file parsed) is provided as well but now with a public interface. If you want to remove the access in `spikeextractors` we can parse the metadata ourselves but that seems like an overkill.
 
 What do you think?

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
h-mayorquin: My mistake, I also got this code from the previous implementation and did not stop to think about it. I just assumed that the metadata needed to be in a certain format and that this pipeline was doing it.

But as far as I remember the metadata needs to be in isoformat, correct? which the first element in the pipeline seems to assume anyway. Testing does confirm it can be removed.

bendichter: NWBFile can take a `datetime` object

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
bendichter: I think it's fine if this is just used for the old extractor and a public attribute for the new extractor

bendichter:
this looks good to me. @CodyCBakerPhD , what do you think? This will touch a lot of conversions, so let's think through this PR carefully.

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
h-mayorquin: Right! which is what we did in the other PRs for this :)

h-mayorquin:
After changing `session_start_time` from string to datetime one of the yaml tests is failing saying that session_start_time should be of type string. This is strange as the other non yaml conversions seem to be working.

```python
        if error is not None:
>           raise error
E           jsonschema.exceptions.ValidationError: datetime.datetime(2020, 11, 3, 10, 35, 10) is not of type 'string'
E           
E           Failed validating 'type' in schema['properties']['NWBFile']['properties']['session_start_time']:
E               {'default': datetime.datetime(2020, 11, 3, 10, 35, 10),
E                'description': 'the start date and time of the recording session',
E                'format': 'date-time',
E                'type': 'string'}
E           
E           On instance['NWBFile']['session_start_time']:
E               datetime.datetime(2020, 11, 3, 10, 35, 10)
```

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -1,36 +1,65 @@
...
+            meta = recording._parent_recording._meta
+        else:
+            meta = recording._meta
+        # imDatPrb_type 0 and 21 correspond to single shank channels
+        # see https://billkarsh.github.io/SpikeGLX/help/imroTables/
+        imDatPrb_type = meta["imDatPrb_type"]
+        if imDatPrb_type not in ["0", "21"]:
+            raise NotImplementedError(
+                "More than a single shank is not supported in spikeextractors, use the new spikeinterface."
+            )
```
CodyCBakerPhD, line 37: So as I understand the slack discussion, this sort of thing would actually more properly be done in the future by using ProbeInterface?

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -43,18 +72,29 @@ def get_source_schema(cls):
...
+            self.recording_extractor = OldToNewRecording(oldapi_recording_extractor=self.recording_extractor)
+        else:
+            file_path = Path(file_path)
+            folder_path = file_path.parent
+            stream_id = "".join(file_path.suffixes[:-1])[1:]
+            super().__init__(folder_path=folder_path, stream_id=stream_id)
+            self.source_data["file_path"] = str(file_path)
+
         if stub_test:
             self.subset_channels = [0, 1]
```
CodyCBakerPhD, line 94: I'd prefer to remove this bit that subsets channels under stub mode; it's not expensive to include a multiple of 384 channels compared to a 20-30kHz temporal resolution, so might as well just stub time for all channels. This helps a lot when testing electrode metadata as well as the overall writing process

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -43,18 +72,29 @@ def get_source_schema(cls):
...
+
```
CodyCBakerPhD, line 95: I'd prefer to remove this bit that subsets channels under stub mode; it's not expensive to include a multiple of 384 channels compared to a 20-30kHz temporal resolution, so might as well just stub time for all channels. This helps a lot when testing electrode metadata as well as the overall writing process

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -90,18 +128,28 @@ def get_source_schema(cls):
...
+        spikeextractors_backend: Optional[bool] = False,
+    ):
+        if spikeextractors_backend:
+            self.RX = se.SpikeGLXRecordingExtractor
+            super().__init__(file_path=str(file_path))
+            self.recording_extractor = OldToNewRecording(oldapi_recording_extractor=self.recording_extractor)
+        else:
+            file_path = Path(file_path)
+            folder_path = file_path.parent
+            stream_id = "".join(file_path.suffixes[:-1])[1:]
```
CodyCBakerPhD, line 144: How robust is this parsing sub-function for the stream id? What if the stem has been altered slightly, say to include extra '.'?

CodyCBakerPhD:
@h-mayorquin Mostly looks good; biggest thing to fix is the stubbing of channels.

After addressing above comments can you do the  usual inspection of performance comparisons between the two backends (for both ap and lf) and post the 4 sample outputs from GIN? (`('ap', 'lf') x (se_backend=True, False)`)

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
h-mayorquin: This is done now.
Out of curiosity, is there a case where we want to subset channels other than eliminating non signal channels as we do in Neuroscope?

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
h-mayorquin: Yes, I saw that and it bothers me. I added now an extra utility to do the parsing in a more robust and, in my view, more transparent / explicit fashion.

h-mayorquin (commented):
I re-structured the code so the functions have a more specific responsibility. I also added a tests for the fetching of the starting time as @CodyCBakerPhD requested in another PR.

To-do:
* Fetch the files for the different combinations of back-end type of signal.
* Do some performance tests.

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -90,18 +108,33 @@ def get_source_schema(cls):
...
+            _assert_single_shank_for_spike_extractors(self.recording_extractor)
+            self.meta = _fetch_meta_dic_for_spikextractors_spikelgx_object(self.recording_extractor)
+            self.recording_extractor = OldToNewRecording(oldapi_recording_extractor=self.recording_extractor)
+
+        else:
+            file_path = Path(file_path)
+            folder_path = file_path.parent
+            super().__init__(folder_path=folder_path, stream_id=self.stream_id)
+            self.source_data["file_path"] = str(file_path)
+            self.meta = self.recording_extractor.neo_reader.signals_info_dict[(0, self.stream_id)]["meta"]
```
bendichter, line 134: It looks like there is some duplicated code here. Is there a way of reducing it? What about a mixin class?

CodyCBakerPhD: For now we'll use a common auxiliary function to reduce code duplication.

Long term plan will indeed be to use the more sophisticated mixin, but that has potential for unexpected behavior if not very thoroughly tested; since this PR already does so much we might as well just leave it for a follow-up PR dedicated just to that.

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -132,19 +143,35 @@ class TestConverter(NWBConverter):
                     interface_kwargs=interface_kwargs,
                 )
             )
-    for suffix in ["ap", "lf"]:
+
+    for suffix, spikeextractors_backend in product(["ap", "lf"], [True, False]):
```
h-mayorquin, line 147: @CodyCBakerPhD 
So, I did not notice until now but this was in the previous form of the test already.
Should we be really testing both "ap" and "lf" signals using `SpikeGLXRecordingInterface` for both?
Shouldn't we use `SpikeGLXLFPInterface` for `lf`?
Was there a reason for having it like this or is it an oversight?

CodyCBakerPhD: You can safely remove it, that's likely from the days before the lfp specific test class

h-mayorquin: All right, done.

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
bendichter: I'd be fine with an aux function, but I'd rather not add duplicate code

CodyCBakerPhD:
@h-mayorquin With the performance results in, last item before approval is just reducing that duplicate code Ben pointed out.

PR #457, opened 2022-04-11
[Feature]: Helper functions for inverse f-string YAML specification

CodyCBakerPhD:
This is a pre-PR introducing some core helper functions that will be used in the inverse f-string iteration feature for the YAML conversion specification.

The docstrings and testing behavior are hopefully self-explanatory enough, but the short version is there are two steps:

(a) `get_fstring_variable_names`: Very simple, just decomposes an f-string into the variable names and the remnants around those variables.

(b) `get_fstring_values_from_filename`: More complicated, but utilizes a dynamically constructed regular expression to grab instances of individual variables from an instance of the f-string rule.


src/nwb_conversion_tools/utils/globbing.py
```diff
@@ -0,0 +1,56 @@
...
+        String containing non-empty substrings enclosed by "{" and "}".
+        These correspond to the names of variables thought to encode the actual filename string.
+    """
+    variable_names, separators = get_fstring_variable_names(fstring=fstring)
+    pattern = "^"
+    for separator in separators[:-1]:
+        pattern += f"{separator}(.+)"
+    if separators[-1] != "":
+        pattern += f"{separators[-1]}"
+    pattern += "$"
```
h-mayorquin, line 48:
```suggestion
    pattern = "^" + "(.+)".join(separators) + "$" 
    # The final pattern should be a regular expression of the form "^separator(.+)separator(.+)separator(.+)separator"
    pattern_match = re.findall(pattern=pattern, string=filename)
```

tests/test_internals/test_globbing_utils.py
```diff
@@ -0,0 +1,74 @@
...
+                    "There is an empty separator between two variables in the f-string! "
+                    "The f-string will not be invertible."
+                ),
+            ):
+                get_fstring_variable_names(fstring="a/{x}{y}/c{z}")
+
+    def test_get_fstring_values_from_filename_non_invertible_assertion(self):
+        with self.assertRaisesWith(
+            exc_type=AssertionError,
+            exc_msg=("Adjacent variable values contain the separator character! The f-string is not invertible."),
```
h-mayorquin, line 30: Not every important but I find the concept of invertible f-string hard to be not very transparent. That is, I can't really understand that assertion message. That said, I have no better option as I don't really know the context in which this will be used.

CodyCBakerPhD: I'll probably add extensive documentation of this kind of thing as an extra section in the YML tutorial. But the basic fix is for the user to rename their files to use more distinctive separations between variables in the f-string.

h-mayorquin:
> 
> Thanks for pointing that out - I took a look at it, unfortunately it does not handle all the edge cases mine does and would be yet another external dependency to add it here. Example of failure
> 
> ```python
> from parse import findall
> fmt = "a/{x}b{y}/c{z}"
> txt = "a/foobbar/cthat"  # also txt = "a/fobobbar/cthat", where x="fobo" and y="bar"
> res = list(findall(fmt, txt))
> ```

To me the killer argument is the need to another extra dependency when the use case is restrictive and contained enough. That said, I don't know why the script above fails but this works as intended:

```python
In [1]: from parse import parse

In [2]: fmt = "a/{x}b{y}/c{z}"

In [3]: txt = "a/foobbar/cthat"

In [4]: parse(fmt, txt).named
Out[4]: {'x': 'foo', 'y': 'bar', 'z': 'that'}


```

CodyCBakerPhD:
> To me the killer argument is the need to another extra dependency when the use case is restrictive and contained enough. That said, I don't know why the script above fails but this works as intended:

Only by chance that it splits on first encounter of the separator character. Change it to 

```python
from parse import parse
fmt = "a/{x}b{y}/c{z}"
txt = "a/fobobbar/cthat" # x = fobo, y = bar
res = parse(fmt, txt).named
# res = {'x': 'fo', 'y': 'obbar', 'z': 'that'}
```
the `re` version fails due to the opposite convention of splitting on final encounter of the separator

```python
import re
from nwb_conversion_tools.utils import get_fstring_variable_names
fmt = "a/{x}b{y}/c{z}"
_, separators = get_fstring_variable_names(fstring=fmt)
txt = "a/fobobbar/cthat"
res = re.findall(pattern="^" + "(.+)".join(separators) + "$", string=txt)
# res = [('fobob', 'ar', 'that')]
```

Though in both cases we could handle non-unique invertibility assertion the same way so this would really just propose replacing the `re` with `parse`.

h-mayorquin:
> > To me the killer argument is the need to another extra dependency when the use case is restrictive and contained enough. That said, I don't know why the script above fails but this works as intended:
> 
> Only by chance that it splits on first encounter of the separator character. Change it to
> 
> ```python
> from parse import parse
> fmt = "a/{x}b{y}/c{z}"
> txt = "a/fobobbar/cthat" # x = fobo, y = bar
> res = parse(fmt, txt).named
> # res = {'x': 'fo', 'y': 'obbar', 'z': 'that'}
> ```
> 
> the `re` version fails due to the opposite convention of splitting on final encounter of the separator
> 
> ```python
> import re
> from nwb_conversion_tools.utils import get_fstring_variable_names
> fmt = "a/{x}b{y}/c{z}"
> _, separators = get_fstring_variable_names(fstring=fmt)
> txt = "a/fobobbar/cthat"
> res = re.findall(pattern="^" + "(.+)".join(separators) + "$", string=txt)
> # res = [('fobob', 'ar', 'that')]
> ```
> 
> Though in both cases we could handle non-unique invertibility assertion the same way so this would really just propose replacing the `re` with `parse`.

My impression is that you can get rid of the first function

> > To me the killer argument is the need to another extra dependency when the use case is restrictive and contained enough. That said, I don't know why the script above fails but this works as intended:
> 
> Only by chance that it splits on first encounter of the separator character. Change it to
> 
> ```python
> from parse import parse
> fmt = "a/{x}b{y}/c{z}"
> txt = "a/fobobbar/cthat" # x = fobo, y = bar
> res = parse(fmt, txt).named
> # res = {'x': 'fo', 'y': 'obbar', 'z': 'that'}
> ```
> 
> the `re` version fails due to the opposite convention of splitting on final encounter of the separator
> 
> ```python
> import re
> from nwb_conversion_tools.utils import get_fstring_variable_names
> fmt = "a/{x}b{y}/c{z}"
> _, separators = get_fstring_variable_names(fstring=fmt)
> txt = "a/fobobbar/cthat"
> res = re.findall(pattern="^" + "(.+)".join(separators) + "$", string=txt)
> # res = [('fobob', 'ar', 'that')]
> ```
> 
> Though in both cases we could handle non-unique invertibility assertion the same way so this would really just propose replacing the `re` with `parse`.

Isn't 

> > To me the killer argument is the need to another extra dependency when the use case is restrictive and contained enough. That said, I don't know why the script above fails but this works as intended:
> 
> Only by chance that it splits on first encounter of the separator character. Change it to
> 
> ```python
> from parse import parse
> fmt = "a/{x}b{y}/c{z}"
> txt = "a/fobobbar/cthat" # x = fobo, y = bar
> res = parse(fmt, txt).named
> # res = {'x': 'fo', 'y': 'obbar', 'z': 'that'}
> ```
> 
> the `re` version fails due to the opposite convention of splitting on final encounter of the separator
> 
> ```python
> import re
> from nwb_conversion_tools.utils import get_fstring_variable_names
> fmt = "a/{x}b{y}/c{z}"
> _, separators = get_fstring_variable_names(fstring=fmt)
> txt = "a/fobobbar/cthat"
> res = re.findall(pattern="^" + "(.+)".join(separators) + "$", string=txt)
> # res = [('fobob', 'ar', 'that')]
> ```
> 
> Though in both cases we could handle non-unique invertibility assertion the same way so this would really just propose replacing the `re` with `parse`.

As things stands with parse you would get rid of the first function (`get_fstring_variable_names`) and the second one would become a call to parse(fstring, filename) with a wrapper for some common errors, wouldn't it be.

Let's see how it looks once we have it implemented. Right now we have the advantage of alpha.

src/nwb_conversion_tools/utils/globbing.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Also returns the separators, the strings remnants that surround the variables.
+
+    Example
+    -------
+        variable_names, separators = get_fstring_variable_names(fstring="a/{x}b{y}/c{z}")
+        variable_names = ["x", "y", "z"]
+        separators = ["a/", "b", "/c"", ""]
+    """
+    matches = re.findall("{.*?}", fstring)
```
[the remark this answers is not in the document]
CodyCBakerPhD: It's a good point; I've added comments to highlight this. I'm not a fan of breaking the actual code lines up one by one, but instead just using one-line summary or comment blocks.

CodyCBakerPhD:
@h-mayorquin I found a case where `parse` differs in an uninformative way

```python
from parse import parse
parse("a/foobthat/cbar/sub-wrong", "a/{x}b{y}/c{z}/sub-{x}")
```

simply returns `None`, same as any other match failure, whereas now

```python
from nwb_conversion_tools.utils import get_fstring_values_from_filename
get_fstring_values_from_filename(filename="a/foobthat/cbar/sub-wrong", fstring="a/{x}b{y}/c{z}/sub-{x}")
```

raises an informative error explaining what happened.

This simply highlights how doing this sort of light-weight thing ourselves allows for further customization if ever need to handle things more specifically than `parse` allows.

src/nwb_conversion_tools/utils/globbing.py
```diff
@@ -0,0 +1,56 @@
...
+    variable_names = [match.lstrip("{").rstrip("}") for match in matches]
+    assert not any(
+        (variable_name == "" for variable_name in variable_names)
+    ), "Empty variable name detected in f-string! Please ensure there is text between all enclosing '{' and '}'."
+    separators = [x.rstrip("{").lstrip("}") for x in re.findall(pattern="^.*?{|}.*?{|}.*?$", string=fstring)]
```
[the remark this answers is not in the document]
h-mayorquin: Fair enough. I personally distrust regex code and I also think that the line commenting plays nice with the diff of VC. 

That said, I agree with your concerns, this case is sufficiently simply and this seems to be an opinion / aesthetical preference more than anything : )

h-mayorquin:
Sounds, good. 
At least the discussion with parse raise some other possible examples and failure modes. Maybe you would like to add those as unit tests to document this discussion and freeze intent?

CodyCBakerPhD:
> At least the discussion with parse raise some other possible examples and failure modes

Agreed, can never have enough tests or documentation. I've expanded the docstring for the second function to describe the uniqueness of invertibility problem. Also emphasized it's the *uniqueness* of the invertibility that's the issue, not the invertibility itself [minor mathematical detail there - for background, any *relation* (set of ordered pairs) is invertible in the sense of reversing the order of the pairs, but not every inverse of a *function* (relation such that no two pairs share the same first component) is itself a function, which is the common usage implied by invertibility]. So really, we're just testing if the f-string instances are functions such that the inverse is therefore unique.

Also good to know this problem has been approached before. I tried looking for similar things before starting this work but somehow was not able to find that library.

I've adjusted/generalized these function names and inputs to be closer to their approach, though I personally object to having keyword arguments that use reserved names like 'format' (see `parse?` for full list).

> Maybe you would like to add those as unit tests to document this discussion

Already did 😉

PR #458, opened 2022-04-18
[Testing]: Add NeuroscopeSorting data tests

CodyCBakerPhD:
Requires a small fix to the allowed metadata schema for the sorting interface to include more general ecephys metadata that might still be desired to be included in historical data where the original raw sources are no longer saved.


src/nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -25,18 +27,41 @@ def get_metadata_schema(self):
...
         metadata_schema["properties"]["Ecephys"]["properties"] = dict(
+            Device=dict(type="array", minItems=1, items={"$ref": "#/properties/Ecephys/properties/definitions/Device"}),
+            ElectrodeGroup=dict(
+                type="array", minItems=1, items={"$ref": "#/properties/Ecephys/properties/definitions/ElectrodeGroup"}
+            ),
+            Electrodes=dict(
+                type="array",
+                minItems=0,
+                renderForm=False,
+                items={"$ref": "#/properties/Ecephys/properties/definitions/Electrodes"},
```
h-mayorquin, line 42: What is this attribute `renderForm` for?

h-mayorquin (approved):
Looks good to me. 
How comes we did not have this attributes in the metadata schema? Were you getting an error without them? I don't understand your comment about historical data and raw sources very well.

bendichter:
@h-mayorquin, @CodyCBakerPhD is saying that even if we don't have the raw data, we might want to know what equipment was used, e.g. the device and electrode group information. He's calling this "historical" because we already have the spike sorted data, so we are unlikely to use the equipment metadata directly. 

`renderForm` is a field used to interact with our metadata form GUI. This tells the form not to render Electrodes. It can only be changed programatically.

bendichter (approved):

CodyCBakerPhD:
@h-mayorquin Yup, exactly what Ben said.

That said, I have never once seen a case where there wasn't at least LFP, so take that as you will. Still possibly at least in concept that they didn't even save that file, but did keep the spike trains.

Issue #449, opened 2022-04-03
[Feature]: tests take a long time to run locally


bendichter:
### What would you like to see added to nwb-conversion-tools?

Especially the ones in test_move_utils. Can we make these faster?

### What solution would you like?

Especially the ones in test_move_utils. Can we make these faster?

### Do you have any interest in helping implement the feature?

No.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/nwb-conversion-tools/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/nwb-conversion-tools/issues)?

CodyCBakerPhD:
Looks like `test_si.py` is the longest set of module tests to run, which we've been aware of for some time and @h-mayorquin was slowly refactoring them in a better fashion.

Also adding here - SpikeInterface recently started using a fancy module-based approach to triggers: https://github.com/SpikeInterface/spikeinterface/blob/master/.github/workflows/full-test.yml#L85-L118

Would be cool if we could do something similar, and changes to things like documentation or changes to differing modalities (an ophys interface vs an ecephys interface) need not trigger tests that only cover things untouched by those changes. Though for safety, should still have daily runs of the full test suite.

h-mayorquin:
Here is a table of the test duration as they stand now. Organized from the longest to the shortest:

As @CodyCBakerPhD says, some of the biggest culprit are the large (non unit) tests from the spikeinterface module. There are other large tests like `CnmfeSegmentationInterface` , `CellExplorerSortingInterface` and `test_pkl_interface`.

|     |   time | test                                                                                                                                                               | test_short_name                                                                                                                  |
|----:|-------:|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------|
|   0 |  59.21 | test_internals/test_si.py::TestExtractors::test_write_sorting                                                                                                      | TestExtractors::test_write_sorting                                                                                               |
|   1 |  27.27 | test_internals/test_si.py::TestExtractors::test_write_recording                                                                                                    | TestExtractors::test_write_recording                                                                                             |
[...]

PR #459, opened 2022-04-18
Warning reduction : remove numpy type warnings.

h-mayorquin:
## Motivation
As @bendichter requested I started looking into reducing the number of warnings we get while running our testing suite.

One of the low hanging fruits and the subject of this PI is the following deprecation in numpy:

https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

There are plenty of errors of this kind, they look like this:
```python
/Users/runner/work/nwb-conversion-tools/nwb-conversion-tools/src/nwb_conversion_tools/utils/types.py:14: 
DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. 
To silence this warning, use `float` by itself.
Doing this will not modify any behavior and is safe. 
If you specifically wanted the numpy scalar type, use `np.float64` here.
```

In this PR I just went through all the instances and changed the casting to the corresponding type. There are similar warnings in `roieextractors` and `spikeextractors` that should be modified as well.


h-mayorquin:
@CodyCBakerPhD is there a way of making the tests to run in the "automated changes" commit?

The tests are running OK for the previous commit before that one.

tests/test_internals/test_movie_interface.py
```diff
@@ -66,7 +66,7 @@ def get_metadata(self):
         return metadata
 
     def test_movie_starting_times(self):
-        starting_times = [np.float(np.random.randint(200)) for i in range(len(self.movie_files))]
+        starting_times = [float(np.random.randint(200)) for i in range(len(self.movie_files))]
```
bendichter, line 69: This is better because it removes the warning, but a better solution yet would be to use the vectorized numpy version:

```python
starting_times = np.random.randint(200, size=(len(self.movie_files), ) dtype=float)
```

bendichter: I don't think the performance would matter in this case though because it's just small test cases, so this PR is fine.

h-mayorquin: Yes, I thought about it as well.

I was thinking at revisiting these tests on detail at some point in the future as we have the issue:
#449

CodyCBakerPhD:
> @CodyCBakerPhD is there a way of making the tests to run in the "automated changes" commit?
> 
> The tests are running OK for the previous commit before that one.

Yeah, switching to pre-commit (like on the NWBInspector) will do that. Feel free to open a PR for that

CodyCBakerPhD (approved):

PR #460, opened 2022-04-18
Not overwriting channel name if recording object has the property

h-mayorquin:
While I was working with a `spikelgx` conversion I realize that our current implementation of `add_electrodes` overwrites the property of channel name if this is available in the recording object (it always assign it to `channel_ids.astype("str")`). 

I modified the logic so this does not happen and added a new test.

I also did other two things in this PR:
* corrected a small bug that was showing up in warnings.
* Split one of the unit tests in two as it was testing two different things in one function.
* Added more descriptive names to the unittests.

### DIFF
--- a/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
+++ b/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
@@ -306,9 +306,12 @@ def add_electrodes(
 
     # Channel name logic
     channel_ids = checked_recording.get_channel_ids()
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
> While I was working with a spikelgx conversion I realize that our current implementation of add_electrodes overwrites the property of channel name if this is available in the recording object (it always assign it to `channel_ids.astype("str")`).

@h-mayorquin Aha, good catch

CodyCBakerPhD (approved):

PR #461, opened 2022-04-18
[Testing]: add NeuroscopeLFP

CodyCBakerPhD:
Confirming with CI, but it appears as if the new SpikeInterface version has an issue finding the header.

If so, assigning @h-mayorquin to fixing this given familiarity with the format and neo.


CodyCBakerPhD:
Yep, confirmed.

Full traceback...

```
tests/test_on_data/test_gin_ecephys.py:79: in test_convert_lfp_to_nwb
    converter = TestConverter(source_data=dict(TestLFP=interface_kwargs))
src/nwb_conversion_tools/nwbconverter.py:66: in __init__
    for name, data_interface in self.data_interface_classes.items()
src/nwb_conversion_tools/nwbconverter.py:67: in <dictcomp>
    if name in source_data
src/nwb_conversion_tools/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py:251: in __init__
    super().__init__(file_path=file_path)
src/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py:25: in __init__
    self.recording_extractor = self.RX(**source_data)
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/spikeinterface/extractors/neoextractors/neuroscope.py:[38](https://github.com/catalystneuro/nwb-conversion-tools/runs/6063870215?check_suite_focus=true#step:23:38): in __init__
    NeoBaseRecordingExtractor.__init__(self, stream_id=stream_id, **neo_kwargs)
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/spikeinterface/extractors/neoextractors/neobaseextractor.py:26: in __init__
    _NeoBaseExtractor.__init__(self, **neo_kwargs)
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/spikeinterface/extractors/neoextractors/neobaseextractor.py:16: in __init__
    self.neo_reader.parse_header()
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/neo/rawio/baserawio.py:185: in parse_header
    self._parse_header()
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/neo/rawio/neuroscoperawio.py:[40](https://github.com/catalystneuro/nwb-conversion-tools/runs/6063870215?check_suite_focus=true#step:23:40): in _parse_header
    tree = ElementTree.parse(filename + '.xml')
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/xml/etree/ElementTree.py:1197: in parse
    tree.parse(source, parser)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <xml.etree.ElementTree.ElementTree object at 0x7f038a273550>
source = 'ephy_testing_data/neuroscope/dataset_1/YutaMouse[42](https://github.com/catalystneuro/nwb-conversion-tools/runs/6063870215?check_suite_focus=true#step:23:42)-1[51](https://github.com/catalystneuro/nwb-conversion-tools/runs/6063870215?check_suite_focus=true#step:23:51)117.eeg.xml'
parser = None

    def parse(self, source, parser=None):
        """Load external XML document into element tree.
    
        *source* is a file name or file object, *parser* is an optional parser
        instance that defaults to XMLParser.
    
        ParseError is raised if the parser fails to parse the document.
    
        Returns the root element of the given source document.
    
        """
        close_source = False
        if not hasattr(source, "read"):
>           source = open(source, "rb")
E           FileNotFoundError: [Errno 2] No such file or directory: 'ephy_testing_data/neuroscope/dataset_1/YutaMouse42-151117.eeg.xml'
```

Looks like (a) the auto-inferring header is not stripping the suffix and (b) extractor is not taking a direct `xml_path` argument, which we needed for various conversions anyway to cover cases where there are multiple `.xml` present.

h-mayorquin:
@CodyCBakerPhD 
So, the problem is a combination of two things.
I) The neoreader for neuroscope in `spikeinterface` only takes a file_path (not possible to specificy the xml_path). 
II) The only extensions that are supported are `.dat` and `.xml`.  (so not support of `.lpf` or `eeg`).

See:
https://github.com/NeuralEnsemble/python-neo/blob/5f8cdb7bddf16cda22d94225b83e636e2023fd10/neo/rawio/neuroscoperawio.py#L27-L40

That is, at the moment neoextractors does not support `.eeg` which is the file_path that you are passing.

bendichter:
If we want to get really creative we could use a softlink to add a file that has the expected name

h-mayorquin:
> If we want to get really creative we could use a softlink to add a file that has the expected name

Yes, that makes sense! Maybe the `pathlib` functionallity works correctly across different operating systems:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

That said, the docs of neo are not very optimistic about the future of neuroscope maintenance there:
https://github.com/NeuralEnsemble/python-neo/blob/7cf65eb0af48fa0c2b245da0c02fee04f63ad4f1/neo/rawio/neuroscoperawio.py#L10-L15

h-mayorquin:
> neo works best for extracting timeseries data. It's not as good for spike sorted data, so we should expect those to be implemented by spikeinterface sometimes.

I was not aware of this division of labor. Good to know.

CodyCBakerPhD:
> I) The neoreader for neuroscope in spikeinterface only takes a file_path (not possible to specificy the xml_path).

That is something that should be supported, either at the SpikeInterface level or the neo level, don't care which. There are many times where there are mismatched `.xml` paths with respect to the `.dat` path and explicit specification was necessary for proper reading of the header.


> II) The only extensions that are supported are .dat and .xml. (so not support of .lfp or eeg).

If they support the access logic for `.dat`, then they support `.eeg` as well, it's the same exact structure.


> That said, the docs of neo are not very optimistic about the future of neuroscope maintenance there:

No need to ask them to do it, can you get those PR started and point this PR's dependencies to those branches to try to get it working?

h-mayorquin:
I think that we can partition this in two issues:
1) Add support for formats other than `.dat`.
2) Add support for manual specification of the xml file.

I open a PR for the first issue. This can be solved completely at the neo level and should get the test here working:
https://github.com/NeuralEnsemble/python-neo/pull/1113

The second issue needs modifications both in neo and in spikeinterface as they modify the function signature. I start working on this now and push a PR as soon as we are done with 1.

h-mayorquin:
It would be good to get @samuelgarcia on the review so he is aware of the larger plans here once he is back from vacations.

h-mayorquin:
Should we close this and open issues for the remaining work?

CodyCBakerPhD:
@h-mayorquin If by 'close' you mean 'merge' - the only thing about pinning minimal requirements now to a neo dev branch is given their release cycles it will be quite some time before it actually goes through, which would prevent us from making a new release of NWBCT in the meantime if we so desired.

In that case I'd be fine leaving this until it at least gets merged to their main.

Commit hashes would be even better, and are you sure we really need it for the minimal reqs?

h-mayorquin:
Pinning the minimal requirements is not necessary for the basic tests. 

All right, so we leave this here then.

samuelgarcia:
Hi all,
I am back soon and will have a look and merge the neo PR.

h-mayorquin:
The PR in neo is ready:
https://github.com/NeuralEnsemble/python-neo/pull/1113

What remains is do the follow-up in `spikeinterface` and then propagate it here.

CodyCBakerPhD:
This PR is just waiting on the change to be  included in an official `neo` release, so this PR doesn't delay our release cycle once the new PyNWB is out.

h-mayorquin:
I will close this as the PR in the new repo in more or less the same state.

PR #464, opened 2022-04-18
Memmap data chunk iterator

bendichter:
## Motivation

Make it easy to create data chunk iterators for sliceable memmap objects. 

@CodyCBakerPhD, how do you feel about this name? It's technically for any *sliceable* object, which is more general. This is demonstrated in the test, which uses an in-memory numpy array instead of a memmapped one, but I named it memmap because I expect that to be the situation where this is used. I don't see much of a point in putting an in-memory array into this except to test it. Nevertheless, I think another viable alternative for name is `SliceableDataChunkIterator`


CodyCBakerPhD:
Indeed, I was looking at this branch over the weekend and thought it looked pretty nice - good to see the tests for it!

Is the progress bar displaying OK? It is *technically* possible to [add explicit tests](https://github.com/hdmf-dev/hdmf/blob/dev/tests/unit/utils_test/test_core_GenericDataChunkIterator.py#L240-L253) for it, though annoying to do so, and occasional differences are sometimes observed for different underlying data types. Would be fine considering it 'OK' if you can personally observe it working on your end.



> This is demonstrated in the test, which uses an in-memory numpy array instead of a memmapped one, but I named it memmap because I expect that to be the situation where this is used.  I don't see much of a point in putting an in-memory array into this except to test it.

A point would be to harness the better default setting of chunk shapes through an `H5DataIO` - but yes, there is no actual 'buffering' going on since the array is already in memory.



> @CodyCBakerPhD, how do you feel about this name? It's technically for any _sliceable_ object, which is more general. Nevertheless, I think another viable alternative for name is `SliceableDataChunkIterator`

That may be a better name if only to refer to how this applies to both `h5py.Dataset.data` as well as `np.memmap` objects, which are both sliceable but the former does not make any textual reference to the phrase 'memmap'.

CodyCBakerPhD (approved):

PR #465, opened 2022-04-18
Update spikelgx to use spikeinterface II proposal

h-mayorquin:
This should close #456 as I branched from there.

This is an alternative proposal instead of mixings or the aux function for __init__.
@bendichter commented that for the `spikeglx` case the electrical series should be stored in acquisition (as it comes directly from the device and it is not processed). In this implementation `SpikeGLXLFPInterface` inherits from `SpikeGLXRecordingInterface` modifying the metadata and its schema when appropriate. This achieves both the sought code reduction while at the same time has a simpler and clear inheritance hierarchy.

I eliminated the gin test for `spikelgx lfp` (as it relies on the fact that the lfp is stored in nwb.procesing module) and added the corresponding test back in the recording extractor part.


bendichter:
yeah, I'm happy with this approach. What do you think, @CodyCBakerPhD ?

CodyCBakerPhD:
> This PR got me thinking about the our assumption of equating "acquisition" with "raw". It is totally conceivable that acquisition process the " truth raw" data in a way that is not available even to the experimenter.

Yes, and this is such a case. There is an line to walk between what we consider to be the difference in any such situation. I simply would have appreciated this behavioral change to occur in a separate PR to one that isn't supposed to have any change to behavior at all.

h-mayorquin:
> > This PR got me thinking about the our assumption of equating "acquisition" with "raw". It is totally conceivable that acquisition process the " truth raw" data in a way that is not available even to the experimenter.
> 
> Yes, and this is such a case. There is a line to walk between what we consider to be the difference in any such situation. I simply would have appreciated this behavioral change to occur in a separate PR to one that isn't supposed to have any change to behavior at all.

@CodyCBakerPhD 
I think that your concern can be addressed by merging the branch from which this came first and the merging this. You will have a commit history that separates the feature addition from the behavioral change.

I also agree that our PR should be more atomic.

CodyCBakerPhD:
> I think that your concern can be addressed by merging the branch from which this came first and the merging this. You will have a commit history that separates the feature addition from the behavioral change.

Except that the first PR was not approved due to duplicated code, so could not be merged. A simple temporary auxiliary function for handling that ought to have taken only a couple of minutes to write and debug.

I could have merged this PR into that one, but then we would have the same exact situation as we do here.

Really, it should just be a chain of PRs making minimal changes with each iteration, and documenting which if any result in specific behavioral changes. Higher merge rate of smaller changelogs is much easier to review.

h-mayorquin:
@CodyCBakerPhD 
I think just keeping this line:
https://github.com/catalystneuro/nwb-conversion-tools/blob/673410a79402b31a919e0e13e2bf7ed84732d15c/src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py#L135

as:
```python
        conversion_options = dict(write_as="lfp", es_key="ElectricalSeries_lfp", stub_test=False)
```

Would have produced got us the introduction of the feature (spikeinterface backend) without the change of behavior (writing to acquisition instead of processing). What do you think? 
I am leaving this here in case we need to turn that back.

CodyCBakerPhD:
> conversion_options = dict(write_as="lfp", es_key="ElectricalSeries_lfp", stub_test=False)

Indeed, that would restore the previous behavior since that is, granted, essentially all the LFP run conversion step does since you hard-coded the metadata step to match.

PR #466, opened 2022-04-19
Fix group creation

h-mayorquin:
## Motivation
The function `add_electrode_groups` did no match the behavior of `add_electrodes` in the sense that the former used the property "group" to decide the name for group creation but the latter gave primacy / priority to the property "group_name". This PR is so they are aligned in which property is used to decide the name of the electrode groups.

Plus, I did some work for reducing some of the warnings.
1) Added a starting time to each to all of our tests.
2) Remove `nwbfile.modules` which is now deprecated. 

I also reduced the number of frames used in `test_si`. I don't think that there is any use for so many frames just for the test there.

### DIFF
--- a/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
+++ b/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
@@ -172,14 +172,20 @@ def add_electrode_groups(recording: SpikeInterfaceRecording, nwbfile: pynwb.NWBF
         metadata = dict()
     if "Ecephys" not in metadata:
         metadata["Ecephys"] = dict()
+    
+    if "group_name" in checked_recording.get_property_keys(): 
+        group_names = np.unique(checked_recording.get_property("group_name"))
+    else:
+        group_names = np.unique(checked_recording.get_channel_groups()).astype("str")
+    
     defaults = [
         dict(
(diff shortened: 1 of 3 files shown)


src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -172,14 +172,20 @@ def add_electrode_groups(recording: SpikeInterfaceRecording, nwbfile: pynwb.NWBF
         metadata = dict()
     if "Ecephys" not in metadata:
         metadata["Ecephys"] = dict()
+
+    if "group_name" in checked_recording.get_property_keys():
+        group_names = np.unique(checked_recording.get_property("group_name"))
+    else:
+        group_names = np.unique(checked_recording.get_channel_groups()).astype("str", copy=False)
```
bendichter, line 179: why call unique?

h-mayorquin: The function returns the name of the group for each of the channels. But we only need to create a group object for each of the unique names.

bendichter: ah, ok, I see

CodyCBakerPhD (approved):

PR #468, opened 2022-04-20
Update default conversion options instead of overwriting for run conversion

h-mayorquin:
## Motivation
Right now, when we run a conversion (with `NwbConverter` class) and we pass some `conversion_options` this overrides all of them:

https://github.com/catalystneuro/nwb-conversion-tools/blob/673410a79402b31a919e0e13e2bf7ed84732d15c/src/nwb_conversion_tools/nwbconverter.py#L149-L150

The current PR solves this by using `dict_deep_update` so the defaults conversion options defined at the interface level are respected at run time.

This PR is a follow up of #465 where we are using the `get_conversion_options` to hard code how and where the recording extractor data should be written. 

Currently, we only define `get_conversion_options` for spikeglx and in our `TutorialRecordingExtractor` but if we want to use this option further we will need something like this.

### DIFF
--- a/src/nwb_conversion_tools/nwbconverter.py
+++ b/src/nwb_conversion_tools/nwbconverter.py
@@ -146,10 +146,15 @@ class NWBConverter:
 
         if metadata is None:
             metadata = self.get_metadata()
-        if conversion_options is None:
-            conversion_options = self.get_conversion_options()
         self.validate_metadata(metadata=metadata)
+
+        if conversion_options is None:
+            conversion_options = dict()
+        default_conversion_options = self.get_conversion_options()
+        conversion_options = dict_deep_update(default_conversion_options, conversion_options)
+
         self.validate_conversion_options(conversion_options=conversion_options)
+
         if save_to_file:
             load_kwargs = dict(path=nwbfile_path)
             if nwbfile_path is None:
--- a/src/nwb_conversion_tools/nwbconverter.py
+++ b/src/nwb_conversion_tools/nwbconverter.py
@@ -151,9 +151,9 @@ class NWBConverter:
         if conversion_options is None:
             conversion_options = dict()
         default_conversion_options = self.get_conversion_options()
-        conversion_options = dict_deep_update(default_conversion_options, conversion_options)
+        conversion_options_to_run = dict_deep_update(default_conversion_options, conversion_options)
 
-        self.validate_conversion_options(conversion_options=conversion_options)
+        self.validate_conversion_options(conversion_options=conversion_options_to_run)
 
         if save_to_file:
             load_kwargs = dict(path=nwbfile_path)
@@ -169,7 +169,9 @@ class NWBConverter:
                 elif nwbfile is None:
                     nwbfile = make_nwbfile_from_metadata(metadata=metadata)
                 for interface_name, data_interface in self.data_interface_objects.items():
-                    data_interface.run_conversion(nwbfile, metadata, **conversion_options.get(interface_name, dict()))
+                    data_interface.run_conversion(
+                        nwbfile, metadata, **conversion_options_to_run.get(interface_name, dict())
+                    )
                 io.write(nwbfile)
             print(f"NWB file saved at {nwbfile_path}!")
         else:
--- a/src/nwb_conversion_tools/nwbconverter.py
+++ b/src/nwb_conversion_tools/nwbconverter.py
@@ -152,7 +152,6 @@ class NWBConverter:
             conversion_options = dict()
         default_conversion_options = self.get_conversion_options()
         conversion_options_to_run = dict_deep_update(default_conversion_options, conversion_options)
-
         self.validate_conversion_options(conversion_options=conversion_options_to_run)
 
         if save_to_file:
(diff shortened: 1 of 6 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
Interestingly, we do have several test functions out there that use `conversion_options`, but I suppose they are all single-interface unit tests whereas this behavior would require a 2+ interface integration test, right?

CodyCBakerPhD:
@h-mayorquin Also can you look into if https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/nwbconverter.py#L95-L100 and https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/basedatainterface.py#L38-L40 are even needed at all? Seems like a round-about way of double-specifying default `kwargs` for all `run_conversion` functions.

Ideally the `conversion_options` dict would just be the minimal subset of override commands, though maybe it needs to load the rest of the options for the schema validation step?

h-mayorquin:
> Interestingly, we do have several test functions out there that use `conversion_options`, but I suppose they are all single-interface unit tests whereas this behavior would require a 2+ interface integration test, right?

I think the core of why this has been flying under the radar is because no interface relies on the `get_conversion_options` for anything that we are testing. 

That said, you are also right about the two interfaces in the specific instance where this bug surfaced. Due to the new changes in spikeglx data interface (#465) the two of them crash when their `get_conversion_options` are overwritten at run time. This is because when force into default behavior both of them try to write to acquisition with the same name and the program fails.

CodyCBakerPhD:
> That said, you are also right about the two interfaces in the specific instance where this bug surfaced. Due to the new changes in spikeglx data interface (https://github.com/catalystneuro/nwb-conversion-tools/pull/465) the two of them crash when their get_conversion_options are overwritten at run time. This is because when force into default behavior both of them try to write to acquisition with the same name and the program fails.

Nice catch, be sure to assign unique names by default for that interface, then.

Also I'd add it to your new To-Do list on https://github.com/orgs/catalystneuro/projects/3/views/1 to add a battery of tests for `conversion_options` support.

h-mayorquin:
> @h-mayorquin Also can you look into if https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/nwbconverter.py#L95-L100 and https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/basedatainterface.py#L38-L40 are even needed at all? Seems like a round-about way of double-specifying default `kwargs` for all `run_conversion` functions.
> 
> Ideally the `conversion_options` dict would just be the minimal subset of override commands, though maybe it needs to load the rest of the options for the schema validation step?

In my view, the second definition in `basedatainterface.py` is important as it ensures that the method always exist for all interfaces and returns an empty dict as default.

The first method is just a convenience for gathering all the conversion options for each interface in a hierarchical dict. Maybe not needed I agree but I also don't think is too much code.

More importantly, I think that the both the metadata and the conversion options should be passed and validated in the initialization of the class. Also, we should make them both attributes of `NWBConverter`. Doing this will allow us to build tests for the validation of both the metadata and the conversion options without the need of calling `run_conversion` and give the latter function a more clear responsibility.

src/nwb_conversion_tools/nwbconverter.py
```diff
@@ -146,10 +146,15 @@ def run_conversion(
...
             metadata = self.get_metadata()
-        if conversion_options is None:
-            conversion_options = self.get_conversion_options()
         self.validate_metadata(metadata=metadata)
+
+        if conversion_options is None:
+            conversion_options = dict()
+        default_conversion_options = self.get_conversion_options()
+        conversion_options = dict_deep_update(default_conversion_options, conversion_options)
+
```
bendichter, line 155: is this ever deep? Would a simple `dict.update` work?

h-mayorquin: It is just one level deep. One for each of the interfaces so I think is needed. Just to be specific, this how the output of `get_conversion_options`  looks for the current Giocomo conversion:
```python
{'SpikeGLXRecording': {'write_as': 'raw', 'es_key': 'ElectricalSeries_raw', 'stub_test': False}, 'SpikeGLXLFP': {'write_as': 'raw', 'es_key': 'ElectricalSeries_lfp', 'stub_test': False}, 'Behavior': {}}
```

A simple loop and update for every interface will do the job. Maybe that's preferable?

bendichter: oh right, one for each of the interfaces. I think the way you have it is fine

bendichter (approved):

src/nwb_conversion_tools/nwbconverter.py
CodyCBakerPhD: Should have depth 2 at the NWBConverter level (depth 1 at interface level)

```python
my_conversion_options = dict(
    Interface1=dict(foo=bar),
    Interface2=dict(this=that),
)
```

PR #472, opened 2022-04-21
Separate add unit table property from add waveforms

h-mayorquin:
## Motivation
This is the first PR to break down all the code changes in #450 as requested by @CodyCBakerPhD .

It just adds a different function called `add_wave_forms_to_units_table` that works separately from `add_units`.


.gitignore
```diff
@@ -121,3 +121,4 @@ venv.bak/
 
 # NWB
 .nwb
+.pre-commit-config.yaml
```
CodyCBakerPhD, line 124: Were you able to get this setup separately to replace the main workflow?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1264,15 +1326,23 @@ def write_sorting(
...
         "processing",
     ], f"Argument write_as ({write_as}) should be one of 'units' or 'processing'!"
+
+    write_units_table_in_processing_module = True
+    units_table_name = units_name
+    units_table_description = units_description
+
     if write_as == "units":
         assert units_name == "units", "When writing to the nwbfile.units table, the name of the table must be 'units'!"
+        write_units_table_in_processing_module = False
```
CodyCBakerPhD, line 1344: You can do the setting of the boolean mediator in one line, to reduce 4 lines and a break to 1 line, no break

```python
   write_units_table_in_processing_module = False if write_as == "units" else True
```

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1290,23 +1360,36 @@ def write_sorting(
...
                 units_name=units_name,
                 units_description=units_description,
             )
+            add_wave_forms_to_units_table(
+                sorting=sorting,
+                nwbfile=nwbfile,
+                skip_features=skip_features,
+                units_table_name=units_table_name,
+                unit_table_description=units_table_description,
+                write_units_table_in_processing_module=write_units_table_in_processing_module,
```
CodyCBakerPhD, line 1374: I think it's fine/readable to remove the above lines and do the kwarg aliasing here

```python
    units_table_name = units_name
    units_table_description = units_description
```

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1290,23 +1360,36 @@ def write_sorting(
...
             write_as=write_as,
             units_name=units_name,
-            units_description=units_description,
+        )
+        add_wave_forms_to_units_table(
+            sorting=sorting,
+            nwbfile=nwbfile,
+            skip_features=skip_features,
+            units_table_name=units_table_name,
+            unit_table_description=units_table_description,
```
CodyCBakerPhD, line 1392: and here, as well

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1138,6 +1135,83 @@ def add_units(
...
+    elif write_as == "processing":
+        ecephys_mod = get_module(
+            nwbfile=nwbfile,
+            name="ecephys",
+            description="Intermediate data from extracellular electrophysiology recordings, e.g., LFP.",
+        )
+        ecephys_mod.add(units_table)
+
+
+def add_wave_forms_to_units_table(
```
CodyCBakerPhD, line 1153: wave_form should probably be one word, `waveform` for consistency with [the SpikeInterface WaveformExtractor](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/waveform_extractor.py)

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1138,6 +1135,83 @@ def add_units(
...
+            description="Intermediate data from extracellular electrophysiology recordings, e.g., LFP.",
+        )
+        if units_table_name not in ecephys_mod.data_interfaces:
+            units_table = pynwb.misc.Units(name=units_table_name, description=unit_table_description)
+            ecephys_mod.add(units_table)
+
+        units_table = ecephys_mod[units_table_name]
+    else:
+        if nwbfile.units is None:
+            nwbfile.units = pynwb.misc.Units(name=units_table_name, description=unit_table_description)
```
CodyCBakerPhD, line 1212: So the logic here is to write the units table to the NWBFile in memory first, then modify the reference to that? 

Also, how would this modular setup of writing handle a case where I try to write (a) waveforms before writing spike times and (b) can it detect mismatches in shapes between the spiketimes for a unit and the shape of the waveform array?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1138,6 +1135,83 @@ def add_units(
...
+        units_table = nwbfile.units
+
     if isinstance(sorting, SortingExtractor):
         all_features = set()
```
CodyCBakerPhD, line 1216: Also can you add a commented placeholder section for the `else` condition of this block to indicate long-term plans to add WaveformExtractor support?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1138,6 +1135,83 @@ def add_units(
...
+    write_units_table_in_processing_module: bool (optional, defaults to False)
+        How to save the units table in the nwb file.
+        - True will save it to the processing module to serve as a historical provenance for the official table.
+        - False will save it to the official NWBFile.Units position; recommended only for the final form of the data.
+    """
+    if isinstance(sorting, SortingExtractor):
+        sampling_frequency = sorting.get_sampling_frequency()
+        if sampling_frequency is None:
+            raise ValueError("Writing a SortingExtractor to an NWBFile requires a known sampling frequency!")
+        checked_sorting = OldToNewSorting(oldapi_sorting_extractor=sorting)
```
CodyCBakerPhD, line 1193: Wait a sec, won't doing this here prevent us from entering 1215 below? In which case, doesn't this make it impossible to write waveforms that are stored on a spikeextractor object as features?

.gitignore
h-mayorquin: It is working locally for me.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Totally agree, good point!

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Done.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Good!

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Yes, that's correct. That was the logic as it was before (see 'add_units' function) and I think it plays well with the requirement of choosing where to write the units table ( e.g in a processing module vs the standard place for the canonical table).

Concerning your questions:
a) This does not work right now. As mentioned in #450, `add_units`, in its current state does not support re-writing of any sort. After the next PR in this series, it should handle it OK.
b) I don't know what do you mean here, I have not modified this function at all so I don't know where this is done, is this related to row mismatch across multiple writings? Maybe if you point me how this problem was solved in the previous version I can say whether it would work with the new logic.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: No, the check that you are talking about is done in the original sorting extractor and not in the checked_recording:
```python
if isinstance(sorting, SortingExtractor):
```

In fact, I am removing the checked_extractor logic here. At the moment, it does not make sense.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Done. Feel free to add something else that captures the current state of affairs.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: b) probably was not solved in previous version. Just occurred to me with the new append feature... the dimensionality of waveforms for a single unit should be `(number of spike times, number of frames, number of channels)`. If, going forward, you attempt to add a waveform column after writing spike times, the first dimension should match or else it isn't clear which spike time indices you are subsetting.

CodyCBakerPhD: > That was the logic as it was before (see 'add_units' function) 

The logic on main:
https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L1199-L1210 adds the units table object to the nwbfile as the very final step (even after writing waveforms).

I think, if you want to split out the waveform step into a separate function altogether, it would make considerably more sense to have it apply functionally to an input `UnitsTable` object, rather than an `nwbfile` where it then takes all those extra arguments to determine 'which' table within the file to act on. This would also simplify the logic of having to create a table in the event it doesn't exist, which seems a very odd way to go about writing waveforms (harkening to point (a) above).

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Nah that's good, thanks.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Aha, yes my bad. Thanks for clarifying.

h-mayorquin: Yes, that confused me the first time btw.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Concerning a) I misunderstood you. Yes, this only keeps backwards compatibility with the previous framework of filling the units table and then immediately add the waveforms. To do this, the proper reference written before by add_units) is fetched and the waveforms are added to this table. 

Concerning the second point in the last response.
Yes, maybe that would make sense. I also thought that it would make sense not only for the `add_waforms` table but also for the `add_units one`.

That said, I think that belongs to a separate PR which we can discuss once we finish refactoring the units_table. The current scope of this is just to separate the two functions while giving backwards compatibility for the framework we had before. 

Once we have an `add_units` function that supports multiple writes we can discuss how the cases that you mention fits into that framework, how they will interact with the refactored add_units function and device a PR specifically to cover those cases that you mention with proper testing.

h-mayorquin: > > That was the logic as it was before (see 'add_units' function)
> 
> The logic on main: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L1199-L1210 adds the units table object to the nwbfile as the very final step (even after writing waveforms).
> 
> I think, if you want to split out the waveform step into a separate function altogether, it would make considerably more sense to have it apply functionally to an input `UnitsTable` object, rather than an `nwbfile` where it then takes all those extra arguments to determine 'which' table within the file to act on. This would also simplify the logic of having to create a table in the event it doesn't exist, which seems a very odd way to go about writing waveforms (harkening to point (a) above).

@CodyCBakerPhD 
Maybe the disagreement comes from here? 
The function has to decide where to write (units vs processing) at the end because it never has to overwrite (if the units table or the processing table already exists it exits without doing anything):

https://github.com/catalystneuro/nwb-conversion-tools/blob/177733629f1cea1a885191f588eb37f1b28c9396/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L1114

As we separate the `add_waveforms` function this assumption **can not* be hold to be true anymore (as previously `add_units` modify the table), so the logic of fetching the place where the table is to be written needs to be added at the beginning as well.

If I understand you well, you strongly believe that this should not be done inside function but maybe I don't understand you well.

CodyCBakerPhD: OK, here's the clearest plan I can layout for the API-level setup with an eye towards the future additions.

For API:

A WaveformExtractor is essentially a container for a RecordingExtractor and SortingExtractor. Though I can also see cases for it containing direct waveform snippets instead of a recording in cases where the original format didn't keep raw data but did record pre-extracted snippets.

As such, we will want a `write_waveforms(...)` function that similar to `write_recording` and `write_sorting` acts on a file level and is mostly a router to a helper function. That helper function *could* be called `add_waveforms(...)` and it *could* include the kind of argument input setup you have now; my point is however, the design of *that* `add_waveforms` function I'm talking about right now should do the following:

```python
def add_waveforms(...):
    # write recording metadata
    # write the sorting extractor contents from the one contained in the WaveformExtractor, most importantly the spike times
    # then add the waveforms for each of the spike times on that table
```

Now, I am proposing that the logic of the commented blocks would be easiest to implement as follows:

```python
units_table = add_units(...)
add_waveforms_to_units_table(units_table=units_table)
```

Basically, the logic as you have it now, would add the waveform column to a units table independently of the spike times, which opens the door to a long series of complications due to shape mismatch assertions between the two that are already handled by the code being moved. There's that complication on top of all of  the excess logic and input arguments for creating a brand new units table in the event there is no existing table, which absolutely should not happen here.

What I am suggesting, then, as clearly as I can state it, is for you to change what you are currently calling `add_waveforms` to a private function `_add_waveforms_to_table(units_table)` that is still called during `add_units(, write_waveforms=True)` for the case when the input sorting object is an old SortingExtractor that can have unit spike features. Then in a future PR we can add a proper `add_waveforms` that truly behaves as expected according to the outline above.

h-mayorquin: I get you now, I did not get what you are saying in the last paragraph from your previous suggestion. 

Having that private method called from inside the function does seem like a better solution to me that is even less of a change.

h-mayorquin: All right, added.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Nice and simple first PR, thanks for the work splitting it.

PR #473, opened 2022-04-21
pin version of roiextractors

h-mayorquin:
Testing pinning exact version of roi extractors

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -16,7 +16,7 @@ opencv-python==4.5.1.48
 spikeextractors==0.9.9
 spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git
 neo==0.10.2
-roiextractors @ git+https://github.com/catalystneuro/roiextractors.git
+roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@10265be4a1eaece7461296a2877cc11c84bf87dc
 pyintan==0.3.0
 pyopenephys==1.1.2
 sonpy>=1.7.1
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -14,7 +14,7 @@ psutil==5.8.0
 lxml==4.6.5
 opencv-python==4.5.1.48
 spikeextractors==0.9.9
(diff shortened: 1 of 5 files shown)


requirements-full.txt
```diff
@@ -14,9 +14,9 @@ psutil==5.8.0
 lxml==4.6.5
 opencv-python==4.5.1.48
 spikeextractors==0.9.9
-spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git
+spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@12b7863684b705e3d0ae8165bb3009143c70530c
 neo==0.10.2
-roiextractors @ git+https://github.com/catalystneuro/roiextractors.git
+roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@10265be4a1eaece7461296a2877cc11c84bf87dc
```
bendichter, line 19: it would be better to cut a release on pypi and pin to that

CodyCBakerPhD: For ROIExtractors we can control that of course, but not for SpikeInterface

CodyCBakerPhD: We're mostly debugging the terrible MacOS+Py3.8 combination of git installation right now

bendichter: what's terrible about it?

CodyCBakerPhD: Producing unexpected behavior never seen before unique to that combination of os and py version, we think we may have pinned it down

CodyCBakerPhD: Can confirm fixed, all Mac CI are past that installation point now

CodyCBakerPhD (approved):

CodyCBakerPhD:
Thanks @h-mayorquin 

This PR

- cleans up numpy float casting deprecation warnings coming from roiextractors
- also reveals behavior of `pip install -e .[full]` occurring after initial `pip install -e .` to not overwrite package versions, will think on how best to deal with that
- will cut a new release of roiextractors soon, looking for other cleanup warnings before then

PR #474, opened 2022-04-21
add default values to run_conversion 

bendichter:
## Motivation

add default values to run_conversion so that you can run it without inputting an NWBFile


CodyCBakerPhD:
Any harm in setting this at the BaseDataInterface level? I'm still working on some logic for allowing automatic NWBFile creation and saving for this similar to how the NWBConverter behaves.

CodyCBakerPhD:
Also, if that ghost failure from the `gin_behavior` tests on MacOS+Py3.8 comes back on this CI I wouldn't worry about it with respect to this PR. Thought we fixed it in https://github.com/catalystneuro/nwb-conversion-tools/runs/6115687975?check_suite_focus=true (notice all checks passing) but apparently it's still hanging around somewhere in there, we'll tackle it more tomorrow.

CodyCBakerPhD (approved):

PR #475, opened 2022-04-21
[Feature]: add verbosity control to main converter

CodyCBakerPhD:
While working on latest conversion project, the constant output from the main converter at each iteration was interfering with the normal in-place behavior of tqdm. This PR offers simple boolean control over whether statements like `"Source data is valid!"` show up during conversion.

In the code, the `source_data` related functions need the different structure because the main function is a classmethod (I don't recall why) that would attempt to otherwise reference the verbosity parameter prior to or independent of class creation.


h-mayorquin:
This is good, we should not print messages when things go well unless required to do so. I swear I remembered and issue about logging that I could not find.

src/nwb_conversion_tools/nwbconverter.py
```diff
@@ -53,14 +53,14 @@ def get_conversion_options_schema(cls):
         return conversion_options_schema
 
     @classmethod
-    def validate_source(cls, source_data):
+    def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
```
h-mayorquin, line 56: Should not we make this false by default?

bendichter:
getting the source schema needs to be a class method but I don't remember if there was a good reason for making validate_source_schema a class method. Doesn't this break is as a class method? If we are going to do that we might was well turn it into a normal instance method.

CodyCBakerPhD:
> P.D. I swear I remembered and issue about logging that I could not find.

https://github.com/catalystneuro/nwb-conversion-tools/issues/162, long-term low-priority

src/nwb_conversion_tools/nwbconverter.py
CodyCBakerPhD: I'm keeping it True for now just for backcompatability

CodyCBakerPhD:
> getting the source schema needs to be a class method but I don't remember if there was a good reason for making validate_source_schema a class method. Doesn't this break is as a class method? If we are going to do that we might was well turn it into a normal instance method.

The workaround to avoid breaking is to add verbose to both the instance `__init__` and to that class method (it's the only class method that has a print statement in it). Otherwise, yeah it didn't work since there was no `cls.verbose` without instantiation.

CodyCBakerPhD:
> This is good, we should not print messages when things go well unless required to do so. Maybe once we add logging capabilities to the conversions we should drop this type of verbose entirely?

Also, just to be clear, I don't mind our default verbosity in general, it can be nice to know that various steps are passing especially when only running a conversion for a single file (and good to know when any given file is done when running on a bunch of files).

h-mayorquin:
> > This is good, we should not print messages when things go well unless required to do so. Maybe once we add logging capabilities to the conversions we should drop this type of verbose entirely?
> 
> Also, just to be clear, I don't mind our default verbosity in general, it can be nice to know that various steps are passing especially when only running a conversion for a single file (and good to know when any given file is done when running on a bunch of files).
> 
> In general it's just good to be able to turn it off and on whenever you do or don't want it for some reason or another.

Yeah, I wonder, in general I prefer programs not telling me when things go as expected. I rather have verbosity as an option for debugging or when I want to be careful. It is a minor issue though.

CodyCBakerPhD:
> Yeah, I wonder, in general I prefer programs not telling me when things go as expected. I rather have verbosity as an option for debugging or when I want to be careful. It is a minor issue though.

Feel free to bring up the behavioral change of setting default verbosity to False as an Issue or in our meeting tomorrow, then. Our default has always been True, so that's why it's set to that in this PR.

bendichter (approved):

PR #477, opened 2022-04-25
Add device metadata for spikeglx interface

h-mayorquin:
## Motivation

The recording extractor for spikeglx already provides some metadata that we are not using: 1) the name of the device and some important information are included in the meta file, 2) the geometry / shape of the electrodes is included in the probe which can be built from the meta file, 3) the group structure for multi-shank electrodes. 

This PR adds the following metadata automatically:
* The group name for each electrode.
* Device for recordings with some relevant information in description.
* Electrodes geometry / contact shapes as discussed with @bendichter .

I am adding this as a draft as I don't know if the metadata descriptions are appropriate and it would be good to discuss.

## How to test the behavior?
I added some tests for the contact shapes.


src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -107,6 +122,37 @@ def get_conversion_options(self):
...
+        imDatBsc_pn = meta.get("imDatBsc_pn", "no base station part number")
+
+        description = (
+            "Imec device \n"
+            f"probe type = {probe_type} \n"
+            f"probe description = {probe_type_description} \n"
+            f"flex part number = {flex_part_number} \n"
+            f"base station connected part number {imDatBsc_pn} \n"
+        )
+
```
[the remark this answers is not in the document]
h-mayorquin: You mean that this should be a string that looks like json format or the field actually can take json? 
I guess the former, right?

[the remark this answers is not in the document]
bendichter: > or the field actually can take json

What do you mean by that?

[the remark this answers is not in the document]
h-mayorquin: I was confused, I did not know that the output of this:

```python
 import json
json_file = json.dumps(['foo', {'bar': ('baz', None, 1.0, 2)}])
type(json_file)
```
is str 

I was thinking that python had a 'json' object type (as there is a datetime type for dates that we can write as strings)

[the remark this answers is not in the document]
bendichter: oh, yes I meant a string that is json format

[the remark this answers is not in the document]
h-mayorquin: All right, the latest commit has that.

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -107,6 +123,36 @@ def get_conversion_options(self):
...
+        dict
+            a dict containing the metadata necessary for creating the device
+        """
+
+        meta = self.meta
+
+        probe_type = str(meta.get("imDatPrb_type", "no probe type"))
+        probe_type_to_probe_description = {"0": "NP1.0", "21": "NP2.0(1-shank)", "24": "NP2.0(4-shank)"}
+        probe_type_description = probe_type_to_probe_description.get(probe_type, "no probe description")
+
```
bendichter, line 141: If these fields are missing I'd rather just not populate them over filling them with filler data like this

bendichter (changes requested):

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
h-mayorquin: Yes, in the latest version they are not added to the dict structure unless they are not available. 

I also added a test for the device metadata.

src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -107,6 +123,36 @@ def get_conversion_options(self):
...
+            metadata_dict.update(probe_type=probe_type, probe_type_description=probe_type_description)
+
+        if "imDatFx_pn" in self.meta:
+            metadata_dict.update(flex_part_number=meta["imDatFx_pn"])
+
+        if "imDatBsc_pn" in self.meta:
+            metadata_dict.update(connected_base_station_part_number=meta["imDatBsc_pn"])
+
+        metadata_json = json.dumps(metadata_dict)
+        description_string = f"Imec device - metadata = {metadata_json}"
```
bendichter, line 151:
```suggestion
        description_string = metadata_json
```

would you mind just having this entire field be valid json?

h-mayorquin: Yes.

There is the following edge case though. when no fields are found, the line as you suggest would return description="{}", what do you think would be good to have in that case?

bendichter: I don't know, in that case "no description" I guess

h-mayorquin: All right, this is done.

h-mayorquin:
Is there anything missing here @bendichter?
I implemented the requested changes but I realized that they are not marked as done yet.

bendichter (approved):

PR #481, opened 2022-04-26
Added empty dict as root value to recursion in `dict_deep_update` to solve warnings.

h-mayorquin:
This is PR aims to address a type of warnings that we consistently get into the CI:

```python
  /home/runner/work/nwb-conversion-tools/nwb-conversion-tools/src/nwb_conversion_tools/utils/dict.py:174: UserWarning: input to update should be a dict, returning output
    warnings.warn("input to update should be a dict, returning output")
```
The reason for this is that the current implementation of `dict_deep_update` returns the value None when a key to update is not available:
https://github.com/catalystneuro/nwb-conversion-tools/blob/ffa6fb81dfdfd1797a790159d82f47a57ebc050b/src/nwb_conversion_tools/utils/dict.py#L180

Coincidentally, this also happens when the recursion algorithm hits the root of the stack and triggers the warnings at the start of the function:
https://github.com/catalystneuro/nwb-conversion-tools/blob/ffa6fb81dfdfd1797a790159d82f47a57ebc050b/src/nwb_conversion_tools/utils/dict.py#L173-L175

So, every time that we try to update a non-existent key of a metadata or source schema with a dict, this error is triggered, see example now:
```python
In [1]: from nwb_conversion_tools.utils.dict import dict_deep_update

In [2]: x = dict(x='2', y='3')

In [3]: z = dict(z=dict(z_deep='4'))

In [4]: dict_deep_update(x, z)
/home/heberto/development/nwb-conversion-tools/src/nwb_conversion_tools/utils/dict.py:174: UserWarning: input to update should be a dict, returning output
  warnings.warn("input to update should be a dict, returning output")
Out[4]: {'x': '2', 'y': '3', 'z': {'z_deep': '4'}}
```

The solution, is to return an empty dict when a key is not find, this works as an stop sign to indicate the root of the recursion while at the same time keeping the same functionality that the algorithm has now. I think the semantics of this are more clear.

Also, in this PR I took the liberty of doing some refactoring of the variables names. The principle there is to make them more specific and indicative of content than they were before.


src/nwb_conversion_tools/utils/dict.py
```diff
@@ -170,17 +170,31 @@ def dict_deep_update(
...
-            d[k] = dict_deep_update(d.get(k, None), v, append_list=append_list, remove_repeats=remove_repeats)
-        elif append_list and isinstance(v, list):
-            for vv in v:
-                d[k] = append_replace_dict_in_list(d.get(k, []), vv, compare_key, list_dict_deep_update, remove_repeats)
+        dict_to_update = deepcopy(dict_to_update)
+
+    for key_to_update, update_values in dict_with_update_values.items():
+        # Update with a dict like object is recursive until an empty dict is found.
+        if isinstance(update_values, collections.abc.Mapping):
+            sub_dict_to_update = dict_to_update.get(key_to_update, dict())
```
CodyCBakerPhD, line 184: The thing is though, without a schema available to reference for that data type the value of this key *ought* to be,  just assuming that it's a dictionary may be incorrect. What if the `key_to_update` is `Elecetrodes` within `Ecephys`? Then it should be a list of dicts instead.

Not to mentioned, a `defaultdict(intented_data_type)` would also perhaps be the better approach if you did know what data type it was beforehand.

CodyCBakerPhD: Ahh, but I guess you know from the above `if` checks that the intended update value is either a dict or a list for this case...

h-mayorquin: In the case of  `key_to_update=Electrodes` the values should be a list. These will not enter that specific if condition but it will enter the one just below:

CodyCBakerPhD:
Oky doky, makes sense to me. In the even the value to update is a dictionary, but the key doesn't exist at that level in the original dictionary, make an empty dictionary so that it can be updated.

CodyCBakerPhD (approved):

h-mayorquin:
The number of warnings decrease from 464 warnings to 344  in the CI test of python 3.9  (from the circa 600 we had at some point).

h-mayorquin:
> Just trying to think through this VERY carefully, are we sure that will not change functionality at all in any situation?

Yes, that's important. I think I should write my thoughts here as a record and in case that we have to get back know what went wrong.

This is the argument for why this should not change the functionality: 

This change only affects the functionality when the `dict_with_update_values` has a key that is not available in the `dict_to_update` (that is the get statement in line [180](https://github.com/catalystneuro/nwb-conversion-tools/blob/ffa6fb81dfdfd1797a790159d82f47a57ebc050b/src/nwb_conversion_tools/utils/dict.py#L180)) . Let's discuss what happens in the old and the new version of the functions:

1) current version in main: this returns `None`, that is, the `sub_dict_to_update` becomes None.
2) proposed version in this PR: this returns `dict()`, that is, the `sub_dict_to_update` becomes `dict()`.

Now, in both cases, what happens is that a new recursion instance of `dict_deep_update` is to be run with the specific values of the `dict_with_update_values` which should be a dict and becomes `sub_dict_with_update_values`. That is, we run the next recursion which is:

1) current version in main: `dict_deep_update(None, sub_dict_with_update_values)` -> this triggers the warnings and the following code is run:
https://github.com/catalystneuro/nwb-conversion-tools/blob/ffa6fb81dfdfd1797a790159d82f47a57ebc050b/src/nwb_conversion_tools/utils/dict.py#L173-L175
This returns the `sub_dict_with_update_values` and assigns them to the key.  This means that the result is: 
```python
dict_to_update[key_existing_only_in_dict_with_values_to_update] = dict_with_values_to_update[key_existing_only_in_dict_with_values_to_update]` 
```
2) proposed version in this PR: `dict_deep_update(dict(), sub_dict_with_update_values)`. -> in this case, as we are now dealing with an empty dict, the same logic keeps applying and the algorithm just keeps adding the values of `sub_dict_with_updated_values` verbatim all the way to the bottom of the structure (either through copying sub-dicts, first if, list appending, second if, or just assignation of root elements, third if, when the stack hits the bottom).  
This means that the result is:
```python
dict_to_update[key_existing_only_in_dict_with_values_to_update] = dict_with_values_to_update[key_existing_only_in_dict_with_values_to_update]` 
```

The results should be the same.

CodyCBakerPhD (approved):

PR #483, opened 2022-04-26
add option to skip electrodes

bendichter:
## Motivation

For the Froemke Lab, they use Electrode 1 to record the waveform of the auditory stimulus, so we don't want to write this data as a neural response. @luiztauffer, I added an optional argument, `skip_electrodes`, to handle this, which is working. Do you think this approach is good or is there a better way?

### DIFF
--- a/nwb_conversion_tools/tools/neo/neo.py
+++ b/nwb_conversion_tools/tools/neo/neo.py
@@ -327,7 +327,7 @@ def add_icephys_recordings(
                 stim_unit = protocol[2][ei]
                 stim_conversion = get_conversion_from_unit(unit=stim_unit)
                 stimulus = stim_classes[icephys_experiment_type](
-                    name=f"stimulus-{si + 1 + simultaneous_recordings_offset}-ch-{ei}",
+                    name=f"stimulus-{si + 1 + simultaneous_recordings_offset:02}-ch-{ei}",
                     description=f"Stim type: {session_stimulus_type}",
                     electrode=electrode,
                     data=protocol[0][si][ei],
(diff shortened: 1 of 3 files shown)


nwb_conversion_tools/tools/neo/neo.py
```diff
@@ -308,7 +311,7 @@ def add_icephys_recordings(
             response_unit = neo_reader.header["signal_channels"]["units"][ei]
             response_conversion = get_conversion_from_unit(unit=response_unit)
             response_gain = neo_reader.header["signal_channels"]["gain"][ei]
-            response_name = f"{icephys_experiment_type}-response-{si + 1 + simultaneous_recordings_offset}-ch-{ei}"
+            response_name = f"{icephys_experiment_type}-response-{si + 1 + simultaneous_recordings_offset:02}-ch-{ei}"
```
luiztauffer, line 314: whats is `:02` doing in here?

luiztauffer: ah, alright, it's an f-string thing

luiztauffer:
@bendichter 
looks good to me!
there's a test error happening which I don't recall happening before:
```
ValueError: No valid arguments were found in the json type mapping!
```

this is happening at 
```
@pytest.mark.parametrize("data_interface", interface_list)
def test_interface_conversion_options_schema(data_interface):
    schema = data_interface.get_conversion_options_schema()
```

so it is probably due to the new argument `skip_electrodes: Optional[tuple] = ()`, which is not contemplated in `annotation_json_type_map`. We could substitute its type for a list, or is there any reason for defining the type as a tuple?

PR #484, opened 2022-04-26
Add_units_table also adds unit_name

h-mayorquin:
## Motivation

This is the second PR to break down all the code changes in https://github.com/catalystneuro/nwb-conversion-tools/pull/450 as requested by @CodyCBakerPhD .

Here functionality for dealing with non int units_ids in sorting extractor objects from `spikeinterface` is added. Plus, as in the case for the electrode table a new column called "units_name" is added with the goal of serving as a cross-writing identifier for row matching in the units table. 

Minor modifications of naming and style preparing for the last and larger refactor are also push forward in this PR. 

## How to test the behavior?
Simple tests were added to `test_si` for three cases:
1) Writing int units_ids
2) Writing string units_ids
3) Not overwriting the property of units_name if this is already in the sorting object.


CodyCBakerPhD:
@h-mayorquin Thanks again for taking the time to break this up. I'll take a look at this in detail tomorrow!

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1028,35 +1028,24 @@ def get_nspikes(units_table: pynwb.misc.Units, unit_id: int):
...
 
 
-def add_units(
+def add_units_table(
     sorting: SpikeInterfaceSorting,
     nwbfile: pynwb.NWBFile,
     property_descriptions: Optional[dict] = None,
     skip_properties: Optional[List[str]] = None,
     skip_features: Optional[List[str]] = None,
-    use_times: bool = True,
```
CodyCBakerPhD, line 1037: Why is `use_times` being removed here?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1065,38 +1054,50 @@ def add_units(
...
         if True and sorting is a spikeextractors SortingExtractor object then waveforms are added to the units table
         after writing.
     """
+    if not isinstance(nwbfile, pynwb.NWBFile):
+        raise TypeError(f"nwbfile type should be an instance of pynwb.NWBFile but got {type(nwbfile)}")
+
     if isinstance(sorting, SortingExtractor):
-        sampling_frequency = sorting.get_sampling_frequency()
-        if sampling_frequency is None:
-            raise ValueError("Writing a SortingExtractor to an NWBFile requires a known sampling frequency!")
```
CodyCBakerPhD, line 1086: What is the behavior when you convert a `SortingExtractor` that doesn't have a sampling frequency using `OldToNewSorting`? Does the resultant object have a default one or something?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1065,38 +1054,50 @@ def add_units(
...
+        )
+        if units_table_name not in nwbfile.processing:
+            units_table = pynwb.misc.Units(name=units_table_name, description=unit_table_description)
+            ecephys_mod.add(units_table)
+
+        units_table = ecephys_mod[units_table_name]
+    else:
+        if nwbfile.units is None:
+            nwbfile.units = pynwb.misc.Units(name=units_table_name, description=unit_table_description)
+        units_table = nwbfile.units
```
CodyCBakerPhD, line 1099: why not combine these two if blocks?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1132,35 +1133,40 @@ def add_units(
...
         if property_name in ["max_channel", "max_electrode"] and nwbfile.electrodes is not None:
             unit_col_args.update(table=nwbfile.electrodes)
+
         units_table.add_column(**unit_col_args)
         aggregated_unit_properties[property_name] = checked_sorting.get_property(key=property_name)
-    for i, unit_id in enumerate(unit_ids):
-        if use_times:
-            spkt = checked_sorting.get_unit_spike_train(unit_id=unit_id, return_times=True)
-        else:
-            spkt = checked_sorting.get_unit_spike_train(unit_id=unit_id) / checked_sorting.get_sampling_frequency()
```
CodyCBakerPhD, line 1141: Ah, so now that these objects are `BaseSorting`, they will always return in units of time (seconds)? And they have no notion of frame to timestamps mapping (the purpose of the `use_times` argument)?

CodyCBakerPhD: Or if they do have timestamps, they use those instead of scaling frames by frequency?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1132,35 +1133,40 @@ def add_units(
...
+    for i, unit_id in enumerate(units_ids):
+        spkt = checked_sorting.get_unit_spike_train(unit_id=unit_id, return_times=True)
         kwargs = {key: val[i] for key, val in aggregated_unit_properties.items()}
 
-        units_table.add_unit(id=int(unit_id), spike_times=spkt, **kwargs)
+        # If the units ids are integer keep the old behavior of asigning table's id equal to unit_ids
+        if str(unit_id).isdigit():
+            id = int(unit_id)
+        else:
+            id = i
```
CodyCBakerPhD, line 1161: Would it pose problems to let `add_unit` set the `id` by default?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1297,17 +1300,21 @@ def write_sorting(
...
+
     assert write_as in [
         "units",
         "processing",
     ], f"Argument write_as ({write_as}) should be one of 'units' or 'processing'!"
+
     if write_as == "units":
         assert units_name == "units", "When writing to the nwbfile.units table, the name of the table must be 'units'!"
+
+    write_in_processing_module = False if write_as == "units" else True
```
CodyCBakerPhD, line 1316: This would be fine to synchronize with above in this PR

CodyCBakerPhD:
@h-mayorquin Overall looks very good, I'll give it another pass later tonight to be sure but will probably approve/merge prior to our meeting tomorrow.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Right, they have that implemented in house in `spikeinterface`:
https://github.com/SpikeInterface/spikeinterface/blob/12b7863684b705e3d0ae8165bb3009143c70530c/spikeinterface/core/basesorting.py#L62-L69

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: I actually don't know the answer to that. I am actually not sure about this piece of code that was like this in the previous implementation. My hunch is that this was required because the sampling frequency was required to scale the spike times to the appropriate units.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: How? The logic of table creation or the one of assignation?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Compatibility with old behavior when the `unit_ids` are actually integers.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: See comments below.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Well, *why* sampling frequency is required for NWB is because it is the original `NwbSortingExtractor` required it (https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/nwbextractors/nwbextractors.py#L1134, or technically an associated electrical series that it would then infer it from).

So that still hasn't technically changed for a roundtrip, unless the new `BaseSorting` has a case to handle it.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD:
```python
if not x:
    do something


if x:
    do something else
else:
    # why not 'do something' here?
    do even more
```

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: So you're saying they wouldn't match the `unit_name` in that case?

h-mayorquin: 1) Let's suppose the unit_ids are 1, 3, 5, 8 (integers) 
2) Automatic assignation of id would created a unit table with ids 1, 2, 3, 4
3) The `NwbSortingExtractor` in `spikeextractors` gets its `units_ids` as 1, 2, 3, 4 and the comparison between the two of them will fail. 

This was happening with the Neuroscope recoding extractor if my memory does not fail me. I don't know if we have (or will have) a gin test with data like this.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Similar logic:
https://github.com/SpikeInterface/spikeinterface/blob/12b7863684b705e3d0ae8165bb3009143c70530c/spikeinterface/extractors/nwbextractors.py#L279-L294

h-mayorquin: But if that's the case, the assertion message seems misleading. We needed it for the round-trip test not because it was necessary to write the table.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Ah, right

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Hmm, it must have been for the `use_times`, then.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: I.e., why not just `nwbfile.units = pynwb.misc.Units(name="units", description=unit_table_description)` at 1098? Saves you three lines, just as readable

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1065,38 +1054,50 @@ def add_units(
...
+    if not write_in_processing_module:
+        units_table_name == "units"
+
+    if write_in_processing_module:
+        ecephys_mod = get_module(
+            nwbfile=nwbfile,
+            name="ecephys",
+            description="Intermediate data from extracellular electrophysiology recordings, e.g., LFP.",
+        )
+        if units_table_name not in nwbfile.processing:
```
CodyCBakerPhD, line 1091: Does this recurse across all the modules of `nwbfile.processing`? Shouldn't it check if it's in the `ecephys_mod`?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Should we remove it then? `spikeinterface` will throw an assertion if they don't have the times and our assertion is misleading.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: Right. I get you.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: You are correct. It should test whether it is in ecephys_mod.data_interfaces

I changed it. I added a test for writing in processing as well.

CodyCBakerPhD (approved):

PR #485, opened 2022-04-27
Change default behavior of setting property name as description in electrode table

h-mayorquin:
## Motivation
As discussed in the meeting yesterday, right now the electrode table is adding the property name as description by default (see lines 305): 

https://github.com/catalystneuro/nwb-conversion-tools/blob/604933380111397ff20b9fb0c847ecf4c1f59ae6/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L302-L312

Some of them then are overridden by the metadata (see lines 311). However, the ones not available in the meatadata will keep the description as property name which means that we are passing some non-meaningful descriptions automatically (things that would raise a warning with the `nwbinspector`). The goal of this PR is to change that behavior.

I also refactored a little bit the logic of the metadata and some naming.

Two side comments that might be addressed in some other PRs:

1) Maybe we want to change the logic of the metadata to only pass the property_descriptions (as we do in `add_units`) instead of requiring to pass the structured metadata dictionary the way that we are doing it right now. The idea, is that if we would like this function to serve as a stand alone functionality it would be too much to ask the users to format their descriptions in the highly structured way that our general metadata dict requires. 

2) Maybe we can also change the name of this function to `add_electrode_table` instead of `add_electrodes` which is very similar to the `add_electrode` function in `pynwb`? I followed the same logic for changing `add_units` to `add_units_table`. My feeling is that plurals are not very good for disambiguation. The situation is not as bad because those functions are not in the same library but nevertheless they are in the same ecosystem. It would be good to hear your takes.


CodyCBakerPhD:
> Maybe we want to change the logic of the metadata to only pass the property_descriptions (as we do in add_units) instead of requiring to pass the structured metadata dictionary the way that we are doing it right now.

No, let's stick to utilizing the common `metadata` structure everywhere this kind of things pops up.

> The idea, is that if we would like this function to serve as a stand alone functionality it would be too much to ask the users to format their descriptions in the highly structured way that our general metadata dict requires.

Which is why it's optional. TBH if all they want to do is specify descriptions, the sub-level of 'Ecephys' that would deal with that is only one extra key from what that other type of argument would require, so as long as it's well documented it should be easy to understand and use.

> Maybe we can also change the name of this function to add_electrode_table instead of add_electrodes which is very similar to the add_electrode function in pynwb?

Similar to your other PR, this kind of thing would require a deprecation cycle. IDK if it's really that important to distinguish.

> I followed the same logic for changing add_units to add_units_table. 

Which makes slightly more sense there because a `Units` table is an actual PyNWB data type, whereas there is only **one** Electrode table that has a defined place in the NWBFile, so any reference to *an* Electrode table is a reference to *that* Electrode table.

h-mayorquin:
> Which is why it's optional. TBH if all they want to do is specify descriptions, the sub-level of 'Ecephys' that would deal with that is only one extra key from what that other type of argument would require, so as long as it's well documented it should be easy to understand and use.

I disagree with your way of thinking about shoving costs under the rug. Complexity of input matters. Any structure that you impose in the final user is a hurdle that they need to overcome. What do we gain from making it more complicated? What do we gain from making them understand a hierarchical structure if they only want to specify the description of their tables?

Your argument is that "it is not that hard" which maybe is true but I feel that the world of computing is full of assumptions about usability by power users that end up being false for most people. Mainly, I think your argument is lacking a reason other than "it should be easy"  to make it more compelling, something like, what do we lose by making it easier?

That said, I think that the most important thing is to keep a consistent style with the other table. Maybe we should change the property descriptions to metadata in the units table (that is, pass the metadata instead of the property_descriptions map)?  What do you think?

CodyCBakerPhD:
> I disagree with your way of thinking about shoving costs under the rug. Complexity of input matters. Any structure that you impose in the final user is a hurdle that they need to overcome. What do we gain from making it more complicated? What do we gain from making them understand a hierarchical structure if they only want to specify the description of their tables?

It's not that much more complicated, and at a minimum it educates them in the structure the greater project uses so they can harness it for their future projects. Why add a special argument for one function out of the dozens that we support just for a functionality that is already otherwise covered? Unless we add it for *all* of our functions, but again why do that when (a) descriptions are just one type of metadata and (b) all metadata is contained in the `metadata` dictionary?



> Maybe we should change the property descriptions to metadata in the units table

Yes, that was mentioned as a part of the refactor for that part, it's something that should have been done a while ago.

h-mayorquin:
> > I disagree with your way of thinking about shoving costs under the rug. Complexity of input matters. Any structure that you impose in the final user is a hurdle that they need to overcome. What do we gain from making it more complicated? What do we gain from making them understand a hierarchical structure if they only want to specify the description of their tables?
> 
> It's not that much more complicated, and at a minimum it educates them in the structure the greater project uses so they can harness it for their future projects. Why add a special argument for one function out of the dozens that we support just for a functionality that is already otherwise covered? Unless we add it for _all_ of our functions, but again why do that when (a) descriptions are just one type of metadata and (b) all metadata is contained in the `metadata` dictionary?

Yes, that seems to be the crux of the issue. 
My answer to your questions at the end is:
Because it is simpler. We should also make all the other functions simpler. It reduces the number of things that the function does. It increases the cohesion of the function and reduces coupling which are good guidelines of good software design. The first by reducing the number of things that the function does: you don't have to parse the metadata anymore. For the second point I would say that having a function that requires access to the deep structure of an argument is a code smell for high coupling. Too much knowledge required.

Parsing the metadata can be done in a specific function like `add_electrodes_info` which is doing [nothing ](https://github.com/catalystneuro/nwb-conversion-tools/blob/5655bdc998abe49cf231fd6e0ade2835e9f82cbf/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L731-L764)now). The argument becomes event stronger if you look at the other functions: add devices spends 80 % of its code just parsing the metadata:

https://github.com/catalystneuro/nwb-conversion-tools/blob/5655bdc998abe49cf231fd6e0ade2835e9f82cbf/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py#L119-L133

Your argument if I can represent it well is that we should educated the users by forcing them to think more about the arguments so they can use that knowledge when they use our other functions. You also believe that it is not so complicated. 

I think that we hit the crux of the issue. Let's agree to disagree. This was my mistake for opening this as a tangential point in a PR. It should have been an issue of its own. Let's focus on the code here which is not related to this debate that took all the space.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -273,25 +273,28 @@ def add_electrodes(
...
         "containing the keys 'name' and 'description'"
     )
+
     assert all(
-        [x["name"] != "group" for x in metadata["Ecephys"]["Electrodes"]]
+        [property["name"] != "group" for property in electrodes_metadata]
     ), "Passing metadata field 'group' is deprecated; pass group_name instead!"
 
+    # Transform to a dict that maps property name to its description
+    property_descriptions = dict()
```
CodyCBakerPhD, line 294: Are there any default descriptions we might be able to provide? Or does PyNWB auto-fill useful descriptions for things like `rel_x`, `impedence`, etc.

h-mayorquin: `Pynwb` does fill descriptions for those values.

CodyCBakerPhD (approved):

PR #488, opened 2022-04-27
[Feature] globus-specific size retrieval helper

CodyCBakerPhD:
Here is a working minimal helper function for evaluating the size of a directory on a globus endpoint.

Our CI may take some configuring to get the test to work, but the test passes on a system with an authenticated local endpoint.

(i) Other approaches:

(a) datalad - there has been a small amount of work in the past trying to get this working (https://github.com/datalad/datalad/issues/3758) which requires git-annex compatibility with globus, leading the development of  https://github.com/CONP-PCNO/git-annex-remote-globus which looked promising, but did not work out of the box on the DANDIHub due to the `no_browser` option not propagating up to it's higher dependencies, so while it may be possible to fix this eventually, I have no clue how long or how much work it would take. Suffice to say it did not work out of the box. 

(b) Globus SDK instead of CLI: While this code pipes directly to the globus CLI for simplicity, their CLI code is all done in Python so it is possible 'in theory' to replicate their code for handling the request through the transfer client. Unfortunately for us, their approach to CLI usage is not to be a simple wrapper around a library function but rather the code within the CLI function does quite a lot to handle the API and parse output: https://github.com/globus/globus-cli/blob/f6d5e96018f3057b7e90810f438967baf8ad5a1e/src/globus_cli/commands/ls.py#L161-L218

So if we wanted to opt for a direct library usage, we would essentially have to do all that plus authentication in https://github.com/globus/globus-cli/blob/f6d5e96018f3057b7e90810f438967baf8ad5a1e/src/globus_cli/commands/ls.py#L122

(ii) Future direction:

(a) This will of course work for datasets using globus specifically, but it's my opinion that any other types of data hosting (google drive, box, and many many others) should be handled with `rclone`.

See https://rclone.org/commands/rclone_size/ for a nice, basic, machine readable size function readily offered on any rclone remote.

See https://rclone.org/ for the list of supported data hosting, which is quite a lot! Basically most sources except globus and git-annex (which datalad would then be the best choice  for specifically).

(b) Technically, datalad also says it supports rclone intermediately: https://github.com/datalad/git-remote-rclone, but have not tested this out yet. So if we wanted to use datalad as the unifying API of both globus and rclone, just be aware it may take some participation on our end to debug things, just like we often have to do with SpikeInterface.

Bonus: If curious why rclone does not support globus, see https://github.com/rclone/rclone/issues/2313.

### DIFF
--- a/tests/test_internals/test_tools.py
+++ b/tests/test_internals/test_tools.py
@@ -1,10 +1,21 @@
+import os
 from datetime import datetime
 
+import pytest
 from pynwb.base import ProcessingModule
 from hdmf.testing import TestCase
 
 from nwb_conversion_tools.tools.nwb_helpers import get_module, make_nwbfile_from_metadata
 from nwb_conversion_tools.tools.data_transfers import get_globus_dataset_content_sizes
+
+try:
+    import globus_cli
+
+    HAVE_GLOBUS, LOGGED_INTO_GLOBUS = True, True
+    if not os.popen("globus ls 188a6110-96db-11eb-b7a9-f57b2d55370d").read():
+        LOGGED_INTO_GLOBUS = False
+except ModuleNotFoundError:
+    HAVE_GLOBUS, LOGGED_INTO_GLOBUS = False, False
 
 
 class TestConversionTools(TestCase):
@@ -37,8 +48,13 @@ class TestConversionTools(TestCase):
             make_nwbfile_from_metadata(metadata=dict())
 
 
+@pytest.mark.skipif(
+    not HAVE_GLOBUS or not LOGGED_INTO_GLOBUS,
+    reason="You must have globus installed and be logged in to run this test!",
+)
 def test_get_globus_dataset_content_sizes():
     """Test is fixed to a subpath that is somewhat unlikely to change in the future."""
+    globus_cli.login_manager.LoginManager._TEST_MODE = True
     assert get_globus_dataset_content_sizes(
         globus_endpoint_id="188a6110-96db-11eb-b7a9-f57b2d55370d",
         path="/SenzaiY/YutaMouse41/YutaMouse41-150821/originalClu/",
--- a/tests/test_internals/test_tools.py
+++ b/tests/test_internals/test_tools.py
@@ -49,7 +49,7 @@ class TestConversionTools(TestCase):
 
 
 @pytest.mark.skipif(
-    not (HAVE_GLOBUS and LOGGED_INTO_GLOBUS)
+    not (HAVE_GLOBUS and LOGGED_INTO_GLOBUS),
     reason="You must have globus installed and be logged in to run this test!",
 )
 def test_get_globus_dataset_content_sizes():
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Putting into draft until I can figure out how to configure the tests to run

CodyCBakerPhD:
Yeah... short of attempting to copy their entire approach to 'faking' authentication they use in their tests (https://github.com/globus/globus-cli/blob/3c3c2bf2e96112ce68c71ca23de6100390201a96/tests/functional/test_login_command.py) it seems like a real pain to get this thing setup on the CI.

The way globus does login/authentication, even through their Python SDK library, requires interactivity in the way of navigating to a generated URL, copy the generated token at the link, and pasting it back into the one-time login instance. No clue how to do something like that in a CI, so just making the tests lazy for now. If I discover a way to enable it in the future as I learn more about the package, I can come back to this.

tests/test_internals/test_tools.py
```diff
@@ -34,3 +46,33 @@ def test_make_nwbfile_from_metadata(self):
             ),
         ):
             make_nwbfile_from_metadata(metadata=dict())
+
+
+@pytest.mark.skipif(
+    not HAVE_GLOBUS or not LOGGED_INTO_GLOBUS,
```
h-mayorquin, line 52: Maybe the equivalent:
```suggestion
    not (HAVE_GLOBUS and LOGGED_INTO_GLOBUS)
```
Is more clear?

h-mayorquin (approved):
Looks nice! 
For people who have access to data we can do the direct conversion with this, right? or what use case do you have in mind.

I have this vague memory that `subprocess` was endorsed over `os.popen`  a couple of years ago. Maybe things have changed.

h-mayorquin:
> Bonus: If curious why rclone does not support globus, see [rclone/rclone#2313](https://github.com/rclone/rclone/issues/2313).

So, lack of interest? : )

CodyCBakerPhD:
>> Bonus: If curious why rclone does not support globus, see https://github.com/rclone/rclone/issues/2313.

> So, lack of interest? : )

Apparently, lol.


> I have this vague memory that subprocess was endorsed over os.popen a couple of years ago. Maybe things have changed.

I can look into that, then.


> For people who have access to data we can do the direct conversion with this, right? or what use case do you have in mind.

For this very early use case, this is for assessing the initial cost of performing a transfer/conversion/upload process. Eventually this kind of thing will lead to the automated transfer and upload, but taking it slow piece by piece.

CodyCBakerPhD:
> I can look into that, then.

OK after playing around with it and reading, it seems `Popen` is better for customized control over execution of command line inputs to the system, but it's quite a bit messier to extract and parse the text output from that process, whereas `os` (granted is simply a wrapper around it plus some other stuff) handles all that for you in a single command.

bendichter (approved):
This looks great! It's a shame the rclone does not support Globus.

PR #489, opened 2022-04-27
[Feature] Add s3 cost equation

CodyCBakerPhD:
Next steps towards the complete SDK for assessing data size, cost, performing transfer, conversion, and upload.

Uses our derived equation for cost of hosting data on S3 throughout the process, assuming no gaps between segments.


bendichter:
This is great! Could you also add a function for estimating the conversion time?

tests/test_internals/test_tools.py
```diff
@@ -76,3 +76,37 @@ def test_get_globus_dataset_content_sizes():
...
+@pytest.mark.skipif(
+    not HAVE_GLOBUS or not LOGGED_INTO_GLOBUS,
+    reason="You must have globus installed and be logged in to run this test!",
+)
+def test_get_s3_conversion_cost_from_globus():
+    content_sizes = get_globus_dataset_content_sizes(
+        globus_endpoint_id="188a6110-96db-11eb-b7a9-f57b2d55370d",
+        path="/SenzaiY/YutaMouse41/YutaMouse41-150821/originalClu/",
+    )
+    assert get_s3_conversion_cost(total_mb=sum(content_sizes.values()) / 1e6) == 1.756555806400279e-13
```
bendichter, line 112: This is not a linear equation, so sum(get_s3_conversion_cost(x) for x in sizes) != get_s3_conversion_cost(sum(sizes)). This works here because it's one session, but for multiple sessions you would need to do sum(get_s3_conversion_cost(x) for x in sizes)

CodyCBakerPhD: Ah, yeah my bad

CodyCBakerPhD: Right, the contents above show it's a collection of Neuroscope spike sorting files (one per shank) so I'm estimate the cost of a single NeuroscopeSorting conversion there.

But yes, it is important to remember that when calculating total cost of a dataset over multiple sessions, you'd have to get the cost for each session and then add all the costs.

bendichter:
looks great!

PR #492, opened 2022-04-28
[Feature] Add automatic dandi upload

CodyCBakerPhD:
WIP, needs a bit more testing and may require further CI configuration.

This would be the finally generic stage of the automated data transfer workflow.

Also, for explanation on why all of this is done via CLI processes - it seems like the `dandi-cli` is similar to the `globus-cli` in that each and every command line function we normally use for the dandi upload process does some pretty substantial processing inside each function - they are not simply `click` wrappers around a single library function. So unless we help them decouple that, or decouple it ourselves, the easiest way to get this kind of thing automated quickly on our end is to essentially run the same set of commands we usually would ourselves, but with automatic shell processes.


src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -84,3 +89,62 @@ def estimate_s3_conversion_cost(
...
+    url_base = "https://gui-staging.dandiarchive.org" if staging else "https://dandiarchive.org"
+    dandiset_url = f"{url_base}/dandiset/{dandiset_id}/{version}"
+
+    os.chdir(nwb_folder_path.parent)
+    validate_return = os.popen(f"dandi validate {nwb_folder_path.name}").read().strip()
+    assert re.fullmatch(
+        pattern=r"^Summary: No validation errors among \d+ file\(s\)$", string=validate_return
+    ), "DANDI validation failed!"
+
+    download_return = os.popen(f"dandi download {dandiset_url}").read()
```
bendichter, line 133: are you downloading data within the NWB folder? I don't know if that's the best idea. It requires write access to the NWB folder, which you may not have, and in general I'd prefer to keep the staged dandiset in a different spot from the original NWB files.

src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -84,3 +89,62 @@ def estimate_s3_conversion_cost(
...
+    assert download_return, "DANDI download failed!"  # output is a bit too dynamic to regex; if it fails it is empty
+
+    os.chdir(nwb_folder_path.parent / dandiset_id)
+    os.system(f"dandi organize ../{nwb_folder_path.name}")
+    assert len(list(Path.cwd().iterdir())) > 1, "DANDI organize failed!"
+
+    dandi_upload_command = "dandi upload -i dandi-staging" if staging else "dandi upload"
+    proc = Popen(dandi_upload_command, shell=True, stdin=PIPE)
+    proc.stdin.write(bytes(f"{api_token}", "utf-8"))
```
bendichter, line 142: It might be easier to set DANDI_API_KEY as an environment variable

src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -84,3 +89,62 @@ def estimate_s3_conversion_cost(
     cost_gb_m = 0.08 / 1e3  # $0.08 / GB Month
     cost_mb_s = cost_gb_m / (1e3 * 2.628e6)  # assuming 30 day month; unsure how amazon weights shorter months?
     return cost_mb_s * total_mb_s
+
+
+def automatic_dandi_upload(
```
bendichter, line 94:
```suggestion
def dandi_upload(
```

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: > are you downloading data within the NWB folder?

Ah, yes technically if the dandiset has previous uploads this will grab those unintentially. I'll add `--download dandiset.yaml`, we just need the shell folder for upload setup.

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: Agreed, it's really the safest way. We still have to pipe it to the dandi upload prompt however, as simply setting it does not prevent the prompt from happening. But definitely preferable to the `input` above.

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: Also added dandiset folder specification so user can control where it gets downloaded in the event the nwb folder source doesn't have write access

src/nwb_conversion_tools/tools/data_transfers.py
bendichter: set, not get. I don't really understand what's happening here.

CodyCBakerPhD: When you run `dandi upload`, it prompts interactively for an API key. We can pass it automatically in the program via the lines above, if we have read it from the system in the first few lines of the function. I'm assuming now that the user has set it as an environment variable prior to running the program.

bendichter: If you set the enviroment variable you won't be prompted for the key

bendichter: See https://www.dandiarchive.org/handbook/10_using_dandi/#storing-access-credentials

CodyCBakerPhD: Ah, I had tried this the python way `import os; os.environ["DANDI_API_KEY"] = "...."` but that did work for me, but setting it in the bash terminal does.

CodyCBakerPhD: Yeah that's apparently a problem of scope.

Within Python instance run inside a bash instance, doing the `os` method only adds it to environment variables inside the Python instance, not inside the outer bash instance.

But setting it at the bash level then using it in Python code called from that same bash instance works fine.

Now instructing users to set it at the bash level before running any code involving this helper function.

CodyCBakerPhD:
OK tests are passing now.

Was having a lot of trouble with getting `os.popen` processes to relinquish file permissions. Solution to this is (a) use the suggested `subprocess` directly (which `os` is essentially a wrapper of) with a `pustil`-based helper that ensures the processes are properly ended after each command.

Otherwise, any attempt to clean up the files or folders fails due to `PermissionErrors` since it thinks the processes are still running.

bendichter (approved):

PR #493, opened 2022-05-02
Refactor units table II

h-mayorquin:
## Motivation
This should close #450 
The refactor of `add_units_table` here add capabilities for re-writing the units_table, uses the new `spikeinterface` methods for extracting properties and aligns the code with that of `add_electrodes`. 
 
## How to test the behavior?
I added tests for the repeated writing behavior.

### DIFF
--- a/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
+++ b/src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
@@ -1075,10 +1075,7 @@ def add_units_table(
     else:
         checked_sorting = sorting
 
-    sampling_frequency = checked_sorting.get_sampling_frequency()
-    if sampling_frequency is None:
-        raise ValueError("Writing a SortingExtractor to an NWBFile requires a known sampling frequency!")
-
+    write_table_first_time = False
     if write_in_processing_module:
         ecephys_mod = get_module(
             nwbfile=nwbfile,
@@ -1088,11 +1085,13 @@ def add_units_table(
         if units_table_name not in ecephys_mod.data_interfaces:
             units_table = pynwb.misc.Units(name=units_table_name, description=unit_table_description)
             ecephys_mod.add(units_table)
+            write_table_first_time = True
 
         units_table = ecephys_mod[units_table_name]
     else:
         if nwbfile.units is None:
             nwbfile.units = pynwb.misc.Units(name="units", description=unit_table_description)
+            write_table_first_time = True
         units_table = nwbfile.units
 
     default_descriptions = dict(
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
Thanks again for splitting this out - I'll take a look over the next couple days!

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1078,10 +1078,7 @@ def add_units_table(
     else:
         checked_sorting = sorting
 
-    sampling_frequency = checked_sorting.get_sampling_frequency()
-    if sampling_frequency is None:
-        raise ValueError("Writing a SortingExtractor to an NWBFile requires a known sampling frequency!")
-
+    write_table_first_time = False
```
CodyCBakerPhD, line 1081: What about

```
write_table_first_time = units_table_name not in ecephys_mod.data_interfaces if write_in_processing_module else nwbfile.units is None
```

Saves you two lines below that are in the double-nested `if` conditions

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1118,51 +1117,108 @@ def add_units_table(
...
+        unit_name_array = data_to_add["unit_name"]["data"]
     else:
         unit_name_array = units_ids.astype("str", copy=False)
-        unit_col_args = dict(
-            name="unit_name",
-            description="unique reference for each unit",
-        )
-        units_table.add_column(**unit_col_args)
-        aggregated_unit_properties["unit_name"] = unit_name_array
+        data_to_add["unit_name"].update(description="unique reference for each unit", data=unit_name_array, index=False)
```
CodyCBakerPhD, line 1140: Let's try our best to keep `descriptions` to be proper sentences (start with capital, end with punctuation). It's too much to require it or even have it be a best practice, but there's no excuse not to do it for anything we set automatically.

CodyCBakerPhD: Also, is the `index=False` explicitly necessary? Shouldn't it be `False` by default? Ditto for below with the `id` key.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1118,51 +1117,108 @@ def add_units_table(
...
+    # If the channel ids are integer keep the old behavior of asigning table's id equal to unit_ids
+    if np.issubdtype(units_ids.dtype, np.integer):
+        data_to_add["id"].update(data=units_ids.astype("int"), index=False)
 
-        # If the units ids are integer keep the old behavior of asigning table's id equal to unit_ids
-        if str(unit_id).isdigit():
-            id = int(unit_id)
-        else:
-            id = i
+    units_table_previous_properties = set(units_table.colnames) - {"spike_times"}
```
CodyCBakerPhD, line 1146: Is "spike_times" all you need or to you also need "spike_times_index"?

CodyCBakerPhD: Also, using literals wherever possible helps readability. *E.g.*, `{"spike_times"}` -> `set(["spike_times"])`. The extra computational time from casting the list ought to be negligible. Otherwise it can be easy to confuse on quick glance with a `dict`.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
```diff
@@ -1118,51 +1117,108 @@ def add_units_table(
...
+    extracted_properties = set(data_to_add)
+    properties_to_add_by_rows = units_table_previous_properties | {"id"}
```
CodyCBakerPhD, line 1148: Personally I'm not sure when it comes to readability. What's better, 

```python
set_c = set_a | set_b
```

or

```python
set_c = set_a.union(set_b)
```

?

CodyCBakerPhD:
Looks good overall! A couple things to think about for follow-ups

1) Would it be possible to generalize this lovely append logic into a common tool helper function or even a class that can be harnessed not only by the `add_electrodes` and `add_units` but also by any future table-related interfaces we might want to add?

2) SpikeInterface objects (new ones) support 'annotations' or something like that, right? Would it be possible to extract metadata descriptions automatically through those?

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: It annoys me that it is a little bit too long and black decomposes it. On the other hand, besides the line saving, we get to handle the whole logic of the assignation in one line instead of having it distributed over the control structures below (hence better for short-memory). I will change it.

I wish we could have the definition closer to where it is in fact used but this version needs to be almost at the beginning.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: I personally find the notation of the left more readable but I have been using sets for two long. What you say sounds true: someone seeing the word set as in the literal is better for avoiding the confusion. I am curios on why python would not allow a directly set construction like `set(element_1, element_2, element_3)` probably hard to parse compared to the normal construction `set(list)`.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
h-mayorquin: I never thought about this one. I am on the fence.
Does comparing a case with more elements case makes you more incline for one or the other?


`total_union = A | B | C | D`
vs
`total_union = A.union(B).union(C).union(D)`

h-mayorquin:
> Looks good overall! A couple things to think about for follow-ups
> 
> 1. Would it be possible to generalize this lovely append logic into a common tool helper function or even a class that can be harnessed not only by the `add_electrodes` and `add_units` but also by any future table-related interfaces we might want to add?
> 2. SpikeInterface objects (new ones) support 'annotations' or something like that, right? Would it be possible to extract metadata descriptions automatically through those?

1) Yes. We might be able to do that. I have been thinking about this. Right now, we only have two cases. I think it would make sense to wait till a third case presents itself so we can generalize.  

2) I thought more about this and it feels less clear to me. So, right now we fill the descriptions to the metadata during the initialization of the data-interfaces ( see the `neuroscope` and `spikeglx` cases) and propagate this all the way to `write_recording` and `write_sorting`. This is our pipeline right now and I don't think that there is a lot to gain from diverting it to the `spikeinterface` object with an annotation for them to do the same propagation at least not evidently. 

Some possible points in this direction that i can think of:
* Maybe this simplifies our metadata? seems unlikely as we might still need to pass some descriptions.  
* More complicated, what If `spikeinterface` objects were to do the annotation themselves for some properties that are consistently found in their data extraction process. That way, we would be aligning and filling a common pipe-line instead of just diverting ours. But I think this falls into a bigger discussion: which responsibilities we want `spikeinterface` to handle vis a vis the ones that `nwb-conversion-tool` handles?. 

There are other important discussions over there like the one that you made of who should handle single interface conversions. I think this is a good issue to discuss in our group meeting now that Alesio and Samuel are back from vacations.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: > Oh, the problem is that this requires you to have previously defined ecephys_mod

Ah, right, dang.

But yeah I generally prefer to try to do these kinds of things in a single clear definition rather than nested logic which requires both mental tracking of nested logic and eye-tracking to navigate code.

src/nwb_conversion_tools/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: It's mostly that the pipe operator is so unique to sets and doesn't show up elsewhere, you have to mentally think about what it means if it's been a while since working with it.

I mean, but multiple unions you could also do it iteratively pairwise regardless (and probably should if it's dynamically sized or larger than 3 items)

CodyCBakerPhD (approved):

PR #494, opened 2022-05-02
[Feature] Automated globus transfer

CodyCBakerPhD:
Finally got this working - tests are passing on any local device logged into globus.

Depends on the process management helpers in #492, which I've now made non-private since it's proving quite useful generally. (@h-mayorquin should be happy to hear it uses `subprocess` instead of `os`)

This is the final item required to fully automated the Tingley conversion, which I'll start doing now.

### DIFF
--- a/src/nwb_conversion_tools/tools/data_transfers.py
+++ b/src/nwb_conversion_tools/tools/data_transfers.py
@@ -79,7 +79,7 @@ def transfer_globus_content(
     display_progress: bool = True,
     progress_update_rate: float = 60.0,
     progress_update_timeout: float = 600.0,
-) -> bool:
+) -> (bool, List[str]):
     """
     Track download progress for transferring content from source_id to destination_id:destination_folder.
 
@@ -119,6 +119,8 @@ def transfer_globus_content(
 
         If 'display_progress'=True (the default), then this function returns the total status of all transfers
         when they either finish or the progress tracking times out.
+    task_ids : list of strings
+        List of the task IDs submitted to globus, if further information is needed to reestablish tracking or terminate.
     """
     source_files = [[source_files]] if isinstance(source_files, str) else source_files
     # assertion check is ensure the logical iteration does not occur over the individual string values
@@ -182,7 +184,7 @@ def transfer_globus_content(
                 task_message = json.loads(_deploy_process(f"globus task show {task_id}", catch_output=True))
                 all_status[j] = task_message["status"] == "SUCCEEDED"
                 all_pbars[j].update(n=task_message["bytes_transferred"])
-    return success
+    return success, list(task_total_sizes)
 
 
 def estimate_total_conversion_runtime(
--- a/src/nwb_conversion_tools/tools/data_transfers.py
+++ b/src/nwb_conversion_tools/tools/data_transfers.py
@@ -119,6 +119,93 @@ def transfer_globus_content(
     task_ids : list of strings
         List of the task IDs submitted to globus, if further information is needed to reestablish tracking or terminate.
     """
+
+    def _submit_transfer_request(
+        source_endpoint_id: str,
+        source_files: Union[str, List[List[str]]],
+        destination_endpoint_id: str,
+        destination_folder: Path,
+    ) -> Dict[str, int]:
+        """Send transfer request to Globus."""
+        folder_content_sizes = dict()
+        task_total_sizes = dict()
+        for j, batched_source_files in enumerate(source_files):
+            paths_file = destination_folder_path / f"paths_{j}.txt"
+            # .as_posix() to ensure correct string form Globus expects
+            # ':' replacement for Windows drives
+            source_folder = Path(batched_source_files[0]).parent.as_posix().replace(":", "")
+            destination_folder_name = destination_folder_path.as_posix().replace(":", "")
+            with open(file=paths_file, mode="w") as f:
+                for source_file in batched_source_files:
+                    file_name = Path(source_file).name
+                    f.write(f"{file_name} {file_name}\n")
+
+            transfer_command = (
+                "globus transfer "
+                f"{source_endpoint_id}:{source_folder} {destination_endpoint_id}:/{destination_folder_name} "
+                f"--batch {paths_file}"
+            )
+            transfer_message = deploy_process(
+                command=transfer_command,
+                catch_output=True,
+            )
+            task_id = re.findall(
+                pattern=(
+                    "^Message: The transfer has been accepted and a task has been created and queued for "
+                    "execution\nTask ID: (.+)$"
+                ),
+                string=transfer_message,
+            )
+            paths_file.unlink()
+            assert task_id, f"Transfer submission failed! Globus output:\n{transfer_message}."
+
+            if source_folder not in folder_content_sizes:
+                contents = get_globus_dataset_content_sizes(globus_endpoint_id=source_endpoint_id, path=source_folder)
+                folder_content_sizes.update({source_folder: contents})
+            task_total_sizes.update(
+                {
+                    task_id[0]: sum(
+                        [
+                            folder_content_sizes[source_folder][Path(source_file).name]
+                            for source_file in batched_source_files
+                        ]
+                    )
+                }
+            )
+        return task_total_sizes
+
+    def _track_transfer(
+        task_ids: List[str],
+        display_progress: bool = True,
+        progress_update_rate: float = 60.0,
+        progress_update_timeout: float = 600.0,
+    ) -> bool:
+        """Track the progress of transfers."""
+        if display_progress:
+            all_pbars = [
+                tqdm(desc=f"Transferring batch #{j}...", total=total_size, position=j, leave=True)
+                for j, total_size in enumerate(task_total_sizes.values(), start=1)
+            ]
+        all_status = [False for _ in task_total_sizes]
+        success = all(all_status)
+        time_so_far = 0.0
+        start_time = time()
+        while not success and time_so_far <= progress_update_timeout:
+            time_so_far = time() - start_time
+            for j, (task_id, task_total_size) in enumerate(task_total_sizes.items()):
+                task_update = deploy_process(f"globus task show {task_id} -Fjson", catch_output=True)
+                task_message = json.loads(task_update)
+                all_status[j] = task_message["status"] == "SUCCEEDED"
+                assert (
+                    all_status[j] != "OK" or all_status[j] != "SUCCEEDED"
+                ), f"Something went wrong with the transfer! Please manually inspect the task with ID '{task_id}'."
+                if display_progress:
+                    all_pbars[j].update(n=task_message["bytes_transferred"] - all_pbars[j].n)
+            success = all(all_status)
+            if not success:
+                sleep(progress_update_rate)
+        return success
+
     source_files = [[source_files]] if isinstance(source_files, str) else source_files
     destination_folder_path = Path(destination_folder)
     # assertion check is ensure the logical iteration does not occur over the individual string values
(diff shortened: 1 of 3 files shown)


src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -61,14 +63,143 @@ def get_globus_dataset_content_sizes(
...
+        deploy_process(
             command=f"globus ls -Fjson {globus_endpoint_id}:{path}{recursive_flag}", catch_output=True, timeout=timeout
         )
     )
     files_and_sizes = {item["name"]: item["size"] for item in contents["DATA"] if item["type"] == "file"}
     return files_and_sizes
 
 
+def transfer_globus_content(
+    source_id: str,
```
bendichter, line 75: would you mind this change to make it a bit more explicit?

```suggestion
    source_globus_id: str,
```

bendichter, line 75: would you mind this change to make it a bit more explicit?

```suggestion
    source_globus_id: str,
```

bendichter: or source_endpoint_id

bendichter: or source_endpoint_id

src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -61,14 +63,143 @@ def get_globus_dataset_content_sizes(
...
+    for j, batched_source_files in enumerate(source_files):
+        paths_file = Path(destination_folder) / f"paths_{j}.txt"
+        # .as_posix() to ensure correct string form Globus expects
+        # ':' replacement for Windows drives
+        source_folder = Path(batched_source_files[0]).parent.as_posix().replace(":", "")
+        destination_folder_name = Path(destination_folder).as_posix().replace(":", "")
+        with open(file=paths_file, mode="w") as f:
+            for source_file in batched_source_files:
+                file_name = Path(source_file).name
+                f.write(f"{file_name} {file_name}\n")
```
bendichter, line 146: The easiest way to pass files in is to create an intermediate text file?

bendichter, line 146: The easiest way to pass files in is to create an intermediate text file?

src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -61,14 +63,143 @@ def get_globus_dataset_content_sizes(
...
+        Maximum amount of time to monitor the transfer progress.
+        You may wish to set this to be longer when transferring very large files.
+        Defaults to 10 minutes.
+
+    Returns
+    -------
+    success : bool
+        If 'display_progress'=False, and the transfer was succesfully initiated, then this function returns True.
+
+        If 'display_progress'=True (the default), then this function returns the total status of all transfers
```
h-mayorquin, line 120:
```suggestion
        If 'display_progress'=True (the default), then this function returns the total size of all transfers
```
Is this what you mean?

src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -61,14 +63,143 @@ def get_globus_dataset_content_sizes(
...
+            ),
+            string=transfer_message,
+        )
+        paths_file.unlink()
+        assert task_id, f"Transfer submission failed! Globus output:\n{transfer_message}."
+
+        if source_folder not in folder_content_sizes:
+            contents = get_globus_dataset_content_sizes(globus_endpoint_id=source_id, path=source_folder)
+            folder_content_sizes.update({source_folder: contents})
+        task_total_sizes.update(
```
h-mayorquin, line 169: It seems a little bit complicated to see what structure you get at the end. Flagging it so I can think a little bit better about it. Maybe there is a better way of writing this.

tests/test_internals/test_tools.py
```diff
@@ -181,3 +182,34 @@ def tearDown(self):
...
+        ]
+        success, task_ids = transfer_globus_content(
+            source_id=source_id,
+            source_files=test_source_files,
+            destination_id=destination_id,
+            destination_folder=self.tmpdir,
+            display_progress=False,
+        )
+        assert success
+        assert task_ids
```
h-mayorquin, line 215: Is this just testing that the function does not return empty arguments? Can we test size file in the temporary folder?

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: The *only* way to use the `--batch` option, regrettably. https://github.com/globus/globus-cli/blob/main/src/globus_cli/commands/transfer.py#L155-L165

Could technically go down to that lower level of the `globus_sdk` library and add the items programmatically instead, but that requires a heck of a lot of other steps to get unpacked and setup before doing that (auth/client/every other option). Would be much harder to maintain on our end.

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: Maybe just 'progress' if 'status' is too equivocal with being boolean.

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: double-nested dict tree - outer level is unique common parent folders, inner level is filename in that folder with value equal to size in bytes.

Not that the JSON content API request is that expensive, but still, you'd really only want it to have to grab the info once per unique folder.

tests/test_internals/test_tools.py
CodyCBakerPhD: `success` is `bool`, so asserts that it is True.

`task_ids` is a list, so asserts that it is not empty.

> Can we test size file in the temporary folder?

Yeah, sure.

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: No, 'status' is correct, everything here is boolean. `success` is the global status, `all(all_status)`, and `all_status` is the list of booleans.

src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -61,14 +63,143 @@ def get_globus_dataset_content_sizes(
...
 
+def transfer_globus_content(
+    source_id: str,
+    source_files: Union[str, List[List[str]]],
+    destination_id: str,
+    destination_folder: FolderPathType,
+    display_progress: bool = True,
+    progress_update_rate: float = 60.0,
+    progress_update_timeout: float = 600.0,
+) -> (bool, List[str]):
```
h-mayorquin: This is flagged in my editor, shouldn't it be `Tuple(bool, List[str])`?

src/nwb_conversion_tools/tools/data_transfers.py
h-mayorquin: I see, so maybe: then this functions returns success=True if all transfer have succeeded?
I think what confuses me is the phrase "returns total status" by which you mean returns `all(all_status)`

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: Done

tests/test_internals/test_tools.py
CodyCBakerPhD: Thanks for pointing this out, actually revealed an issue with the relative globus paths. Working on it now.

h-mayorquin (commented):
Shouldn't this be two functions? 
The first puts forward the globus process and returns one argument that is the structure of what is to be transfered: the dictionary "task_total_sizes" (it returns an empty dict if it fails).
(lines 125-179)
The second gets that dictionary and monitors the progress as it occurs. Returns True if everything went well, returns false if anything failed in the middle.
(lines 180-200)

Those functions would have a more clear responsibility.

CodyCBakerPhD:
> Shouldn't this be two functions?
The first puts forward the globus process and returns one argument that is the structure of what is to be transfered: the dictionary "task_total_sizes" (it returns an empty dict if it fails).
(lines 125-179)
The second gets that dictionary and monitors the progress as it occurs. Returns True if everything went well, returns false if anything failed in the middle.
(lines 180-200)
>
> Those functions would have a more clear responsibility.

I can make them private sub-functions if you think that would be better. The real trouble is that subfunction 1 is asynchronous but we want to force it to lock the running of a script that would then run the full data conversion the moment the transfer completes (assuming `success=True`, else need manual intervention). So they do need to go hand-in-hand for the primary function call.

src/nwb_conversion_tools/tools/data_transfers.py
CodyCBakerPhD: > I see, so maybe: then this functions returns success=True if all transfer have succeeded?

Correct.

> I think what confuses me is the phrase "returns total status" by which you mean returns all(all_status)

Yeah, 'total status' = combined success of all batches

h-mayorquin:
> > Shouldn't this be two functions?
> > The first puts forward the globus process and returns one argument that is the structure of what is to be transfered: the dictionary "task_total_sizes" (it returns an empty dict if it fails).
> > (lines 125-179)
> > The second gets that dictionary and monitors the progress as it occurs. Returns True if everything went well, returns false if anything failed in the middle.
> > (lines 180-200)
> > Those functions would have a more clear responsibility.
> 
> I can make them private local sub-functions if you think that would be better. The real trouble is that subfunction 1 is asynchronous but we want to force it to lock continuation of a script that would then run the full data conversion the moment the transfer completes (assuming `success=True`, else need manual intervention). So they do need to go hand-in-hand for the primary function call.

I think that making them private local sub-functions would at least help from a readability point of view. For the case that you describe, could you not lock the start of the conversion to the success of the hypothetical sub_function 2?

CodyCBakerPhD:
> I think that making them private local sub-functions would at least help from a readability point of view. For the case that you describe, could you not lock the start of the conversion to the success of the hypothetical sub_function 2? the second function would be a tracking a function (with the option of displaying progress) which will liberate the script once all the transfers are succesfully.

Gotta remember to do that though from an API point of view. Much safer to not leave room for that error on the user side (mostly my side, in this case). This also puts the behavior closer to the recently added figshare download tools.

h-mayorquin:
> > I think that making them private local sub-functions would at least help from a readability point of view. For the case that you describe, could you not lock the start of the conversion to the success of the hypothetical sub_function 2? the second function would be a tracking a function (with the option of displaying progress) which will liberate the script once all the transfers are succesfully.
> 
> Gotta remember to do that though from an API point of view. Much safer to not leave room for that error on the user side (mostly my side, in this case). This also puts the behavior closer to the recently added figshare download tools.

Fair enough, I still favor the approach of using two internal private methods or at least adding comments at the beginning of each of the indicated sections for improving readability. 

Something that surfaced form discussion, it seems that you are tying your locking mechanism to display progress. If the function needs to be locked in some specific context maybe you want to separate display from locking? I think that the function will be easier to document with that distinction.

CodyCBakerPhD:
> Fair enough, I still favor the approach of using two internal private methods or at least adding comments at the beginning of each of the indicated sections for improving readability.

Already done!

>  it seems that you are tying your locking mechanism to display progress. If the function needs to be locked in some specific context maybe you want to separate display from locking? I think that the function will be easier to document with that distinction.

Are you sure about that?

CodyCBakerPhD:
> Are you sure about that?

Trick question, I started it out that way but realized exactly what you are saying, so I fixed it. 😉

h-mayorquin:
> > Are you sure about that?
> 
> Trick question, I started it out that way but realized exactly what you are saying, so I fixed it. wink

: D
Yep, I looked at the old function and I had that in my mind when I was writing. 

Two minor details: 
1) The first private method is using the outer scope variable `destination_folder_path` instead of the local one `destination_folder` inside of the first sub_function. I think that the outer scope name is better btw. 
2) The second private method is usign the outer scope `task_total_sizes` instead of the intended `tasks_ids` which is not used.

CodyCBakerPhD:
> Two minor details:
>
> The first private method is using the outer scope variable destination_folder_path instead of the local one destination_folder inside of the first sub_function. I think that the outer scope name is better btw.
The second private method is usign the outer scope task_total_sizes instead of the intended tasks_ids which is not used.

Good catches! Thanks

bendichter:
Why are `_submit_transfer_request` and `_track_transfer` defined within the scope of `transfer_globus_content`?

CodyCBakerPhD:
> Why are _submit_transfer_request and _track_transfer defined within the scope of transfer_globus_content?

See discussion with Heberto, who requested this for readability

bendichter:
I feel like marking the functions as private with a leading underscore is good enough. I generally only define a function within another function if I want to share scope. But this is more of a stylistic thing, and if this way is easier for you to understand then I think it's fine

bendichter (approved):

src/nwb_conversion_tools/tools/data_transfers.py
```diff
@@ -61,14 +63,175 @@ def get_globus_dataset_content_sizes(
...
+        source_endpoint_id: str,
+        source_files: Union[str, List[List[str]]],
+        destination_endpoint_id: str,
+        destination_folder_path: Path,
+    ) -> Dict[str, int]:
+        """Send transfer request to Globus."""
+        folder_content_sizes = dict()
+        task_total_sizes = dict()
+        for j, batched_source_files in enumerate(source_files):
+            paths_file = destination_folder_path / f"paths_{j}.txt"
```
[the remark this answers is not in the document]
CodyCBakerPhD: I tried that at first however you are not guaranteed to have write permissions there. The function can only be guaranteed to have write permission at the destination folder path. I mean, I could make a temporary sub-directory there but the effect would be the same

CodyCBakerPhD:
> I feel like marking the functions as private with a leading underscore is good enough. I generally only define a function within another function if I want to share scope. But this is more of a stylistic thing, and if this way is easier for you to understand then I think it's fine

It's entirely for style - I'm personally fine with not even using sub-functions at all, but @h-mayorquin found it stylistically easier to understand.

It is perhaps conceivable that I might eventually want the task tracking to be available as a global function, I'll have to see after getting some practical use out of this. But I could also see a class-based approach being the superior API for all of this, so rearranging stuff would be best done then.

h-mayorquin:
> > I feel like marking the functions as private with a leading underscore is good enough. I generally only define a function within another function if I want to share scope. But this is more of a stylistic thing, and if this way is easier for you to understand then I think it's fine

I actually would have preferred to have the functions as independent private methods. I somehow misunderstood that this was not doable for locking reasons.

PR #496, opened 2022-05-03
Make nwbfile argument optional in `run_conversion` for recording and sorting interfaces

h-mayorquin:
Simple change that allows recording and sorting interfaces to run without a `nwbfile` by making the argument optional.

My understanding is that @bendichter used the same pattern in the ophys module for the current fromeke conversion.

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,4 +1,4 @@
-dandi==0.36.0
+dandi @ git+https://github.com/dandi/dandi-cli.git@9a749401aa2dec2ba59f352205b02a3d8083b1d8  # that uses schema v0.7.x
 pynwb==2.0.0
 hdmf==3.2.1
 tqdm==4.60.0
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD (dismissed):

h-mayorquin:
There are some concurrent issues with the dandi uploads it seems. Any idea on what this could be, @CodyCBakerPhD ?

CodyCBakerPhD:
> There are some concurrent issues with the dandi uploads it seems. Any idea on what this could be, @CodyCBakerPhD ?

I'm working on it. Doesn't seem like there's a new release of the dandischema even though the server expects it

CodyCBakerPhD:
IDK if we can't get it working maybe we'll just wait on https://github.com/dandi/dandi-cli/pull/987#issuecomment-1116338126

CodyCBakerPhD (approved):

PR #495, opened 2022-05-03
Conversion examples in the documentation and docstrings

h-mayorquin:
## Motivation

@bendichter mentioned in of our team meetings that it would be an idea to have a gallery of examples that we could quicly point out to the users. Also, we have issue #442 to add examples in the docstrings of non-data. This PR is to test whether it is feasible to integrate both approaches and at the same subject the docstring and the example gallery to testing suite using the capabilities provided by `doctest`:

To-do:
- [x] Add an example to the gallery to showcase format and idea.
- [x] Test locally with test.
- [x] Add tests to CI.

After the gallery examples are done we can add links to each of the docstrings in their respective data-interfaces so the examples are accessible from there.

PR #498, opened 2022-05-04
Documentation: Pip installable environment and build instructions

h-mayorquin:
As we did for the testing, I am adding a pip installable environment for the documentation. Also, adding instructions in how to build the documentation locally to the readme file.

Edit:
This is failing the build in read the docs. @CodyCBakerPhD, @bendichter can I get admin rights over there to check and understand the build process? This will also be useful for #495 .


bendichter (approved):
looks great

CodyCBakerPhD:
The traceback at https://readthedocs.org/projects/nwb-conversion-tools/builds/16820312/ has the entire collection of workflow actions and outputs, which cannot be altered.

The only thing we can control as far as dependencies is what goes into the `requirement-rtd.txt` folder, which it uses when calling `pip install -r requirements-rtd.txt` right before `pip install .`

We can technically allow access to the 'global site-packages' on readthedocs, which is essentially all of the most recent versions of the most common deps, but I advise against this since it can cause many dependency conflicts. We can also disable the installation of the package altogether, but that seems the opposite of what we want to do here.

CodyCBakerPhD:
@h-mayorquin So basically, if you want to fix the docs build (necessary before merging), I'd suggest restoring the state of the `requirements-rtd.txt`, which has been configured to what the readthedocs CI currently requires.

Note that as of today the ephy testing data issued a new hash which triggered cache updates, which triggered an issue with the MacOS latest version of git-annex. Some basic attempts to fix it have failed so I'll just raise an issue and work on it elsewhere - temporarily disabling full tests on Mac until then.

h-mayorquin:
> The traceback at https://readthedocs.org/projects/nwb-conversion-tools/builds/16820312/ has the entire collection of workflow actions and outputs, which cannot be altered.

Yup, I am aware of this.

> The only thing we can control as far as dependencies is what goes into the `requirement-rtd.txt` folder, which it uses when calling `pip install -r requirements-rtd.txt` right before `pip install .`
> 
> We can technically allow access to the 'global site-packages' on readthedocs, which is essentially all of the most recent versions of the most common deps, but I advise against this since it can cause many dependency conflicts. We can also disable the installation of the package altogether, but that seems the opposite of what we want to do here.

So in the advanced seetings, we have requirements file pointed to `requirements-rdt.txt`. Do we have the `install project` option activated? 
image

By looking at the trace it seems that we do.

h-mayorquin:
> @h-mayorquin So basically, if you want to fix the docs build (necessary before merging), I'd suggest restoring the state of the `requirements-rtd.txt`, which has been configured to what the readthedocs CI currently requires.

Check out the [configuration file](https://docs.readthedocs.io/en/stable/config-file/v2.html) that I added:
https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python
It seems that it gives us enough control of python version to allow a more similar build to one that we have locally without requiring [access]('global site-packages' ) to the 'global site-packages'  which is what you wanted to avoid. I think this is good because it allows us to eliminate the hardcoded dependencies so we can rely only the ones that we have in `requirements-minimal.txt` avoiding duplication.

> Note that as of today the ephy testing data issued a new hash which triggered cache updates, which triggered an issue with the MacOS latest version of git-annex. Some basic attempts to fix it have failed so I'll just raise an issue and work on it elsewhere - temporarily disabling full tests on Mac until then.

Is always the MacOS, eh?

CodyCBakerPhD (approved):

PR #502, opened 2022-05-06
Icephys restore2

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist


src/nwb_conversion_tools/utils/json_schema.py
```diff
@@ -12,7 +12,7 @@
 
 def get_base_schema(tag=None, root=False, id_=None, **kwargs) -> dict:
     """Return the base schema used for all other schemas."""
-    base_schema = dict(required=[], properties={}, type="object", additionalProperties=False)
+    base_schema = dict(required=[], properties={}, type="object", additionalProperties=True)
```
CodyCBakerPhD, line 15: This one line can actually have  variety of unintended consequences since it affects the base of every schema.

Is this actually needed, or can you just set `additionalProperties=True` in the one part of the sub-schema where it is actually needed?

src/nwb_conversion_tools/tools/neo/neo.py
```diff
@@ -0,0 +1,602 @@
...
+    elif unit in ["uA", "uV"]:
+        conversion = 1e-6
+    elif unit in ["mA", "mV"]:
+        conversion = 1e-3
+    elif unit in ["A", "V"]:
+        conversion = 1.0
+    else:
+        conversion = 1.0
+        warnings.warn("No valid units found for traces in the current file. Gain is set to 1, but this might be wrong.")
+    return float(conversion)
```
CodyCBakerPhD, line 118: I had asked Luiz to this before, but this logic seems very generally applicable (not specific to neo). Could you make it a `util` function instead?

src/nwb_conversion_tools/tools/neo/neo.py
```diff
@@ -0,0 +1,602 @@
...
+    cmd_channel : int, optional
+        ABF command channel (0 to 7). Defaults to 0.
+    """
+    try:
+        traces, titles, units = neo_reader.read_raw_protocol()
+        return traces[segment][cmd_channel], titles[segment][cmd_channel], units[segment][cmd_channel]
+    except Exception as e:
+        msg = ".\n\n WARNING - get_command_traces() only works for AxonIO interface."
+        e.args = (str(e) + msg,)
+        return e
```
CodyCBakerPhD, line 90: Did we ever get an answer on this one? What can go wrong, why, and is there not a more specific error like `NotImplementedError`?

src/nwb_conversion_tools/tools/neo/neo.py
```diff
@@ -0,0 +1,602 @@
...
+            response_unit = neo_reader.header["signal_channels"]["units"][ei]
+            response_conversion = get_conversion_from_unit(unit=response_unit)
+            response_gain = neo_reader.header["signal_channels"]["gain"][ei]
+            response_name = f"{icephys_experiment_type}-response-{si + 1 + simultaneous_recordings_offset:02}-ch-{ei}"
+
+            response = response_classes[icephys_experiment_type](
+                name=response_name,
+                description=f"Response to: {session_stimulus_type}",
+                electrode=electrode,
+                data=neo_reader.get_analogsignal_chunk(block_index=0, seg_index=si, channel_indexes=ei),
```
CodyCBakerPhD, line 320: Skipping iterative write is fine as long as we can guarantee that in practice, such files are never more than 16GB.

We should enable compression, though, just wrap in an `H5DataIO` and pass the compression options at https://github.com/catalystneuro/nwb-conversion-tools/pull/502/files#diff-c2e3bb3498bea80fe7bf252c1c7cda1365a1aa1fb0f4ba8fe3f5fece8d16e0bfR378-R379

src/nwb_conversion_tools/tools/neo/neo.py
```diff
@@ -0,0 +1,602 @@
...
+    nwbfile=None,
+    use_times: bool = False,
+    metadata: dict = None,
+    write_as: str = "raw",
+    es_key: str = None,
+    write_scaled: bool = False,
+    compression: Optional[str] = "gzip",
+    compression_opts: Optional[int] = None,
+    iterator_type: Optional[str] = None,
+    iterator_opts: Optional[dict] = None,
```
CodyCBakerPhD, line 381: Don't need these if we're not using iterative write.

CodyCBakerPhD: This function is actually taking a lot of arguments that are spikeinterface-reminiscent but complete unused here, can you review them and remove the unused?

CodyCBakerPhD:
For the sake of getting this PR through, willing to drop the request of centralizing metadata schema definitions.

A few comments, mostly minor except for changes to base util behavior and enabling of compression.

src/nwb_conversion_tools/utils/json_schema.py
bendichter, line 15:
```suggestion
    base_schema = dict(required=[], properties={}, type="object", additionalProperties=False)
```

src/nwb_conversion_tools/tools/neo/neo.py
bendichter: we could but let's wait until we need it somewhere else

src/nwb_conversion_tools/tools/neo/neo.py
bendichter: I removed these unused args in the latest commit

src/nwb_conversion_tools/tools/neo/neo.py
bendichter: added compression

src/nwb_conversion_tools/tools/neo/neo.py
bendichter: also, I don't see anywhere were this function is called. @luiztauffer, what is the intended purpose of this function? How would it ideally fit into a conversion workflow?

luiztauffer: this is probably from an older version and its use has been replaced
the `read_raw_protocol` is specific for Icephys, and other NEO readers would not work with it. This function would unpack the content, which is basically traces, title and unit for the stimulation signals
this could be useful I guess, but right now we're not using it, so we could remove it if you prefer

src/nwb_conversion_tools/datainterfaces/icephys/abf/abfdatainterface.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # LabMetadata
+        if any(x in metafile_data for x in ("cell_id", "slice_id", "targeted_layer", "inferred_layer")):
+            metadata["ndx-dandi-icephys"] = dict(
+                # Required fields for DANDI
+                cell_id=metafile_data.get("cell_id"),
+                slice_id=metafile_data.get("slice_id"),
+                # Lab specific metadata
+                targeted_layer=metafile_data.get("targeted_layer", "unknown"),
+                inferred_layer=metafile_data.get("inferred_layer", "unknown"),
```
bendichter, line 91:
```suggestion
                targeted_layer=metafile_data.get("targeted_layer"),
                inferred_layer=metafile_data.get("inferred_layer"),
```

bendichter:
@CodyCBakerPhD I don't think the tests are running. Do you know why that might be?

CodyCBakerPhD:
@ben Tests ran before the black changes: https://github.com/catalystneuro/nwb-conversion-tools/runs/6374151347?check_suite_focus=true will approve and merge assuming the remaining ones pass

CodyCBakerPhD (approved):

PR #505, opened 2022-05-06
Allow sorting interface to `run_conversion` without a nwbfile

h-mayorquin:
## Motivation
Similar to #503 and to #496. I am building examples in #495 that require calling the `run_conversion` directly from the data interface. As things stand, to do this with a sortingextractor interface it is necessary to pre define an nwbfile. This PR solves this by adding to the function an argument for `save_path`.


CodyCBakerPhD:
Fine for now, just keep in mind we should keep it in the back of our minds to unify the keyword arguments between data interfaces and the main converter to be as consistent as possible

CodyCBakerPhD (approved):

PR #510, opened 2022-05-09
Metadata to handle `datetime` objects. 

### LINKED ISSUE #463: Metadata is not able to handle session_start_time as a datetime object.
I was discussing this with @CodyCBakerPhD in our catch up last week. Here is a minimal example:

```python
from datetime import datetime
from pathlib import Path
from nwb_conversion_tools import NWBConverter, RecordingTutorialInterface


class TutorialNWBConverter(NWBConverter):
        data_interface_classes = dict(
            RecordingTutorial=RecordingTutorialInterface
        )

duration = 10.0  # Seconds
num_channels = 4
num_units = 10
sampling_frequency = 30000.0  # Hz
stub_test = False
test_dir = Path(".")
output_file = str(test_dir / "TestTutorial.nwb")
source_data = dict(
    RecordingTutorial=dict(duration=duration, num_channels=num_channels, sampling_frequency=sampling_frequency),
)
converter = TutorialNWBConverter(source_data=source_data)
metadata["NWBFile"]["session_start_time"] = datetime.now().astimezone()
conversion_options = dict(RecordingTutorial=dict(stub_test=stub_test))
converter.run_conversion(
    metadata=metadata,
    nwbfile_path=output_file,
    save_to_file=True,
    overwrite=True,
    conversion_options=conversion_options,
)
```
And here is the trace:

```python
[ValidationError: datetime.datetime(2022, 4, 18, 16, 55, 54, 179964, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200), 'CEST')) is not of type 'string'

Failed validating 'type' in schema['properties']['NWBFile']['properties']['session_start_time']:
    {'description': 'the start date and time of the recording session',
     'format': 'date-time',
     'type': 'string'}

On instance['NWBFile']['session_start_time']:
    datetime.datetime(2022, 4, 18, 16, 55, 54, 179964, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200), 'CEST'))]()
```

h-mayorquin:
This should close #463.

Right now, this is the approach suggested in the issue. However -and as usual- the ophys classes give problems. The problem is that some of the unities in the metadata are non standard numeric quantities: `np.int16` and they can't be encoded by default. This seems patchable but I wanted to discuss before looking deeper into this issue.

Other downsides of these approach is that we have to encode and then re-encode. Maybe there is a way to avoid that? 

A simpler version of this that avoids this problems is in #509


CodyCBakerPhD:
I think it would be worth looking into, it's only 2 of the ophys interfaces and it likely involves a field that gets set during the tests. Specifically, not bad to include suggestions from https://stackoverflow.com/questions/11942364/typeerror-integer-is-not-json-serializable-when-serializing-json-in-python

such as 

```python
def convert(o):
    if isinstance(o, any_int_type): return int(o)  
    raise TypeError
```

for `any_int_type` either the subclass method you used in the recent table refactors, or maybe `from numbers import ?` has something for that?

h-mayorquin:
> I think it would be worth looking into, it's only 2 of the ophys interfaces and it likely involves a field that gets set during the tests. Specifically, not bad to include suggestions from https://stackoverflow.com/questions/11942364/typeerror-integer-is-not-json-serializable-when-serializing-json-in-python
> 
> such as
> 
> ```python
> def convert(o):
>     if isinstance(o, any_int_type): return int(o)  
>     raise TypeError
> ```
> 
> for `any_int_type` either the subclass method you used in the recent table refactors, or maybe `from numbers import ?` has something for that?

What do you think of the encoding and de-coding of the metadata?
Do you think there is a way to avoid that?

h-mayorquin:
So, will the latest changes all the tests are passing. 
I still feel that the following piece of code is unnecessary complexity at this point:

https://github.com/catalystneuro/nwb-conversion-tools/blob/40c2de3eb3ee1124572dae9705c4f235a5485a84/src/nwb_conversion_tools/nwbconverter.py#L110-L114

Do we gain anything else other than solving this problem by doing it this way?

CodyCBakerPhD:
> Do we gain anything else other than solving this problem by doing it this way?

It's doing the same thing as any other solution, including the 'simple' one of https://github.com/catalystneuro/nwb-conversion-tools/pull/509.

Difference being, with this approach we can integrate it more seamlessly into the `utils.json_schema` instead of in the NWBConverter class or anywhere else so it can operate completely under the hood rather than modifying elements on the metadata dictionary in-place. Also places no restrictions on what the metadata field contents are, just tells JSON how to read them.

bendichter (approved):

PR #513, opened 2022-05-10
[Bug] Only run DANDI test if key is present

### LINKED ISSUE #497: `test_dandi_upload` should take into account if environment variable is present or not
The following test is failing in my system because i don't have the dandi key set:
https://github.com/catalystneuro/nwb-conversion-tools/blob/d2650eef7fce8b6973b8a48f5ec0df0239d8033e/src/nwb_conversion_tools/tools/data_transfers.py#L181-L184

Test should not fail in expected behavior. I think that:
1) The test should be skipped if the environment variable is not present, or/and
2) The machinery described in the following should be actually tested:
https://github.com/catalystneuro/nwb-conversion-tools/blob/d2650eef7fce8b6973b8a48f5ec0df0239d8033e/src/nwb_conversion_tools/tools/data_transfers.py#L181-L184

CodyCBakerPhD:
fix #497


tests/test_internals/test_tools.py
```diff
@@ -26,6 +26,7 @@
         LOGGED_INTO_GLOBUS = False
 except ModuleNotFoundError:
     HAVE_GLOBUS, LOGGED_INTO_GLOBUS = False, False
+HAVE_DANDI_KEY = "DANDI_API_KEY" in os.environ
```
h-mayorquin, line 29: Nice, I was wondering how to do that. Did not know it was that straightforward.

h-mayorquin (approved):
Thanks!

PR #525, opened 2022-05-13
[Refactor] standardize sorting interface kwargs

CodyCBakerPhD:
Continuation of #516 for sorting interfaces.


CodyCBakerPhD:
@h-mayorquin The main reason I have to set `verbose=False` in all the gallery docs is because it otherwise expects no output when the `NWBFile saved at...` message outputs. Is it possible to tell the output of  `>>>` blocks to expect some type of regular expression? (the exact name of the files in the tests belongs to a temporary directory, and local output would of course differ from one person's system to the next)

h-mayorquin:
> @h-mayorquin The main reason I have to set `verbose=False` in all the gallery docs is because it otherwise expects no output when the `NWBFile saved at...` message outputs. Is it possible to tell the output of `>>>` blocks to expect some type of regular expression? (the exact name of the files in the tests belongs to a temporary directory, and local output would of course differ from one person's system to the next)

Not as far as I know. I think that we should keep verbose=False here, we can also use the skiptest functionallity:
https://docs.pytest.org/en/6.2.x/doctest.html#skipping-tests

CodyCBakerPhD:
> Another option might be to we use the skiptest functionality:
https://docs.pytest.org/en/6.2.x/doctest.html#skipping-tests

But wouldn't skipping it prevent the line from getting run (since the results of that line are needed for the next lines)?

I'd prefer to not have to specify it at all in the minimal copy/paste examples since it begs the question of why we're changing it there to begin with.

src/nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -99,26 +101,37 @@ def subset_sorting(self):
...
-        metadata: dict = None,
-        stub_test: bool = False,
-        save_path: OptionalFilePathType = None,
+        nwbfile_path: OptionalFilePathType = None,
+        nwbfile: Optional[NWBFile] = None,
+        metadata: Optional[dict] = None,
         overwrite: bool = False,
+        stub_test: bool = False,
         write_ecephys_metadata: bool = False,
+        save_path: OptionalFilePathType = None,  # TODO: to be removed
```
h-mayorquin, line 110: Maybe adding the return to docstring?
Is the only thing I could find.

h-mayorquin (approved):
It seems that it was a lot of busy work to "verbosify" everything  : )

src/nwb_conversion_tools/datainterfaces/ecephys/basesortingextractorinterface.py
CodyCBakerPhD: Looks like that's missing from all the others, too - might as well do them all in one-shot in a follow up

CodyCBakerPhD:
> It seems that it was a lot of busy work to "verbosify" everything

Yeah, almost regretting making it optional.

PR #526, opened 2022-05-13
[Refactor] standardize LFP conversion kwargs

CodyCBakerPhD:
Continuation of #516 for LFP interfaces.

Removal of docstring allows it to completely inherit the one from the parent, which is identical.

### DIFF
--- a/src/nwb_conversion_tools/datainterfaces/ecephys/baselfpextractorinterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ecephys/baselfpextractorinterface.py
@@ -30,78 +30,37 @@ class BaseLFPExtractorInterface(BaseRecordingExtractorInterface):
 
     def run_conversion(
         self,
-        nwbfile: NWBFile,
-        metadata: dict = None,
+        nwbfile_path: OptionalFilePathType = None,
+        nwbfile: Optional[NWBFile] = None,
+        metadata: Optional[dict] = None,
+        overwrite: bool = False,
         stub_test: bool = False,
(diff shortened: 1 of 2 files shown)


h-mayorquin (approved):
Looks good to me. 

I did not know that doc-strings were inherited from parent if absent. I am mildly concern that when people read the code from an editor they will now need to navigate to the parent class if they want to see the docstring. However, I am also happy about the code reduction. Your call.

CodyCBakerPhD:
> I did not know that doc-strings were inherited from parent if absent. I am mildly concern that when people read the code from an editor they will now need to navigate to the parent class if they want to see the docstring.

TBH even in an editor they should have an iPython somewhere in the IDE - reading the code is not the most recommended way to parse docstrings, the `?` operator or API readthedocs are (*e.g.* `NeuroscopeLFPInterface.run_conversion?`).

CodyCBakerPhD:
> I did not know that doc-strings were inherited from parent if absent. I am mildly concern that when people read the code from an editor they will now need to navigate to the parent class if they want to see the docstring.

What I would *really* like, and maybe you can look into it once you've cleared out some of your many high-priority tasks, is if there's any way we can efficiently cross-inherit or inter-define docstrings. That is, you may have noticed that many of the same keywords from the new context class down to the tools and interfaces are copying/pasting the same blocks of text. It would be nice if those blocks only had to be written in one place and then linked when needed (especially nice for fixing typos instead of having to hunt them down everywhere that re-uses them).

PR #527, opened 2022-05-15
Add timestamps for MovieDataInterface

weiglszonja:
## Motivation

MovieDataInterface currently does not support creating an ImageSeries object with predefined timestamps. 
The aim of this PR is to run conversion with the capability of supplying a list of timestamps for the movie. 
Based on:
catalystneuro/fee-lab-to-nwb#7

## How to test the behavior?

There is a new test created for having timestamps:

`test_movie_interface.TestMovieInterface.test_movie_timestamps`

## Checklist


src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -213,8 +216,11 @@ def run_conversion(
             if external_mode:
                 with VideoCaptureContext(str(file_list[0])) as vc:
                     fps = vc.get_movie_fps()
+                    if timestamps is None:
+                        timestamps = starting_times[j] + vc.get_movie_timestamps()
```
CodyCBakerPhD, line 220: If passing `timestamps`, `starting_times` should not be passed (one-or-the-other kind of thing). The `timestamps` values should be incremented to the `session_start_time` before passing into here.

src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -287,10 +293,11 @@ def run_conversion(
...
                 image_series_kwargs.update(data=data)
-                if check_regular_series(series=timestamps):
-                    image_series_kwargs.update(starting_time=starting_times[j], rate=fps)
-                else:
-                    image_series_kwargs.update(timestamps=timestamps)
+
+            if check_regular_series(series=timestamps):
+                image_series_kwargs.update(starting_time=starting_times[j], rate=fps)
+            else:
+                image_series_kwargs.update(timestamps=timestamps)
```
CodyCBakerPhD, line 300: This is great, you read my mind!

Although... I wonder if maybe also we should add a check that internally recorded `fps` value is equal to `timestamps[1] - timestamps[0]` after knowing that the `timestamps` are regular? If unequal, perhaps warn the user and use the timestamps rate since they thought ahead enough to specify that separately?

tests/test_internals/test_movie_interface.py
```diff
@@ -218,3 +218,24 @@ def test_movie_stub(self):
             for no in range(len(metadata["Behavior"]["Movies"])):
                 movie_interface_name = metadata["Behavior"]["Movies"][no]["name"]
                 assert mod[movie_interface_name].data.shape[0] == 10
+
+    def test_movie_timestamps(self):
+        timestamps = list(np.linspace(0.0, 1.16, 30)) + list(np.linspace(50.0, 51.16, 30))
```
CodyCBakerPhD, line 223: I love the approach here, in general the tests look great!

One thing, though, that we've been getting more into the habit of doing (since the tests themselves take so long to run) is using only a minimal amount of data in all test cases.

So here, we could simplify the testing timestamps to be simply `timestamps = [1, 2, 4]`

weiglszonja: Thanks for the feedback! Yeah, I'll change this.

src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
weiglszonja: Thanks! I added a new check, and modified this part accordingly. I also added a new unittest to test this behavior, let me know if you think it's unnecessary or you were thinking about something different!

tests/test_internals/test_movie_interface.py
```diff
@@ -218,3 +218,46 @@ def test_movie_stub(self):
...
+            overwrite=True,
+            conversion_options=conversion_opts,
+            metadata=self.metadata,
+        )
+
+        with NWBHDF5IO(path=self.nwbfile_path, mode="r") as io:
+            nwbfile = io.read()
+            mod = nwbfile.acquisition
+            metadata = self.nwb_converter.get_metadata()
+            for no in range(len(metadata["Behavior"]["Movies"])):
```
CodyCBakerPhD, line 260: I know you might see this scattered around this interface (it was a bad habit of Sakshams), but I would not use 'no' as an abbreviation of 'number'. I'd just use 'num' or 'number' when referring to a total quantity and 'idx' or 'index' when iterating a list numerically.

src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD: Something I particularly love to see when adding warnings is usage of the `self.assertWarnsWith` and `self.assertRaisesWith` functions of the `TestCase` class imported from `hdmf.testing`. This lets you check the exact text that is expected in the raising condition.

tests/test_internals/test_movie_interface.py
```diff
@@ -218,3 +218,46 @@ def test_movie_stub(self):
...
+            overwrite=True,
+            conversion_options=conversion_opts,
+            metadata=self.metadata,
+        )
+
+        with NWBHDF5IO(path=self.nwbfile_path, mode="r") as io:
+            nwbfile = io.read()
+            acquisition_module = nwbfile.acquisition
+            metadata = self.nwb_converter.get_metadata()
+            for movie_num in range(len(metadata["Behavior"]["Movies"])):
```
h-mayorquin, line 269: To get the interface name:

```suggestion
    for movie_metadata in metadata["Behavior"]["Movies"]:
        movie_interface_name = movie_metadata["name"]
```

src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
```diff
@@ -287,10 +293,17 @@ def run_conversion(
...
-                else:
-                    image_series_kwargs.update(timestamps=timestamps)
+
+            is_regular, rate = check_regular_series(series=timestamps)
+            if is_regular:
+                if fps != rate:
+                    warn(
+                        f"The fps={fps} from movie data is unequal to the difference in"
+                        f"regular timestamps. Using fps={rate} from timestamps instead."
+                        , UserWarning)
```
weiglszonja: @CodyCBakerPhD I'm not sure why but this warning doesn't get triggered, even when i'm in debug mode and put a breakpoint at this line it gets there, but no warning is triggered -- the unittest is failing currently. Do you have any idea why?

tests/test_internals/test_movie_interface.py
weiglszonja: Thanks @h-mayorquin !

requirements-rtd.txt
```diff
@@ -3,5 +3,5 @@ Jinja2<3.1
 sphinx==3.5.1
 sphinx_rtd_theme==0.5.1
 readthedocs-sphinx-search==0.1.0rc3
-sphinx-toggleprompt @ git+https://github.com/jurasofish/sphinx-toggleprompt@default_on_off 
+sphinx-toggleprompt==0.2.0
```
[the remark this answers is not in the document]
CodyCBakerPhD: @h-mayorquin They were amazingly fast on integrating that and releasing, always good to see!

src/nwb_conversion_tools/utils/checks.py
```diff
@@ -2,6 +2,9 @@
...
 def check_regular_series(series: np.ndarray, tolerance_decimals: int = 9):
-    """General purpose function for checking if the difference between all consecutive points in a series are equal."""
-    uniq_diff_ts = np.unique(np.diff(series).round(decimals=tolerance_decimals))
-    return len(uniq_diff_ts) == 1
+    """General purpose function for checking if the difference between all consecutive points in a series are equal.
+    If it is, returns the assessed sampling rate."""
+    diff_ts = np.diff(series).round(decimals=tolerance_decimals)
+    uniq_diff_ts = np.unique(diff_ts)
+    rate = diff_ts[0] if len(uniq_diff_ts) == 1 else None
+    return len(uniq_diff_ts) == 1, rate
```
CodyCBakerPhD, line 10: @h-mayorquin FYI this kind of thing is exactly why I wanted all of these util functions centralized somewhere - we had decided the NWBInspector was a good place to centralize such things since it's secondary dependencies are extremely minimal on top of PyNWB.

This type of added functionality would be useful to have over there, but now we have to both remember to update it in both places, which is a very easy thing to overlook, and takes twice as much work to integrate.

Would it bother you overly much if we just had a separate repo altogether, like `cn_utils`, that both the NWBInspector and NWBCT can both depend on?

tests/test_internals/test_movie_interface.py
```diff
@@ -218,3 +219,55 @@ def test_movie_stub(self):
...
+            f"Using fps={rate} from timestamps instead."
+        )
+
+        with NWBHDF5IO(path=self.nwbfile_path, mode="r") as io:
+            nwbfile = io.read()
+            acquisition_module = nwbfile.acquisition
+            metadata = self.nwb_converter.get_metadata()
+            for movie_metadata in metadata["Behavior"]["Movies"]:
+                movie_interface_name = movie_metadata["name"]
+                with self.assertWarnsWith(warn_type=UserWarning, exc_msg=expected_warn_msg):
```
CodyCBakerPhD, line 271: @weiglszonja Oh, my bad - these lines just check via read mode that the output values are what you expect

src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py
CodyCBakerPhD: @weiglszonja See comments below - the context must encapsulate the function call that actually gives the warning

src/nwb_conversion_tools/utils/checks.py
```diff
@@ -2,6 +2,9 @@
 
 
 def check_regular_series(series: np.ndarray, tolerance_decimals: int = 9):
```
h-mayorquin, line 4: Following the design principle that functions should do one thing I think that you should not return the rate. Calculating the rate is only one line (9) and can be done in-line when needed (or you can have another function if you think that's better). 

Also, maybe we can have a more specific name than "regular"? Is regular a terminology that is used in `pynwb`? Maybe `check_regular_rate`?

h-mayorquin:
Not directly related to this PR as it is solving a very specific issue but I feel that this function already has too much cylomatic complexity (lots of loops and ifs) and we are adding a bit more here. At some point, some refactoring to simplify this should be done, maybe having a `write_image_series` in `spikeinterface` for one image and loop over there.

CodyCBakerPhD:
> Not directly related to this PR as it is solving a very specific issue but I feel that this function already has too much cylomatic complexity (lots of loops and ifs) and we are adding a bit more here. At some point, some refactoring to simplify this should be done, maybe having a write_image_series in spikeinterface for one image and loop over there.

Sure, low priority, though.

src/nwb_conversion_tools/utils/checks.py
CodyCBakerPhD: The function already has implicit access to the exact piece of information needed, and it is still just doing one thing (the same thing it always did).

In particular, adding an extra return value does not break any compatibility since the original usage is not changed in any way, and the extra value is informative to parsing the result.

h-mayorquin: > The function already has implicit access to the exact piece of information needed, and it is still just doing one thing (the same thing it always did).

It is not doing one thing but two: It checks that the series has a constant rate (which is consistent with the name) and it calculate the rate (which is not consistent with the name).  

My argument is not about backwards compatibility which is fine as you are saying. Is about sticking to design standards to improve readability.

CodyCBakerPhD: >It is not doing one thing but two: It checks that the series has a constant rate (which is consistent with the name) and it calculate the rate (which is not consistent with the name).

And a function that calculates the rate would need to check if the series is regular. The thing you are 'doing' is checking if the series is regular. The function could likewise return the `rate` as a non-`None` value, and the boolean logic asserted as `rate is None`. The two outputs are inherently related.

h-mayorquin: Yes. But that way you separate the two things. I thinks is way better to definine an extra function `calculate_regular_sampling_rate` that returns the sampling rate if the series is regular and and assertion (preferred) or None if it is not. That function would be doing what it tells you it does. `check_regular_series` does an extra thing that is not evidence from the name. 

I feel that something being a necessary step for something else does not indicate the two things are the same and readability trumps line reduction.

CodyCBakerPhD: > I feel that something being a necessary step for something else does not indicate the two things are the same and readability and clarity of intent trumps line reduction.

@h-mayorquin Please explain how

```python
is_regular, rate = check_regular_series(series=series)
```
is not both readable and minimal.

Since we're apparently at a stalemate, maybe let @weiglszonja decide between the two since it's her PR.

h-mayorquin: I think this is good. Here is a rehash of my claims and my issue with the current approach:
* It does not follow the "functions should do one thing principle". The reason here is that the function `check_regular_series` does two things: 1) it check if the series is regular, 2) it calculates the rate. 
* The second claim has to do with readability and that function naming should reflect intent and be transparent. Right now the function `check_regular_series`  does a second thing (calculating the rate) that is not reflected in the name of the function. This makes it harder to read as it executes a non stated operation (calculating the rate) and also assumes that the users know that the rate is to be returned as the second argument (that is, an implicit contract). I think this is bad because it forces knowledge into the users and also reduces discoverability (i.e. if we had a function `calculate_series_rate` it is obvious what it does just from the name if you are looking for that functionality).
* I am proposing that we define another function called `calculate_series_rate` as this name is clear in intent. I think this is more readable and more flexible as there are situations where we might need to do the first thing `check_regular_series` but not the second `calculate_regular_rate`.  I agree that the second function might as well use the former (for example if you want to throw an assertion error or a warning) and that's fine. Functions can support each other.

Overall, my feeling is that you oppose this because you don't want to add more lines. That's why I said that "readability and clarity of intent trumps line reduction". Besides that, I am failing to imagine other argument to not separate them. Maybe I am missing another concern of yours?

CodyCBakerPhD: > It does not follow the "functions should do one thing principle". The reason here is that the function check_regular_series does two things: 1) it check if the series is regular, 2) it calculates the rate.

What I'm trying to show you is these two things _are_ the same thing, it's in fact how (1) determines regularity. It calculates the first-order difference (in nomenclature, 'rate' is the first-order difference of time) of the series and evaluates if they are all the same value (*i.e.*, 'constant', therefore 'regular').

Now if we have a function `calculate_series_rate` that returns a scalar when the series is regular and `None` if not, then the existence of another separate function `check_regular_series`, which is simply equivalent to the boolean evaluation `calculate_series_rate(series=series) is not None`, becomes redundant as a function unto itself, because the boolean evaluation is sufficient anywhere that logic is needed.

You say it helps discoverability, but that's heavily contingent on documentation and otherwise is a slippery slope to having a large number of functions spread across multiple projects that all evaluate similar if not identical logic, which leads to high redundancy and low efficiency, which just makes more work for ourselves regarding maintenance. That also harkens to my point of centralizing such helper functions, as a change to even a single line amplifies the amount of work we have to do across projects, which reduces our billable productivity.

h-mayorquin: I understand how the calculation rate is done. You are saying that the two things are the same if one is required in the calculation of the other. I disagree with that. One operation is a boolean, the other returns a number. I think that's the crux of the issue. You are lumping all of that in the box of "things required to calculate the rate" which produces two things as an output and the name you use for that box is `check_regular_series`. I am advocating for more precision. 

In that regard, I am totally fine with the second approach you propose above. It appears to me that is reasonable that a `calculate_series_constant_rate` function returns None if the function is not regular and returns the rate (as expected from the name) if it exists. It avoids the need of a a function just to check for regularity and another for the rate which you want to avoid. I think that seems like a good compromise.

For the discoverability I was thinking on something simpler (doing a search function for rate in the repo which what I do when I am looking for functionality) and what a crawler will get in the read the docs. That said, I don't think that discoverability is a particularly strong argument here.

CodyCBakerPhD: We'll go for the revised option, then, unless @weiglszonja has a preference for the other ways

```python
def calculate_regular_series_rate(series) -> Optional[Real]:
    diff_ts = np.diff(series).round(decimals=tolerance_decimals)
    uniq_diff_ts = np.unique(diff_ts)
    rate = diff_ts[0] if len(uniq_diff_ts) == 1 else None
    return rate
```

and regularity inferred from this output being `not None`

weiglszonja: @CodyCBakerPhD, @h-mayorquin I read through your comments, it was very insightful, I would go for the revised option as well.

tests/test_internals/test_movie_interface.py
```diff
@@ -218,3 +219,54 @@ def test_movie_stub(self):
...
+            metadata=self.metadata,
+        )
+
+        with NWBHDF5IO(path=self.nwbfile_path, mode="r") as io:
+            nwbfile = io.read()
+            acquisition_module = nwbfile.acquisition
+            metadata = self.nwb_converter.get_metadata()
+            for movie_metadata in metadata["Behavior"]["Movies"]:
+                movie_interface_name = movie_metadata["name"]
+                np.array_equal(timestamps, acquisition_module[movie_interface_name].timestamps[:])
```
h-mayorquin, line 242: Does this work?
Shouldn't it be?
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

tests/test_internals/test_movie_interface.py
```diff
@@ -218,3 +219,54 @@ def test_movie_stub(self):
...
+        )
+
+        rate = np.round(timestamps[1] - timestamps[0], 9)
+        fps = 25.0
+        expected_warn_msg = (
+            f"The fps={fps} from movie data is unequal to the difference in "
+            f"regular timestamps. Using fps={rate} from timestamps instead."
+        )
+
+        with self.assertWarnsWith(warn_type=UserWarning, exc_msg=expected_warn_msg):
```
h-mayorquin, line 257: Maybe this should be a separate test exclusive for the warning?

h-mayorquin (commented):
Looks great. One suggestion and one comment.

PR #531, opened 2022-05-23
Add `.yaml` to accepted yaml extensions in `load_dict_from_file` function

h-mayorquin:
## Motivation

Quick PR as described in the title. I also changed one of the test files to the new accepted extension to test the behavior.

PR #532, opened 2022-05-24
Small refactor of `test_gin_behavior` to pin new CI error

h-mayorquin:
The goal of this PR is to pin down the error that we are getting in #531.

I did a small refactor in the test that is failing to see if the error still appears / is reproducible here.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -49,53 +49,53 @@ jobs:
           git config --global user.name "CI Almighty"
       - name: Install nwb-conversion-tools with minimal requirements
         run: pip install .[test]
-      - name: Run minimal tests
-        run: pytest tests/test_internals -n auto --dist loadscope
+      # - name: Run minimal tests
+      #   run: pytest tests/test_internals -n auto --dist loadscope
       - name: Install full requirements (-e needed for codecov report)
         run: pip install -e .[full]
 
-      - name: Get ephy_testing_data current head hash
-        id: ecephys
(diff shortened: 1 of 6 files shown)


h-mayorquin:
@CodyCBakerPhD 
If you notice,  the CI is downloaded the ecephys data again instead of using the cache here. Shouldn't it be using the cache from the other tests that we just ran in the other PR?

CodyCBakerPhD:
> Shouldn't it be using the cache from the other tests that we just ran in the other PR?

Not until GitHub Caching Action allows either a control functionality, or better handling of a 'global' cache 😉 One of my only complaints about the functionality. See https://github.com/actions/cache/issues/692 and https://github.com/actions/cache/issues/342

h-mayorquin:
> > Shouldn't it be using the cache from the other tests that we just ran in the other PR?
> 
> Not until GitHub Caching Action allows either a control functionality, or better handling of a 'global' cache wink One of my only complaints about it. See [actions/cache#692](https://github.com/actions/cache/issues/692) and [actions/cache#342](https://github.com/actions/cache/issues/342) for background.

Interesting. So, the scope of the cache is within this branch and the parent one (which has not been run and therefore cache miss?).

CodyCBakerPhD:
> Interesting. So, the scope of the cache is within this branch and the parent one (which has not been run and therefore cache miss?).

Theoretically, but it also occasionally fails to find any parent. Also, none of this is specifiable and is at the mercy of the CI instance.

CodyCBakerPhD:
@h-mayorquin Great, thanks for looking into that!

CodyCBakerPhD (approved):

PR #534, opened 2022-05-30
Fix order of TwoPhotonSeries metadata update

weiglszonja:
## Motivation

This issue came up in catalystneuro/fee-lab-to-nwb#11, that adding a custom description for a `TwoPhotonSeries` gets overwritten by a default value. The order of metadata update should be reversed in this case, such that the default values should be updated with the "new" metadata values. 

## How to test the behavior?
```
```

## Checklist


h-mayorquin:
Check out #535. I think it solves the same issue.

CodyCBakerPhD:
Shifting this over to #535

PR #535, opened 2022-05-30
Update default parameters for ophys

h-mayorquin:
While working in the Seidemann conversion I have been prototyping a new ophys interface. I have noticed that there is some required metadata parameters (the device name in the imaging plane and the plane name in the TwoPhotonSeries) that our current defaults do not fill. This PR aligns the metadata behavior of `roieextractors` with the one `spikeinterface` (as libraries of `nwb-conversion-tools`) where the required metadata to run is pre-filled and updated if provided.

For some reason those were added manually in our `gin_test` so I removed them from there and that works as a test.

### DIFF
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -183,16 +183,19 @@ def get_nwb_imaging_metadata(imgextractor: ImagingExtractor):
     """
     metadata = default_ophys_metadata()
     # Optical Channel name:
-    for i in range(imgextractor.get_num_channels()):
-        ch_name = imgextractor.get_channel_names()[i]
-        if i == 0:
-            metadata["Ophys"]["ImagingPlane"][0]["optical_channel"][i]["name"] = ch_name
+    channel_name_list = imgextractor.get_channel_names()
+    if channel_name_list is None:
+        channel_name_list = ["generic_name"] * imgextractor.get_num_channels()
+
+    for index, channel_name in enumerate(channel_name_list):
+        if index == 0:
+            metadata["Ophys"]["ImagingPlane"][0]["optical_channel"][index]["name"] = channel_name
         else:
             metadata["Ophys"]["ImagingPlane"][0]["optical_channel"].append(
                 dict(
-                    name=ch_name,
+                    name=channel_name,
                     emission_lambda=np.nan,
-                    description=f"{ch_name} description",
+                    description=f"{channel_name} description",
                 )
             )
     # set imaging plane rate:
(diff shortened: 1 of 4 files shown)


h-mayorquin:
I think this should fix #534 as well.

CodyCBakerPhD:
@h-mayorquin Thanks for taking a look at this.

I'll enlist @weiglszonja help in reviewing this one - does this branch fix your issues Szonja?

h-mayorquin:
I feel that the caches do not work anymore ever.

CodyCBakerPhD:
> I feel that the caches do not work anymore ever.

According to https://github.com/catalystneuro/nwb-conversion-tools/runs/6662498429?check_suite_focus=true, both the ophys and behavior caches are still fine.

The biggest dataset, the ephys, also receives rather regular updates (and hence a new hash key associated with the cache which bumps the version). Also, it can help discovery to do a top-to-bottom reset of the unique IDs across the various workflows.

h-mayorquin:
> @h-mayorquin Thanks for taking a look at this.
> 
> I'll enlist @weiglszonja help in reviewing this one - does this branch fix your issues Szonja?

@weiglszonja 
You might need the branch in #535.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -205,6 +208,13 @@ def get_nwb_imaging_metadata(imgextractor: ImagingExtractor):
     metadata["Ophys"]["ImagingPlane"][0].update(imaging_rate=rate)
     # TwoPhotonSeries update:
     metadata["Ophys"]["TwoPhotonSeries"][0].update(dimension=imgextractor.get_image_size(), rate=rate)
+
+    device_name = metadata["Ophys"]["Device"][0]["name"]
+    metadata["Ophys"]["ImagingPlane"][0]["device"] = device_name
+
+    plane_name = metadata["Ophys"]["ImagingPlane"][0]["name"]
+    metadata["Ophys"]["TwoPhotonSeries"][0]["imaging_plane"] = plane_name
+
```
weiglszonja, line 217: Why not have this incorporated into `default_ophys_metadata()`? https://github.com/catalystneuro/nwb-conversion-tools/blob/dacddd4a485bce35cd68c97ef5355aaa84c1ff2c/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py#L39

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -183,16 +183,19 @@ def get_nwb_imaging_metadata(imgextractor: ImagingExtractor):
     """
     metadata = default_ophys_metadata()
```
weiglszonja, line 184: I would rename this function at this point to `get_default_ophys_metadata()`

weiglszonja (commented):
Thank you @h-mayorquin, this is great!

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
h-mayorquin: I tried this. This function is also used in the segmentation pipeline and it causes problems over there. @CodyCBakerPhD mentioned that there is an incoming refactor of `write_segmentation`, I think we will be able to do it then.

weiglszonja: ah okay, sorry then, I haven't checked write_segmentation thoroughly yet.

h-mayorquin: Is a good idea! let's revisit it when we are doing the refactor of `write_segmentation`.

CodyCBakerPhD (approved):

PR #536, opened 2022-05-31
Refactor of `add_two_photon_series`

h-mayorquin:
This follows the PR in #535 and I think will be necessary for getting the metadata correctly.

This is mostly a refactoring of the `add_two_photon_series` and as a bonus we can get rid of the function `safe_update` and use `dict_deep_update` in every instance of it.

### DIFF
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -26,16 +26,6 @@ from ..nwb_helpers import get_default_nwbfile_metadata, make_nwbfile_from_metada
 from ...utils import FilePathType, OptionalFilePathType, dict_deep_update
 
 
-# TODO: This function should be refactored, but for now seems necessary to avoid errors in tests
-def safe_update(d, u):
-    for k, v in u.items():
-        if isinstance(v, abc.Mapping):
-            d[k] = safe_update(d.get(k, {}), v)
-        else:
-            d[k] = v
-    return d
-
-
 def default_ophys_metadata():
     """Fill default metadata for optical physiology."""
     metadata = get_default_nwbfile_metadata()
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
@h-mayorquin Cool, thanks.

I've also assigned @weiglszonja to refactor the `write_segmentation`, so I'll enlist her help reviewing this PR (when it's ready) and enlist your help in reviewing her PR (when it's ready).

weiglszonja (dismissed):
This is great! it works with [#11](https://github.com/catalystneuro/fee-lab-to-nwb/pull/11)

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -95,52 +85,50 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
...
+        image_plane_metadata["device"] = device
+        imaging_plane = nwbfile.create_imaging_plane(**image_plane_metadata)
+        two_p_series_kwargs.update(imaging_plane=imaging_plane)
+
+        # Add timestamps or rate
+        timestamps = imaging.frame_to_time(np.arange(imaging.get_num_frames()))
+        if use_times:
+            two_p_series_kwargs.update(timestamps=H5DataIO(timestamps, compression="gzip"))
             if "rate" in two_p_series_kwargs:
                 del two_p_series_kwargs["rate"]
```
weiglszonja, line 126: @h-mayorquin 
How about we put a warning here that using timestamps and rate is not supported and that rate will be removed?

h-mayorquin:
@weiglszonja can we merge this before we keep with the discussion in #540?

weiglszonja:
Yeah, sure!

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -26,16 +27,6 @@
...
 
 
-# TODO: This function should be refactored, but for now seems necessary to avoid errors in tests
-def safe_update(d, u):
-    for k, v in u.items():
-        if isinstance(v, abc.Mapping):
-            d[k] = safe_update(d.get(k, {}), v)
-        else:
-            d[k] = v
-    return d
```
CodyCBakerPhD, line 36: @h-mayorquin Thank goodness, glad we don't need this any more.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -95,53 +86,52 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
...
         def data_generator(imaging):
             for i in range(imaging.get_num_frames()):
-                yield imaging.get_frames(frame_idxs=[i]).T
+                yield imaging.get_frames(frame_idxs=[i]).squeeze().T
 
         data = H5DataIO(
             DataChunkIterator(data_generator(imaging), buffer_size=buffer_size),
             compression=True,
         )
+        two_p_series_kwargs = two_photon_series_metadata
```
CodyCBakerPhD, line 108: Does this form a reference or a copy to the original metadata dictionary?

If the former, we should cop/ydeepcopy this so we can modify it without modifying the original structure.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -95,53 +86,52 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
...
+        two_p_series_kwargs.update(data=data)
 
-        # using internal data. this data will be stored inside the NWB file
-        two_p_series_kwargs = dict_deep_update(
-            metadata["Ophys"]["TwoPhotonSeries"][0],
-            dict(data=data, imaging_plane=imaging_plane),
-        )
+        # Add the image plane
+        image_plane_metadata = metadata["Ophys"]["ImagingPlane"][0]
```
CodyCBakerPhD, line 112: Similar question to above

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
h-mayorquin: Good point. This did not occur to me before.

My first thought was thinking that the metadata should end as closely as possible to the final form as it is written in the `nwb_file` but then it seems way more important than functions should not have side effects. I can see this producing a hard to debug error in the future and if we want the former thing we should sort this out somewhere else.

Should be sorted in the last commit.

h-mayorquin:
@CodyCBakerPhD @weiglszonja 
Should we merge this?

Issue #537, opened 2022-05-31
[Bug] metadata of segmentation extractor and imaging extractor interfaces are incompatible  


h-mayorquin:
The metadata of `BaseImagingExtractorInterface` requires the metadata of `TwoPhotonInterface` to be an array:
https://github.com/catalystneuro/nwb-conversion-tools/blob/a03bf773bb9808255eeb893ba2d2cbb5211a5758/src/nwb_conversion_tools/datainterfaces/ophys/baseimagingextractorinterface.py#L36-L45

Whereas the one from `BaseSegmentationExtractorInterface` requires it to be an object

https://github.com/catalystneuro/nwb-conversion-tools/blob/a03bf773bb9808255eeb893ba2d2cbb5211a5758/src/nwb_conversion_tools/datainterfaces/ophys/basesegmentationextractorinterface.py#L31-L36

If you run that line in main you get the following:
```python
from nwb_conversion_tools.utils import get_schema_from_hdmf_class
from pynwb.ophys import TwoPhotonSeries

get_schema_from_hdmf_class(TwoPhotonSeries)["type"]
```

outputs `object`

What should we do? What should be aligned with what?

Here is an example that reproduces the error with a data class (I think we can use something of this kind to test general metadata combination)

```python

import unittest
from datetime import datetime

from nwb_conversion_tools import Suite2pSegmentationInterface, TiffImagingInterface
from nwb_conversion_tools import NWBConverter

from .setup_paths import OPHYS_DATA_PATH, OUTPUT_PATH


TiffImagingInterface_source_data = dict(
    file_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "Tif" / "demoMovie.tif"), sampling_frequency=15.0
)
Suite2pSegmentationInterface_source_data = dict(file_path=str(OPHYS_DATA_PATH / "segmentation_datasets" / "suite2p"))

interface_to_source_data_map = dict(
    TiffImagingInterface=TiffImagingInterface_source_data,
    Suite2pSegmentationInterface=Suite2pSegmentationInterface_source_data,
)

interface_to_class_map = dict(
    TiffImagingInterface=TiffImagingInterface,
    Suite2pSegmentationInterface=Suite2pSegmentationInterface,
)

interfaces_to_combine = ["TiffImagingInterface", "Suite2pSegmentationInterface"]


class TestConverter(NWBConverter):
[...]

PR #538, opened 2022-05-31
Fix metadata combination of imaging and segmentation interfaces 

h-mayorquin:
This is a test for #537 that I think can serve as a general boilerplate for similar combinations.  Right now I have only written the test.

Edit: The issue is now fixed and the data is aligned.

### DIFF
--- a/src/nwb_conversion_tools/datainterfaces/ophys/basesegmentationextractorinterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ophys/basesegmentationextractorinterface.py
@@ -38,6 +38,8 @@ class BaseSegmentationExtractorInterface(BaseDataInterface, ABC):
 
         # Temporary fixes until centralized definition of metadata schemas
         metadata_schema["properties"]["Ophys"]["properties"]["ImagingPlane"].update(type="array")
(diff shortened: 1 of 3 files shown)


src/nwb_conversion_tools/datainterfaces/ophys/basesegmentationextractorinterface.py
```diff
@@ -38,13 +38,16 @@ def get_metadata_schema(self):
 
         # Temporary fixes until centralized definition of metadata schemas
         metadata_schema["properties"]["Ophys"]["properties"]["ImagingPlane"].update(type="array")
+        metadata_schema["properties"]["Ophys"]["properties"]["TwoPhotonSeries"].update(type="array")
```
CodyCBakerPhD, line 41: Yep, this is what I would have recommended from your original issue.

the `get_schema_from_hdmf_class` can provide a nice base filler of common items but so often we need to hand code little things like that in the `get_metadata_schema` methods.

Yet another reason in the end it might just be best to have all the metadata housed across hard-coded `.yml` files so it's all transparent, but also easy for the interface classes to read and use to set what they need.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin Thanks for catching this, and for keeping the changes minimal.

PR #540, opened 2022-06-01
Refactor of add_two_photon_series II

h-mayorquin:
Small PR that is a continuation of #536 

The biggest change is the removal of `use_times` and handles this logic to`roiextractors` in the same way that we did with `spikeinterface`. 

Also, the definition of `calculate_regular_series_rate` is changed. The one in main is not calculating the rate but the period of the signal.

### DIFF
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -143,7 +143,7 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
     """
 
     if use_times:
-        warn("use times is deprecate and will be removed on or after August 1st, 2022.")
+        warn("Keyword argument 'use_times' is deprecate dand will be removed on or after August 1st, 2022.")
 
     metadata_copy = deepcopy(metadata)
     metadata_copy = dict_deep_update(get_nwb_imaging_metadata(imaging), metadata_copy)
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -79,7 +79,7 @@ def get_default_ophys_metadata():
 
 def get_nwb_imaging_metadata(imgextractor: ImagingExtractor):
     """
-    Convert metadata from the segmentation into nwb specific metadata.
+    Convert metadata from the ImagingExtractor into nwb specific metadata.
 
     Parameters
     ----------
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -267,7 +267,7 @@ def write_imaging(
         assert isinstance(nwbfile, NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
 
     if use_times:
-        warn("use times is deprecate and will be removed on or after August 1st, 2022")
+        warn("Keyword argument 'use_times' is deprecated and will be removed on or after August 1st, 2022.")
 
     # TODO on or after August 1st, 2022, remove argument and deprecation warnings
     if save_path is not None:
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -143,7 +143,7 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
     """
 
     if use_times:
-        warn("Keyword argument 'use_times' is deprecate dand will be removed on or after August 1st, 2022.")
+        warn("Keyword argument 'use_times' is deprecated and will be removed on or after August 1st, 2022.")
 
     metadata_copy = deepcopy(metadata)
     metadata_copy = dict_deep_update(get_nwb_imaging_metadata(imaging), metadata_copy)
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -142,10 +142,8 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
     Adds two photon series from imaging object as TwoPhotonSeries to nwbfile object.
     """
 
-
     if use_times:
         warn("use times is deprecate and will be removed on or after August 1st, 2022.")
-
 
     metadata_copy = deepcopy(metadata)
     metadata_copy = dict_deep_update(get_nwb_imaging_metadata(imaging), metadata_copy)
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -103,11 +103,7 @@ def get_nwb_imaging_metadata(imgextractor: ImagingExtractor):
                 )
             )
     # set imaging plane rate:
-    rate = (
-        np.float("NaN")
-        if imgextractor.get_sampling_frequency() is None
-        else float(imgextractor.get_sampling_frequency())
-    )
+    rate = np.nan if imgextractor.get_sampling_frequency() is None else float(imgextractor.get_sampling_frequency())
 
     # adding imaging_rate:
     metadata["Ophys"]["ImagingPlane"][0].update(imaging_rate=rate)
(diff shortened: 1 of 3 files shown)


src/nwb_conversion_tools/utils/checks.py
```diff
@@ -10,5 +10,5 @@ def calculate_regular_series_rate(series: np.ndarray, tolerance_decimals: int =
     rate is a scalar otherwise it is None."""
     diff_ts = np.diff(series).round(decimals=tolerance_decimals)
     uniq_diff_ts = np.unique(diff_ts)
-    rate = diff_ts[0] if len(uniq_diff_ts) == 1 else None
+    rate = 1.0 / diff_ts[0] if len(uniq_diff_ts) == 1 else None
```
weiglszonja, line 13: Yeah you're right, thanks for spotting this!

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -80,7 +134,7 @@ def add_devices(nwbfile: NWBFile, metadata: dict):
     return nwbfile
 
 
-def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=False):
+def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10):
```
weiglszonja, line 137: Can we add type hints since it is being refactored?

h-mayorquin: Yes, good idea. We need to add unit-tests for this at some point as well and a doc-string. But right now I just wanted to get rid of the `use_times` value as `roieextractors` can can decide that on the spot according to what's available.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -71,6 +76,55 @@ def get_default_ophys_metadata():
...
+
+    device_name = metadata["Ophys"]["Device"][0]["name"]
+    metadata["Ophys"]["ImagingPlane"][0]["device"] = device_name
+
+    plane_name = metadata["Ophys"]["ImagingPlane"][0]["name"]
+    metadata["Ophys"]["TwoPhotonSeries"][0]["imaging_plane"] = plane_name
+
+    # remove what Segmentation extractor will input:
+    _ = metadata["Ophys"].pop("ImageSegmentation")
+    _ = metadata["Ophys"].pop("Fluorescence")
```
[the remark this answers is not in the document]
h-mayorquin: Yes, I think that's a good idea. Get some very basic and then add only what's necessary for the steps. I don't like those removals (`pop`) on the flight. 

That said, right now metadata is a bit muddled as you have been seeing (also check #538). I would like to wait till we are done with the conversions to do further changes in that direction.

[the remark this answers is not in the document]
CodyCBakerPhD: Towards the eventual future of just pulling all this stuff from a common centralized metadata `.yml` file, I wonder if we could make use of https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-10.2 for such fields, and simply specify it then within the schema itself.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Maybe keep it but at a deprecation warning that informs on run that the argument is no longer needed? (can use the same August deadline as the other deprecation)

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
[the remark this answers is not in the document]
h-mayorquin: https://python-jsonschema.readthedocs.io/en/stable/faq/#why-doesn-t-my-schema-s-default-property-set-the-default-on-my-instance

Here is more on that.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
h-mayorquin: Sure, why not?
It is back with a warning in the last commit. 
Feel free to add a more descriptive warning if you feel like : )

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -72,6 +77,55 @@ def get_default_ophys_metadata():
     return metadata
 
 
+def get_nwb_imaging_metadata(imgextractor: ImagingExtractor):
+    """
+    Convert metadata from the segmentation into nwb specific metadata.
```
CodyCBakerPhD, line 82:
```suggestion
    Convert metadata from the ImagingExtractor into nwb specific metadata.
```

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -88,6 +142,9 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
     Adds two photon series from imaging object as TwoPhotonSeries to nwbfile object.
     """
 
+    if use_times:
+        warn("use times is deprecate and will be removed on or after August 1st, 2022.")
```
CodyCBakerPhD, line 146:
```suggestion
        warn("Keyword argument 'use_times' is deprecate dand will be removed on or after August 1st, 2022.")
```

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -88,6 +142,9 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
     Adds two photon series from imaging object as TwoPhotonSeries to nwbfile object.
     """
 
+    if use_times:
+        warn("Keyword argument 'use_times' is deprecate dand will be removed on or after August 1st, 2022.")
```
CodyCBakerPhD, line 146:
```suggestion
        warn("Keyword argument 'use_times' is deprecated and will be removed on or after August 1st, 2022.")
```

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -254,14 +261,14 @@ def write_imaging(
...
-    use_times: bool (optional, defaults to False)
-        If True, the times are saved to the nwb file using imaging.frame_to_time(). If False (defualt),
-        the sampling rate is used.
     """
     assert save_path is None or nwbfile is None, "Either pass a save_path location, or nwbfile object, but not both!"
     if nwbfile is not None:
         assert isinstance(nwbfile, NWBFile), "'nwbfile' should be of type pynwb.NWBFile"
 
+    if use_times:
+        warn("use times is deprecate and will be removed on or after August 1st, 2022")
```
CodyCBakerPhD, line 270:
```suggestion
        warn("Keyword argument 'use_times' is deprecated and will be removed on or after August 1st, 2022.")
```

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -72,6 +77,55 @@ def get_default_ophys_metadata():
...
+                dict(
+                    name=channel_name,
+                    emission_lambda=np.nan,
+                    description=f"{channel_name} description",
+                )
+            )
+    # set imaging plane rate:
+    rate = (
+        np.float("NaN")
+        if imgextractor.get_sampling_frequency() is None
```
CodyCBakerPhD, line 108: Can this actually ever be `None` currently? If so that's a problem, it's more expected (as per our previous conversations) that this be a required value for all extractors.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -72,6 +77,55 @@ def get_default_ophys_metadata():
...
+            metadata["Ophys"]["ImagingPlane"][0]["optical_channel"].append(
+                dict(
+                    name=channel_name,
+                    emission_lambda=np.nan,
+                    description=f"{channel_name} description",
+                )
+            )
+    # set imaging plane rate:
+    rate = (
+        np.float("NaN")
```
CodyCBakerPhD, line 107: I show `np.float("NaN")` as being deprecated soon - can you use `np.nan` instead?

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
h-mayorquin: If so, then we can get rid of this.

h-mayorquin: @CodyCBakerPhD OK, in fact, the class `Hdf5ImagingExtractor(ImagingExtractor)` does not require `sampling_frequency` in its required values and we have it like that in our tests.

CodyCBakerPhD: Dang, we should probably either fix that, or it's 'optional' because it might be an attribute associated with the `Dataset` object (but required if the `Dataset` doesn't have one).

CodyCBakerPhD (approved):

PR #541, opened 2022-06-04
Added check to avoid adding roi response series with only 0 values

h-mayorquin:
A small fix for `write_segmentation` to avoid writing response series consisting of only zero values.

### DIFF
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -524,9 +524,8 @@ def write_segmentation(
         roi_response_dict = {key: value for key, value in roi_response_dict.items() if value is not None}
         for signal, response_series in roi_response_dict.items():
 
-            # all_data_is_zero = [value == 0 for value in list(response_series.flatten())]
-            all_data_is_zero = np.isclose(response_series, 0).all()
-            if not all_data_is_zero:
+            not_all_data_is_zero = any(x != 0 for x in np.ravel(response_series))
+            if not_all_data_is_zero:
                 data = np.asarray(response_series)
                 trace_name = "RoiResponseSeries" if signal == "raw" else signal.capitalize()
                 trace_name = trace_name if plane_no_loop == 0 else trace_name + f"_Plane{plane_no_loop}"
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -507,23 +507,32 @@ def write_segmentation(
             )
 
             ps = image_segmentation.create_plane_segmentation(**input_kwargs)
-        # Fluorescence Traces:
+
+        # Fluorescence Traces - This should be a function on its own.
         if "Flourescence" not in ophys.data_interfaces:
             fluorescence = Fluorescence()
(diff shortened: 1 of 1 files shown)


src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -503,23 +503,32 @@ def image_mask_iterator():
...
+
         rate = np.nan if segext_obj.get_sampling_frequency() is None else segext_obj.get_sampling_frequency()
-        for i, j in roi_response_dict.items():
-            data = getattr(segext_obj, f"_roi_response_{i}")
-            if data is not None:
-                data = np.asarray(data)
-                trace_name = "RoiResponseSeries" if i == "raw" else i.capitalize()
+
+        # Filter empty data
+        roi_response_dict = {key: value for key, value in roi_response_dict.items() if value is not None}
```
CodyCBakerPhD, line 524: A bit fancier/more elegant approach to filtering iterables in Python:

```python
roi_response_dict = dict(filter(lambda elem: elem[1] is not None, roi_response_dict .items()))
```

`elem` is an `item` in the `dict`, and `elem[1]` is the value (`elem[0]` is the key)

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -503,23 +503,32 @@ def image_mask_iterator():
...
+        for signal, response_series in roi_response_dict.items():
+
+            # all_data_is_zero = [value == 0 for value in list(response_series.flatten())]
+            all_data_is_zero = np.isclose(response_series, 0).all()
```
CodyCBakerPhD, line 528: I would rather this logical test be completely lazy in terms of memory - if the series is not being buffered, and it if it quite long, then the call to `np.isclose()` generates a full boolean array of the same shape. The array generated has, granted, `itemsize=1` for all elements, but that could still create instantaneous RAM usage of 50% of the original array amount if it uses `int16`.

The most efficient way I can think of would be

```python
any((x != 0 for x in np.ravel(response_series)))
```

`np.ravel` returns a 1D lazy view of the original array (does not create a copy of the shape) for easy iteration over all elements, the logical application is done via a generator so it technically only yields one logical element from the equality operation performed on a single element from the view at a time, and `any` even has a early return functionality where it stops iteration the first time it receives a `True` value from the iterable.

h-mayorquin: This is clever, thanks.

h-mayorquin: Done in the last commit.

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
h-mayorquin: Not sure it looks more elegant to me for this case. I think filter makes sense to me if the filter is more complex.

CodyCBakerPhD: Just wanted you to know about it - another plus is that `filter` is a lazy generator, so it doesn't actually 'apply' the filter until you call an executable on it. This doesn't make a difference here but it can in the future, and can be nice if the thing you are filtering is large.

CodyCBakerPhD (approved):

Issue #326, opened 2021-12-14
semi-specified buffer size


bendichter:
For a `GenericDataChunkIterator`, I'd like to be able to set a buffer shape to:

buffer_shape=(1, None)
buffer_gb=10

This would be useful when you know you only want to read one channel at a time, but don't know how many time points you want to read.

CodyCBakerPhD:
On it - this might take a minute though, it always made the assumption that you specify only one or the other, not a partial of `buffer_shape`

bendichter:
Yeah. I would have used this with the DJ project and I feel like this might be useful in other scenarios as well, but I think it's non-critical.

PR #542, opened 2022-06-06
[Refactoring/Enhancement] Pad buffering in GenericDataChunkIterator

CodyCBakerPhD:
It has been known since the original implementation of the `GenericDataChunkIterator` that the default buffer shape method (https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/data_utils.py#L310-L329) given a target buffer size were a tad inefficient with extremely lopsided distributions of axes lengths (*i.e.*, if one axis is many orders of magnitude longer than any others). The main justification of being OK with this was because we allowed manual specification of the `buffer_shape` which gives the user full control over calculating such things.

So far this has not been too severe; statistically speaking, with respect to the set of permutations of 2D arrays ranging from `1,000-10,000 x 1,000-100,000`, the default method is able to specify a buffer shape within at most a 5% error from the target `buffer_gb` size (so only 50 MB off). However, when comparing 2D arrays with one axis in the range of `~1,000` and the other around `~100,000,000`, the resulting size of the buffer shapes tend to vastly underestimate the target.

With respect to our most common use case of this iterator (`ecephys` traces) this essentially means that single probe SpikeGLX recordings of more than 30 minutes in duration will start to experience decreased speeds with iterative write (unless, of course, a hard-coded buffer shape was originally specified).

Anyway, this PR introduces some heuristics for buffer 'padding' that produce equivalent performance in the previous axis scaling regimes but fixes the issue for very disproportionate ratios by filling the buffer with as many contiguous chunks along an axis as possible.

A follow-up to this improvement to allow partial manual size specification of shape parameters (requested by #326) will also help in these kinds of cases by allowing the user to specify the semantic equivalent of 'load as much of these axes that I've set to `None` as possible'.


src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,58 @@
...
+
+        # Early termination condition
+        if np.prod(maxshape) * self.dtype.itemsize / 1e9 < buffer_gb:
+            return tuple(self.maxshape)
+
+        chunks_per_axis = np.ceil(maxshape / self.chunk_shape)
+        buffer_shape = np.array(self.chunk_shape)
+        buffer_bytes = chunk_bytes
+        axis_sizes_bytes = maxshape * self.dtype.itemsize
+        chunk_size_order = np.argsort(self.chunk_shape)
```
h-mayorquin, line 28:
```
        chunk_size_order = np.argsort(self.chunk_shape)
```
I would like to avoid the double indexing below and use something like:

```

index_of_smallest, index_of_second_smallest, *_ = np.argsort(x)

Comment (h-mayorquin) [inline src/nwb_conversion_tools/tools/hdmf.py:28]:
```
        chunk_size_order = np.argsort(self.chunk_shape)
```
I would like to avoid the double indexing below and use something like:

```

index_of_smallest, index_of_second_smallest, *_ = np.argsort(x)

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,58 @@
...
+        full_axes_used = np.zeros(shape=len(self.maxshape), dtype=bool)
+        if small_axis_fill_size <= target_buffer_bytes:
+            buffer_bytes = small_axis_fill_size
+            buffer_shape[chunk_size_order[0]] = self.maxshape[chunk_size_order[0]]
+            full_axes_used[chunk_size_order[0]] = True
+        for axis, chunks_on_axis in enumerate(chunks_per_axis):
+            if full_axes_used[axis]:  # If the smallest axis, skip since already used
+                continue
+            if chunks_on_axis * buffer_bytes <= target_buffer_bytes:  # If multiple axes can be used together
+                buffer_bytes *= chunks_on_axis
```
h-mayorquin, line 49: Maybe we can use a better variable name that indicates that this variable `buffer_bytes` is being filled here. `buffer_bytes_so_far` or something like that.

h-mayorquin, line 49: Maybe we can use a better variable name that indicates that this variable `buffer_bytes` is being filled here. `buffer_bytes_so_far` or something like that.

tests/test_internals/test_hdmf.py
```diff
@@ -0,0 +1,36 @@
...
+
+def test_early_exit():
+    """Uses a 32 byte array with 1 GB buffer size (default) and 1 MB chunk size (default)."""
+    iterator = SliceableDataChunkIterator(data=np.random.random(size=(2, 2)))
+    assert iterator.maxshape == iterator.buffer_shape
+
+
+def test_buffer_padding_long_shape():
+    """Uses ~8 MB array with 11 MB buffer size and 1 MB chunk size (default)."""
+    iterator = SliceableDataChunkIterator(data=np.zeros(shape=(160000, 20)), buffer_gb=1.1e-2)
```
h-mayorquin, line 23: Maybe we should use `np.empty` instead of `np.zeros`. Not sure what will happen under the hood (of the DataChunkIterator` but if it works well it should be faster.

h-mayorquin, line 23: Maybe we should use `np.empty` instead of `np.zeros`. Not sure what will happen under the hood (of the DataChunkIterator` but if it works well it should be faster.

h-mayorquin (commented):
I finally took a first look at this. 

Some comments here and there.

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,58 @@
 from typing import Tuple
 
 import numpy as np
-from hdmf.data_utils import GenericDataChunkIterator
+from hdmf.data_utils import GenericDataChunkIterator as HDMFGenericDataChunkIterator
+
+
+class GenericDataChunkIterator(HDMFGenericDataChunkIterator):
+    def _get_default_buffer_shape(self, buffer_gb: float = 1.0) -> Tuple[int]:
+        chunk_bytes = np.prod(self.chunk_shape) * self.dtype.itemsize
```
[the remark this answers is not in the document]
CodyCBakerPhD: For now, we are locked into using `buffer_gb` and `chunk_mb` since that is what made it into the `hdmf` release. We really prefer to not make backwards incompatible changes to that package unless they are critical since it's the most upstream/general package.

Eventually, it would be nice to fully implement a rigorous way of parsing strings similar, if not identical, to SpikeInterface: https://github.com/SpikeInterface/spikeinterface/blob/ce2af6098ea4e9caf2410bf5f3d8a5708bc56841/spikeinterface/core/job_tools.py#L21-L22

and that additional feature would be a good excuse to deprecate the name.


> I just encountered this code for the first time and my first thought was that it was a pointer to the bytes.

Pointer to the bytes? How do you mean?

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,58 @@
...
+            np.array(self.chunk_shape) > 0
+        ), f"Some dimensions of chunk_shape ({self.chunk_shape}) are less than zero!"
+
+        maxshape = np.array(self.maxshape)
+
+        # Early termination condition
+        if np.prod(maxshape) * self.dtype.itemsize / 1e9 < buffer_gb:
+            return tuple(self.maxshape)
+
+        chunks_per_axis = np.ceil(maxshape / self.chunk_shape)
```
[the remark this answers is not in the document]
CodyCBakerPhD: Sure

tests/test_internals/test_hdmf.py
CodyCBakerPhD: Sure

CodyCBakerPhD:
@h-mayorquin Any idea why all your review comments got duplicated 2-4 times?

src/nwb_conversion_tools/tools/hdmf.py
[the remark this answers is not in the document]
h-mayorquin: Yes, keeping it compatible with `hdfm5` seems like a more important concern than code readability. I do think though that we should just re-defined the variables at the beginning for clarity so we used those interanlly for readability. 


You can have a variable that is an actual reference to the [bytes](bytearray).

h-mayorquin:
> @h-mayorquin Any idea why all your review comments got duplicated 2-4 times?

Not really sure. Maybe it is because I have more than one tab opened when I submit the review.

src/nwb_conversion_tools/tools/hdmf.py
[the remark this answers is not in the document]
CodyCBakerPhD: > You can have a variable that is an actual reference to the [bytes](https://github.com/catalystneuro/nwb-conversion-tools/pull/bytearray).

Link is broken - what did you mean to reference?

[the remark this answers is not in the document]
CodyCBakerPhD: > I do think though that we should just re-defined the variables at the beginning for readability in the rest of the function.

This function isn't intended to live here for any longer than it takes to ensure that it functions well and without error, at which time it will be propagated up to `hdmf`.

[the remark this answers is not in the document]
h-mayorquin: https://docs.python.org/3/library/functions.html#func-bytes

All right.

CodyCBakerPhD:
@h-mayorquin Check out latest changes reflecting your suggestions.

I also added the original one-shot calculation (unpadded), which can produce better results for more squarish shapes. The method will now use whichever shape produces a closer fill to the buffer size, while still not taking very many iterations to perform said estimation.

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,77 @@
...
+        sub_square_buffer_shape = np.array(self.chunk_shape)
+        if min(axis_sizes_bytes) > target_buffer_bytes:
+            k1 = np.floor((target_buffer_bytes / chunk_bytes) ** 0.5)
+            for axis in [smallest_chunk_axis, second_smallest_chunk_axis]:
+                sub_square_buffer_shape[axis] = k1 * sub_square_buffer_shape[axis]
+            return tuple(sub_square_buffer_shape)
+
+        # Original one-shot estimation has good performance for certain shapes
+        k2 = np.floor(
+            (buffer_gb * 1e9 / (np.prod(self.chunk_shape) * self.dtype.itemsize)) ** (1 / len(self.chunk_shape))
```
h-mayorquin, line 40:
```suggestion
            (buffer_gb * 1e9 / chunk_bytes) ** (1 / len(self.chunk_shape))
```

I would write:
```python
chunks_in_buffer = np.floor(buffer_gb * 1e9 / chunk_bytes)
chunks_per_axis  = chunks_in_buffer ** (1 / self.chunk_shape.ndim)
```
Instead of just k. But I have seen that they have this k values in the `hdfm` code. Up to you.

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,77 @@
...
+        )
+        unpadded_buffer_shape = tuple(
+            [
+                min(max(int(x), self.chunk_shape[j]), self.maxshape[j])
+                for j, x in enumerate(k2 * np.array(self.chunk_shape))
+            ]
+        )
+        unpadded_buffer_bytes = np.prod(unpadded_buffer_shape) * self.dtype.itemsize
+
+        # Otherwise, try to start by filling the smallest axis completely or calculate best partial fill
```
h-mayorquin, line 50: Otherwise? should there be a condition here or is this referring to the case when the minimal smallest full axis does not fit within the buffer size? If the latter, maybe this should be before the definition / construction of `unpadeed_buffer_bytes` above.

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,77 @@
...
+        k2 = np.floor(
+            (buffer_gb * 1e9 / (np.prod(self.chunk_shape) * self.dtype.itemsize)) ** (1 / len(self.chunk_shape))
+        )
+        unpadded_buffer_shape = tuple(
+            [
+                min(max(int(x), self.chunk_shape[j]), self.maxshape[j])
+                for j, x in enumerate(k2 * np.array(self.chunk_shape))
+            ]
+        )
+        unpadded_buffer_bytes = np.prod(unpadded_buffer_shape) * self.dtype.itemsize
```
h-mayorquin, line 48: I think you don't need those parentheses there. I think you can do tuple comprehension directly without them.

Also, up to you, but I would do two list comprehension here for clarity / readability. One that calculates the max between equi-distributed shape and another that clips over max shape. But I think you like brevity more than I do.  Along those lines, you could also use something like `np.clip` to sort this out as well in one sweep I think (to clip equi-distributed shapes between max_shapes and chunk_shapes.

h-mayorquin (commented):
Looks great, thanks for taking into account my suggestions.

One think that I would like to have for the testing is for them to provide a construction or the expected shape. Right now they seem hard-coded and I don't know why I should expect those values and therefore it would be less obvious what broke if they fail. I was trying to think about how to do that though and it did not come easy. Maybe it is not feasible or easy in this case without basically reproducing the code of the function in the test? 

What do you think about this?

src/nwb_conversion_tools/tools/hdmf.py
CodyCBakerPhD: You mean separate compressions over the min vs. the max?

CodyCBakerPhD: `np.clip` is nice - thanks for pointing that out!

src/nwb_conversion_tools/tools/hdmf.py
CodyCBakerPhD: adjusted naming and added steps

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -1,7 +1,76 @@
...
+        if min(axis_sizes_bytes) > target_buffer_bytes:
+            k1 = np.floor((target_buffer_bytes / chunk_bytes) ** 0.5)
+            for axis in [smallest_chunk_axis, second_smallest_chunk_axis]:
+                sub_square_buffer_shape[axis] = k1 * sub_square_buffer_shape[axis]
+            return tuple(sub_square_buffer_shape)
+
+        # Original one-shot estimation has good performance for certain shapes
+        chunk_to_buffer_ratio = buffer_gb * 1e9 / chunk_bytes
+        chunk_scaling_factor = np.floor(chunk_to_buffer_ratio ** (1 / num_axes))
+        unpadded_buffer_shape = [
```
h-mayorquin, line 42: I meant that you could have vectorized the whole thing:
```python
unpadded_buffer_shape = np.clip(a=chunk_scaling_factor * np.array(self.chunk_shape), a_min=self.chunk_shape, a_max=self.maxshape).astype("int")
```

h-mayorquin (approved):

h-mayorquin:
> > and I don't know why I should expect those values
> 
> Based on the intercombinations of conditions that lead into those different blocks of the method,
> 
> `test_buffer_padding_long_shape` with shape `(10**7, 20)` -> results in shape `(68482, 20)` by using the entire second axis and partial fill on first (both steps of new method)
> 
> `test_buffer_padding_mixed_shape` with shape `(20, 40, 2401)` -> results in shape `(16, 32, 1920)` by using the original method of 'similar rectangle of smaller size'. This is also probably currently tested on the `hdmf` side so this test probably won't propagate.
> 
> `test_min_axis_too_large` with shape `(1000, 100)` but artificially low chunk to buffer size and ratio -> results in shape `(20, 20)` by using the new condition where no consecutive axis slice fits in the buffer (original method would have collapsed the buffer shape to a chunk shape)

Thanks for describing them. Maybe it would be a good idea to add them in the test docstrings? As this is moving to `hdfm` I don't know what standards they have over there.

src/nwb_conversion_tools/tools/hdmf.py
CodyCBakerPhD: Hm... I'll have to try that - maybe in a followup PR just so we can get this through

PR #544, opened 2022-06-09
[Bug] source data validation fails when using new numpy typing

h-mayorquin:
This is a PR to showcase a bug. I will opened issue #545 concurrently to describe this in more detail.


h-mayorquin:
> @h-mayorquin Thanks for pointing this out - will be a while until a solution for iterating over deep annotation types is in place.
> 
> Do you mind if I close this and add a link to the workflow run onto the issue?

That's OK. I was wondering if that could be possible. No point for this being here.

PR #546, opened 2022-06-11
Unit test `add_two_photon_series` in the `roiextractors.py` module  

h-mayorquin:
During the current conversion (Seidemann) I am getting duplicated device metadata where specifying the actual metadata for ophys does not overwrite the defaults. This is a follow up of #543 where I corrected the same for `add_device` and this is the analgous for `add_two_photon_series`:

Add unit tests for  `add_two_photon_series` in roieextractors
Modify the function in such a way that the metadata for ophys / device is passed the defaults are not included.

EDIT: This is doing the same change as #547 as well.

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -14,9 +14,9 @@ psutil==5.8.0
 lxml==4.6.5
 opencv-python==4.5.1.48
 spikeextractors==0.9.9
-spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@12b7863684b705e3d0ae8165bb3009143c70530c
+spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@261c5f94359ce5631f2d8f19bac923ceeffb63c7
 neo==0.10.2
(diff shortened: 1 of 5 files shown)


requirements-full.txt
```diff
@@ -14,9 +14,9 @@ psutil==5.8.0
 lxml==4.6.5
 opencv-python==4.5.1.48
 spikeextractors==0.9.9
-spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@12b7863684b705e3d0ae8165bb3009143c70530c
+spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@261c5f94359ce5631f2d8f19bac923ceeffb63c7
```
h-mayorquin, line 17:
```suggestion
spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@12b7863684b705e3d0ae8165bb3009143c70530c
spikeinterface @ 
```

CodyCBakerPhD (approved):

PR #547, opened 2022-06-11
Correct order of writing in `add_two_photon_series`

h-mayorquin:
As discussed with @CodyCBakerPhD it seems that the `write_imaging` pipeline and in particular the `add_two_photon_series` function is writing our images inverted. 

This PR is connected to the proper correction in`roiextractors` and the requirements consequently pinned to them:
https://github.com/catalystneuro/roiextractors/pull/136

### DIFF
--- a/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
+++ b/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
@@ -185,7 +185,7 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
     # Add the data
     def data_generator(imaging):
         for i in range(imaging.get_num_frames()):
-            yield imaging.get_frames(frame_idxs=[i]).squeeze()
+            yield imaging.get_frames(frame_idxs=[i]).squeeze().T
 
     data = H5DataIO(
         DataChunkIterator(data_generator(imaging), buffer_size=buffer_size),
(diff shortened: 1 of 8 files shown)


h-mayorquin:
The changes introduce on this were already merged as of #546. We should use this PR to pin the `requirements` to a specific commit once we merge  https://github.com/catalystneuro/roiextractors/pull/136 in `roieextractors` .

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -185,7 +185,7 @@ def add_two_photon_series(imaging, nwbfile, metadata, buffer_size=10, use_times=
     # Add the data
     def data_generator(imaging):
         for i in range(imaging.get_num_frames()):
-            yield imaging.get_frames(frame_idxs=[i]).squeeze().T
+            yield imaging.get_frames(frame_idxs=[i]).squeeze()
```
h-mayorquin, line 189: @CodyCBakerPhD Thanks!

CodyCBakerPhD:
Leaving this open until we get a chance to talk about it with Ben...

Be prepared for the possible answer 'NWB ophys (x,y) values actually mean (cols,rows)' lol. We really should just use better language on the NWB schema, so thanks for getting that conversation started.

h-mayorquin:
> Leaving this open until we get a chance to talk about it with Ben...
> 
> Be prepared for the possible answer 'NWB ophys (x,y) values actually mean (cols,rows)' lol. We really should just use better language on the NWB schema, so thanks for getting that conversation started.

https://github.com/NeurodataWithoutBorders/nwb-schema/issues/522

CodyCBakerPhD:
> https://github.com/NeurodataWithoutBorders/nwb-schema/issues/522

Yep, that's what I meant. Thanks for linking.

TBH my own opinion is that on the schema side, it might make the most sense to a new user to refer to their underlying image as 'width' x 'height', since those are quantities they can think about without referencing the data storage (to them it may only appear as an picture on their screen, they may never even interact with it programmatically).

CodyCBakerPhD:
Turns out NWB expects width x height, which is the roiextractors equivalent of cols x rows, hence the transposition here.

PR #548, opened 2022-06-11
Added stubbing for write imaging

### LINKED ISSUE #530: [Feature]: propagate roiextractor stubbing

Recently requested and approved at the roiextractors level: https://github.com/catalystneuro/roiextractors/issues/103

Would need to be propagated here.

_No response_

Want to be able to stub imaging/segmentation objects for testing a conversion.

Yes.

### Code of Conduct

h-mayorquin:
Should close #530.

This propagates the work that @CodyCBakerPhD did in `roieextractors`. I already used it successfully in the current conversion.

### DIFF
--- a/src/nwb_conversion_tools/datainterfaces/ophys/baseimagingextractorinterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ophys/baseimagingextractorinterface.py
@@ -77,10 +77,18 @@ class BaseImagingExtractorInterface(BaseDataInterface):
         nwbfile: Optional[NWBFile] = None,
         metadata: Optional[dict] = None,
         overwrite: bool = False,
+        stub_test: bool = False,
         save_path: OptionalFilePathType = None,
     ):
+
+        if stub_test:
+            stub_frames = min([100, self.imaging_extractor.get_num_frames()])
+            imaging_extractor = self.imaging_extractor.frame_slice(start_frame=0, end_frame=stub_frames)
+        else:
+            imaging_extractor = self.imaging_extractor
+
         write_imaging(
-            imaging=self.imaging_extractor,
+            imaging=imaging_extractor,
             nwbfile_path=nwbfile_path,
             nwbfile=nwbfile,
             metadata=metadata,
--- a/tests/test_internals/test_roi_extractors_module.py
+++ b/tests/test_internals/test_roi_extractors_module.py
@@ -199,7 +199,7 @@ class TestAddTwoPhotonSeries(unittest.TestCase):
         data_in_hdfm_data_io = acquisition_modules[self.two_photon_series_name].data
         data_chunk_iterator = data_in_hdfm_data_io.data
         two_photon_series_extracted = np.concatenate([data_chunk.data for data_chunk in data_chunk_iterator])
-        assert two_photon_series_extracted.shape == (self.num_frames, self.rows, self.columns)
+        assert two_photon_series_extracted.shape == (self.num_frames, self.columns, self.rows)  # to be fixed/clarified soon
 
         # Check device
         devices = self.nwbfile.devices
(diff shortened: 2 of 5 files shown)


src/nwb_conversion_tools/datainterfaces/ophys/baseimagingextractorinterface.py
```diff
@@ -77,10 +77,18 @@ def run_conversion(
         nwbfile: Optional[NWBFile] = None,
         metadata: Optional[dict] = None,
         overwrite: bool = False,
+        stub_test: bool = False,
         save_path: OptionalFilePathType = None,
     ):
+
+        if stub_test:
+            stub_frames = min([100, self.imaging_extractor.get_num_frames()])
```
CodyCBakerPhD, line 85: I was thinking we should set this to

```python
stub_frames = min([max(self.imaging_extractor.get_image_size()) + 1, self.imaging_extractor.get_num_frames()])
```

to avoid triggering the `check_data_orientation` from the NWBInspector.

tests/test_internals/test_roi_extractors_module.py
```diff
@@ -199,7 +199,7 @@ def test_add_two_photon_series(self):
         data_in_hdfm_data_io = acquisition_modules[self.two_photon_series_name].data
         data_chunk_iterator = data_in_hdfm_data_io.data
         two_photon_series_extracted = np.concatenate([data_chunk.data for data_chunk in data_chunk_iterator])
-        assert two_photon_series_extracted.shape == (self.num_frames, self.rows, self.columns)
+        assert two_photon_series_extracted.shape == (self.num_frames, self.columns, self.rows)  # to be fixed/clarified soon
```
h-mayorquin: Thanks, I was coming here to this.

src/nwb_conversion_tools/datainterfaces/ophys/baseimagingextractorinterface.py
h-mayorquin: I feel that we should not make these type of decisions base on the failure in a edge case of the inspector. 

I actually suggest two things to avoid the problem with our own stubbed files:
1) If available, used the property `dimension` in the [TwoPhotonSeries](https://pynwb.readthedocs.io/en/stable/pynwb.ophys.html#pynwb.ophys.TwoPhotonSeries) to figure out which axes are the ones corresponding to the images (we can actually set this by default in our conversion pipeline using `imaging_extractor.get_image_size()`)
2) When the property `dimension` is not available, `nwbinspector` could check for two equal size (or close enough) sizes for the images (usually images are equal rows and columns).

These solution has the advantage that will generalize beyond our use case here.

CodyCBakerPhD: > I feel that we should not make these type of decisions base on the failure in a edge case of the inspector.

It's not an edge case, the NWBInspector correctly raises in the situation, it's just that the real data doesn't match this situation.

It would be nice if the NWBInspector reports accurately reflected what we would see if run on the full file as opposed to a stub, since the purpose of a stub is to have a 'smaller' version of the file that looks as close as possible to the full one.

We can of course always mentally ignore it in the report, but still, I don't see what the big deal is about making the report as reflective as possible.

> 1. If available, used the property dimension in the [TwoPhotonSeries](https://pynwb.readthedocs.io/en/stable/pynwb.ophys.html#pynwb.ophys.TwoPhotonSeries) to figure out which axes are the ones corresponding to the images (we can actually set this by default in our conversion pipeline using imaging_extractor.get_image_size())
> 2. When the property dimension is not available, nwbinspector could check for two equal size (or close enough) sizes for the images (usually images are equal rows and columns).

This is an extremely generic check on all time series and it works simply by asserting that the longest axis is the zero axis (time). For small images (no axis above 100 in pixels) this is already guaranteed by the hardcoded value, but would then become triggered for large images simply because the value is hardcoded.

Other checks do more sophisticated things for particular objects, similar to what you are suggesting. But those are other checks, not this check.

CodyCBakerPhD: Just an idea - we can always add in later (together with other interfaces, as well)

CodyCBakerPhD (approved):

PR #550, opened 2022-06-13
YAML conversion specification CLI

CodyCBakerPhD:
WIP


CodyCBakerPhD:
Updated `--help`:
image

src/nwb_conversion_tools/tools/yaml_conversion_specification/yaml_conversion_specification.py
```diff
@@ -4,14 +4,50 @@
...
 
 from ...nwbconverter import NWBConverter
 from ...utils import dict_deep_update, load_dict_from_file, FilePathType, OptionalFolderPathType
 
 
+@click.command()
+@click.argument("specification-file-path")
+@click.option(
+    "--data-folder",
+    help="Modules to import prior to reading the file(s).",
```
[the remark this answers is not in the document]
h-mayorquin: Should it not be "folder_path` and `output_path` to keep it consistent with our use in other places?

src/nwb_conversion_tools/tools/yaml_conversion_specification/yaml_conversion_specification.py
```diff
@@ -4,14 +4,50 @@
...
+@click.option(
+    "--output-folder",
+    default=None,
+    help="Folder to save the output NWBFile to.",
+    type=click.Path(writable=True),
+)
+@click.option("--overwrite", help="Overwrite an existing NWBFile at the location.", is_flag=True)
+def run_conversion_from_yaml_cli(
+    specification_file_path: str,
+    data_folder: Optional[str] = None,
```
h-mayorquin, line 33: `OptionalFolderPathType`?

h-mayorquin (commented):
Looks good to me. I have never seen click. I guess it makes it easier to write scripts as you only add a decorator instead of writing a dedicated file. 

Is there any other reason to use `click` instead of using [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) the recommended command-line parsing module in the Python standard library? Maybe we want to avoid introducing another dependency (not sure about this).

src/nwb_conversion_tools/tools/yaml_conversion_specification/yaml_conversion_specification.py
CodyCBakerPhD: Nope, annotation types for CLI functions must be strictly strings, `None`, or occasionally `bool` if a 'flag'. It has its own system for specifying that an argument is a file/folder type of path, but does not interact with the `pathlib` library in that way.

CodyCBakerPhD:
> I have never seen click. I guess it makes it easier to write scripts as you only add a decorator instead of writing a dedicated file.

> Is there any other reason to use click instead of using [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) the recommended command-line parsing module in the Python standard library? Maybe we want to avoid introducing another dependency (not sure about this as I don't see an update in requirements.txt here, maybe something else is installing it).

`click` is the most popular/efficient CLI and is used in many other cases across our ecosystem, such as the DANDI API. Here's some more reading: https://click.palletsprojects.com/en/8.1.x/why/#why-not-argparse

One thing I particularly love about it is the semantic similarity 'did you mean...' as well as good `help` functionality for ill argument order/construction/type.

src/nwb_conversion_tools/tools/yaml_conversion_specification/yaml_conversion_specification.py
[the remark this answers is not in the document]
CodyCBakerPhD: Hm.. probably an oversight of https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/tools/yaml_conversion_specification/yaml_conversion_specification.py#L16-L19

These mimic the library usage of the function, but those should also probably have `_path` on the end of the name to be similar to the DataInterfaces, that is true.

I'll change that here and add a deprecation warning.

h-mayorquin:
> > I have never seen click. I guess it makes it easier to write scripts as you only add a decorator instead of writing a dedicated file.
> 
> > Is there any other reason to use click instead of using [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) the recommended command-line parsing module in the Python standard library? Maybe we want to avoid introducing another dependency (not sure about this as I don't see an update in requirements.txt here, maybe something else is installing it).
> 
> `click` is the most popular/efficient CLI and is used in many other cases across our ecosystem, such as the DANDI API. Here's some more reading: https://click.palletsprojects.com/en/8.1.x/why/#why-not-argparse
> 
> One thing I particularly love about it is the semantic similarity 'did you mean...' as well as good `help` functionality for ill argument order/construction/type.

So this comes with already in the dependencies of dandi then?
I think that we should make a explicit dependency to it.

CodyCBakerPhD:
> So this comes with already in the dependencies of dandi then?
I think that we should make a explicit dependency to it.

According to https://github.com/dandi/dandi-cli/blob/master/setup.cfg#L33, yes.

Since `dandi` is in the minimal requirements, and `click` comes with `dandi`, IDK that we need an explicit specification here. They aren't switching from it any time soon (if ever) and it's not possible to install NCT without `dandi`.

h-mayorquin:
> > So this comes with already in the dependencies of dandi then?
> > I think that we should make a explicit dependency to it.
> 
> According to https://github.com/dandi/dandi-cli/blob/master/setup.cfg#L33, yes.
> 
> Since `dandi` is in the minimal requirements, and `click` comes with `dandi`, IDK that we need an explicit specification here. They aren't switching from it any time soon (if ever) and it's not possible to install NCT without `dandi`.

I am thinking on the long term project of having the dependencies specified so users can select which parts of the library they want to install (as in, if the only want to do conversions with specific formats they can install just the libraries necessary for that pipeline as in neo). From that perspective, click is used in two separate functions, 1) the `cli-yaml` pipeline,  but right now it comes bundled with the part that does, 2) handling automatic uploading (the `dandi` library and all the infrastructure that you are creating).

I just think that it will be hard to remember where it comes from if we need to do that and that and, also, specially for dependencies, explicit is better than implicit.

CodyCBakerPhD:
> 2) handling automatic uploading (the dandi library and all the infrastructure that you are creating).

Also the YAML specification feature uses `dandi` library for `organize` if names aren't provided in the specification.

The question there is whether we want the YAML specification feature to come with minimal installation (I'm thinking yes, since it is the intended use of the package going into the future with the AWS integration).

h-mayorquin:
> > 2. handling automatic uploading (the dandi library and all the infrastructure that you are creating).
> 
> Also the YAML specification feature uses `dandi` library for `organize` if names aren't provided in the specification.
> 
> The question there is whether we want the YAML specification feature to come with minimal installation (I'm thinking yes, since it is the intended use of the package going into the future with the AWS integration).

That's true. I remember now that in fact, that's the only thing that you are using `dandi` for. So they are bundled together because you use dandi to create names with their specific formats for the file, right?

CodyCBakerPhD:
> That's true. I remember now that in fact, that's the only thing that you are using dandi for. So they are bundled together because you use dandi to create names with their specific formats for the file, right?

Yep, which may (and probably will) change over time and can be pretty complicated, so better to just call it directly.

requirements-minimal.txt
```diff
@@ -14,3 +14,4 @@ neo>=0.9.0
 roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@c23e4a00324b381f2aad25d685e1cd68b588b563  # v0.4.7, pre-transpose fix
 lxml>=4.6.5
 scipy>=1.4.1
+click  # must be unspecified version to handle platform/pyVersion
```
h-mayorquin, line 17: I was thinking that loose dependency should work but you tried that, how comes?

h-mayorquin (approved):

requirements-minimal.txt
CodyCBakerPhD: This is something you have to do when the package has different min/max version builds available on different platforms , so this allows it to infer which one is proper for a given setup

We also use it for sonpy since it has similar problems

PR #561, opened 2022-06-28
Add spikeinterface support to intan

h-mayorquin:
Adding `spikeinterface` support to `IntanRecordingInterface`. This should be done after #558

PR #562, opened 2022-06-28
Added docstring and refactor metadata to `IntanRecordingInterface`

h-mayorquin:
Added docstring to intan and some refactoring of the`get_metadat` function. This should be done after #561.

### DIFF
--- a/src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
@@ -129,9 +129,10 @@ class IntanRecordingInterface(BaseRecordingExtractorInterface):
         return metadata_schema
 
     def get_metadata(self):
+        metadata = super().get_metadata()
+        ecephys_metadata = metadata["Ecephys"]
 
         # Add device
-        ecephys_metadata = dict()
         device = dict(
             name="Intan",
             description="Intan recording",
@@ -172,7 +173,4 @@ class IntanRecordingInterface(BaseRecordingExtractorInterface):
                 dict(name="custom_channel_name", description="Custom channel name assigned in Intan.")
             )
 
-        # Build metadata dictionary and return complete structure
-        metadata = dict()
-        metadata.update(Ecephys=ecephys_metadata)
         return metadata
--- a/src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
@@ -76,8 +76,8 @@ class IntanRecordingInterface(BaseRecordingExtractorInterface):
         self,
         file_path: FilePathType,
         stream_id: Optional[str] = "0",
-        spikeextractors_backend: Optional[bool] = False,
-        verbose: Optional[bool] = True,
+        spikeextractors_backend: bool = False,
+        verbose: bool = True,
     ):
         """Load and prepare raw data and corresponding metadata from the Intan format (.rhd or .rhs files).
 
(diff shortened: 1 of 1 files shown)


src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -79,6 +79,21 @@ def __init__(
         spikeextractors_backend: Optional[bool] = False,
```
bendichter, line 79:
```suggestion
        spikeextractors_backend: bool = False,
```

src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -79,6 +79,21 @@ def __init__(
...
+        """Load and prepare raw data and corresponding metadata from the Intan format (.rhd or .rhs files).
+
+
+        Parameters
+        ----------
+        file_path : FilePathType
+            Path to either a rhd or a rhs file
+        stream_id : Optional[str], optional
+            The stream of the data for spikeinterface, "0" by default.
+        spikeextractors_backend : Optional[bool], optional
```
bendichter, line 91:
```suggestion
        spikeextractors_backend : bool
```

src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -79,6 +79,21 @@ def __init__(
...
+            False by default. When True the interface uses the old extractor from the spikextractors library instead
+            of a new spikeinterface object.
+        verbose : Optional[bool], optional
```
bendichter, line 94:
```suggestion
        spikeextractors_backend: bool = False,
        verbose: bool = True,
    ):
        """Load and prepare raw data and corresponding metadata from the Intan format (.rhd or .rhs files).


        Parameters
        ----------
        file_path : FilePathType
            Path to either a rhd or a rhs file
        stream_id : str, optional
            The stream of the data for spikeinterface, "0" by default.
        spikeextractors_backend : bool
            False by default. When True the interface uses the old extractor from the spikextractors library instead
            of a new spikeinterface object.
        verbose : bool
```

bendichter: `Optional` is used from accepting the `None` value, so it is not appropriate here: https://docs.python.org/3/library/typing.html#typing.Optional

h-mayorquin: Thanks, I added this somewhere else where it should not be. I will fix.

src/nwb_conversion_tools/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -114,43 +129,50 @@ def get_metadata_schema(self):
...
             )
         if "custom_channel_name" in recording_extractor_properties:
-            ecephys_metadata["Ecephys"]["Electrodes"].append(
+            ecephys_metadata["Electrodes"].append(
                 dict(name="custom_channel_name", description="Custom channel name assigned in Intan.")
             )
 
-        return ecephys_metadata
+        # Build metadata dictionary and return complete structure
+        metadata = dict()
```
bendichter, line 176: you aren't calling `super().get_metadata`?

h-mayorquin: I just went through the git history of the file and it was never there. Let me check if this allows for code reduction.

h-mayorquin: Added in in the last commit.

CodyCBakerPhD:
@h-mayorquin So this just needs metadata tests now for the Intan interface, right?

h-mayorquin:
> @h-mayorquin So this just needs metadata tests now for the Intan interface, right?

Yes. Along with most of the interfaces.

PR #570, opened 2022-06-29
Remove optional types in pphy sorting data interface

h-mayorquin:
By mistake I added Optional typing to some parameters that can't be None. This PR should address this by removing it.


CodyCBakerPhD:
My mistake for not noticing as well

PR #575, opened 2022-06-30
Add ImagingExtractorDataChunkIterator

### LINKED ISSUE #518: [Feature]: create RoiextractorsDataChunkIterator

The roiextractirs write methods uses the old DataChunkIterator class - I would like this updated to use the new GenericDataChunkIterator.

_No response_

MovieDataChunkIterator

Would also like to buffer over multiple frames and have efficient cross-frame chunking for smaller image sizes

Yes.

### Code of Conduct

weiglszonja:
## Motivation

`ImagingExtractorDataChunkIterator` is a new data chunk iterator for ImagingExtractor objects that uses the `GenericDataChunkIterator` class to enable buffering over multiple frames. Primarily it can be used for writing imaging data to an NWB File.

Notes / questions:

1.
`ImagingExtractorDataChunkIterator` buffers over multiple frames, but cross-frame chunking for smaller image sizes is not possible until ImagingExtractors enable retrieving frames with selected pixel range (i.e. not only frame indices but also pixel by pixel indices) -- until then, even if  `ImagingExtractorDataChunkIterator` enables selection for all dimensions (not only the first dimension) `ImagingExtractor.get_video()` would anyhow return the full image size.

2.
Evaluate performance improvement over `DataChunkIterator`
TODO: timeit to check performance improvement over the previous implementation 

3.
`GenericDataChunkIterator` asserts that chunk_shape evenly divides the buffer_shape. `ImagingExtractorDataChunkIterator` uses default chunk_shape retrieved from the parent class, which might not evenly divide the buffer_shape that was calculated here.

Resolve #518

## How to test the behavior?

TODO: add more unittests
```
test_imaging_extractor_data_chunk_iterator()
```

## Checklist


CodyCBakerPhD:
> ImagingExtractorDataChunkIterator buffers over multiple frames, but cross-frame chunking for smaller image sizes is not possible until ImagingExtractors enable retrieving frames with selected pixel range (i.e. not only frame indices but also pixel by pixel indices) -- until then, even if ImagingExtractorDataChunkIterator enables selection for all dimensions (not only the first dimension) ImagingExtractor.get_video() would anyhow return the full image size.

You should be able to chunk over any sub-pixel region as long as the number of frames in that chunk evenly divides the number of frames in the buffer (though it's typical to figure out the chunk shape first, and then multiply that).

If you meant buffering over sub-pixel regions, yeah that's not supported until ROIExtractors supports it, but this iterator is primarily meant for writing the data to a TwoPhotonSeries in an NWBFile, so we do want to extract all the data.


> GenericDataChunkIterator asserts that chunk_shape evenly divides the buffer_shape. ImagingExtractorDataChunkIterator uses default chunk_shape retrieved from the parent class, which might not evenly divide the buffer_shape that was calculated here.

Let me know what exact error you get there for what exact combination of shapes and we can try to debug this.

bendichter:
> GenericDataChunkIterator asserts that chunk_shape evenly divides the buffer_shape.

@CodyCBakerPhD what if a dimension has a length that is prime?

CodyCBakerPhD:
@weiglszonja It's also worth taking a look at how the MovieDataChunkIterator resolved these issues, since the API is quite similar between the two: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/datainterfaces/behavior/movie/movie_utils.py#L145

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -90,3 +91,66 @@ def _get_maxshape(self) -> tuple:
...
+        display_progress: bool = False,
+        progress_bar_options: Optional[dict] = None,
+    ):
+        self.imaging_extractor = imaging_extractor
+
+        if buffer_gb is None and buffer_shape is None:
+            buffer_gb = 1.0
+
+        if buffer_shape is None:
+            buffer_shape = self._get_default_buffer_shape(buffer_gb=buffer_gb)
```
CodyCBakerPhD, line 118: This should be called after getting the default chunk shape (or otherwise manually set chunk shape): https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/data_utils.py#L327

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -90,3 +91,66 @@ def _get_maxshape(self) -> tuple:
...
+            chunk_mb=chunk_mb,
+            chunk_shape=chunk_shape,
+            display_progress=display_progress,
+            progress_bar_options=progress_bar_options,
+        )
+
+        assert self.buffer_shape[1:] == self.maxshape[1:], (
+            f"Except from the first axis, the buffer shape ({self.buffer_shape}) "
+            f"must be equal to max shape ({self.maxshape})."
+        )
```
CodyCBakerPhD, line 134: Assertions like this would typically go at the start so we can avoid triggering anything else as early as possible. You'd just have to get the `maxshape` for comparison to a manually set `buffer_shape` (and only in the case where the `buffer_shape` is set manually to begin with)

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -90,3 +91,66 @@ def _get_maxshape(self) -> tuple:
...
+
+    def _get_default_buffer_shape(self, buffer_gb: float) -> tuple:
+        num_frames_in_buffer_gb = int(
+            buffer_gb * 1e9 / (np.prod(self.imaging_extractor.get_image_size()) * self._get_dtype().itemsize)
+        )
```
CodyCBakerPhD, line 139: A better approach here would be to calculate the exact size of the chunk shape (same `np.prod` approach as above but using units of chunks instead of frames) and calculating how many chunks you can fit into the buffer.

Since frames is the only 'free' dimension here, the calculation is very similar but that will ensure that axis is a proper multiple (i.e., that the buffer frames divided by chunk frames is an integer)

src/nwb_conversion_tools/tools/hdmf.py
```diff
@@ -90,3 +92,85 @@ def _get_maxshape(self) -> tuple:
...
+                f"must be equal to max shape ({self._maxshape})."
+            )
+
+        if buffer_shape[0] % chunk_shape[0] != 0:
+            # warn if the first axis is not a multiple of the chunk shape
+            warn(
+                f"The first axis of the buffer shape ({buffer_shape[0]}) must be divisible by the first axis of the chunk shape ({chunk_shape[0]})"
+            )
+            # override the chunk shape?
+            chunk_shape = tuple([chunk_shape[0] - (buffer_shape[0] % chunk_shape[0])]) + chunk_shape[1:]
```
CodyCBakerPhD, line 140: I wouldn't override it magically in the event the shape isn't correct - it should be set correctly either manually, or with 100% reliability from the default

CodyCBakerPhD:
As derived in our meeting earlier

image

weiglszonja:
The tests are currently failing because I have `hdmf==3.2.1` (listed in `requirements-full.txt`) while the test dependencies are installed from `requirements-minimal.txt` which permits having newer versions of hdmf (`hdmf>=3.2.1`).

This assertion error message is different in the newer version (`hdmf==3.3.2`):
https://github.com/hdmf-dev/hdmf/blob/0e50852145bbf9a4676b6edc936317f882f9a47e/src/hdmf/data_utils.py#L232

@CodyCBakerPhD, should I make the changes based on the newer version of `hdmf` while also updating the package version in `requirements-full.txt` to match what is in the minimal or is it harmful?

CodyCBakerPhD:
@weiglszonja `hdmf` version is bumped in https://github.com/catalystneuro/nwb-conversion-tools/pull/592 as well, you can mirror it here in `requirements-full` no problem.

For testing, what you can do is `@pytest.mark.skip` that particular assertion test whenever the detected current version of `hdmf` is `<3.3.2`. Let me know if you have questions on how to do that. I'd also add a test likewise to be run only if `>=3.3.2` that uses the new assertion message.

weiglszonja:
ah okay, I'll try it, thanks @CodyCBakerPhD!

weiglszonja:
@CodyCBakerPhD I moved the class to `tools/roiextractors`, and also moved the tests to `tests/test_internals/test_imagingextractordatachunkiterator.py `

h-mayorquin:
Should we close this @weiglszonja ?

weiglszonja:
Closed as it was merged in https://github.com/catalystneuro/neuroconv/pull/54

Issue #568, opened 2022-06-29
Add test for `ScanImageImagingInterface` in `test_gin_ophys.py`.


h-mayorquin:
We have some data on:

`OPHYS_DATA_PATH / "imaging_datasets" / "Tif" / "sample_scanimage.tiff"`

Should we add a conversion to nwb test with this data?

CodyCBakerPhD:
Yes, when we add the ScanImageTiffDataInterface

h-mayorquin:
@CodyCBakerPhD 
So this is different from this?
https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py

CodyCBakerPhD:
Oops, forgot we already added that (I was looking for it under the 'Tiff' heading)

CodyCBakerPhD:
Yeah, could you add that test in, then? Not sure how that slipped through the cracks

h-mayorquin:
> Oops, forgot we already added that (I was looking for it under the 'Tiff' heading)

Maybe it should be there?

CodyCBakerPhD:
> Maybe it should be there?

Depends on how we want things organized for discoverability - based only on their file suffixes, or on the software responsible for generating the file? It is a very slightly different format than normal `.tiff` files, so not exactly the same as the other interface

h-mayorquin:
> > Maybe it should be there?
> 
> Depends on how we want things organized for discoverability - based only on their file suffixes, or on the software responsible for generating the file? It is a very slightly different format than normal `.tiff` files, so not exactly the same as the other interface

If you asked me before I would say the software, but the fact that you failed to found it is evidence for the other position.

CodyCBakerPhD:
> If you asked me before I would say the software, but the fact that you failed to found it is evidence for the other position.

lol 😂 I also might not have been trying that hard TBH, just a quick glance. Also source of the mental mistake might be that I originally (and arbitrarily, I thought, at the time) placed it according to suffix in ROIExtractors. So maybe that should be fixed to adhere to your preference of indexing by software?

h-mayorquin:
I am really not sure, like I would not put cell explorer with something else that is `.mat` for example as it is highly specific. In this case though, I am thinking that they are similar enough that they should be together (I am thinking that we might be able to combine them in such a way that we cover all the cases and use the other as a fallback).

PR #580, opened 2022-06-30
Refactor metadata extraction for `ScanImageImagingInterface`

h-mayorquin:
I am trying to add the gin test for `ScanImageImagingInterface` #568 

However, the metadata that we have right for `ScanImageImagingInterface` does not follow the the right format (the problem is that our format expected lists for imaging plane and two photon series and Ben added the metadata as dicts) and therefore the metadata validation fails.

The PR with the added test comes chained to this. 

As a bonus, one of the warnings that was generated by improper use of `dict_deep_update` is removed


src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -49,12 +50,17 @@ def __init__(
...
+        metadata = super().get_metadata()
 
-        new_metadata = dict(Ophys=dict(TwoPhotonSeries=dict(description=json.dumps(self.image_metadata))))
+        extracted_session_start_time = dateparse(self.image_metadata["state.internal.triggerTimeString"])
+        metadata["NWBFile"] = dict(session_start_time=extracted_session_start_time)
 
-        if "state.internal.triggerTimeString" in self.image_metadata:
-            new_metadata["NWBFile"] = dict(
-                session_start_time=dateparse(self.image_metadata["state.internal.triggerTimeString"])
-            )
```
bendichter, line 58: why did you change this?

bendichter: Now if this field is not present the entire conversion fails

h-mayorquin: When I modified the metadata to wrap the dicts into list I was getting problems matching the data to the expected results in the test (as everything was string) so I re-wrote everything to make the comparison specific to each of the levels. The failure mode that you pointed out is something that did not foresee. 

This could be fixed either by testing the present of the extracted_session_time or by writing everything to a temporary metadata dict and then using `deep_dict_update` as you had it before. I prefer the former, do you have a preference?

h-mayorquin: Btw, while lookin at the extracted json I saw the following:

```
    "state.internal.startupTimeString": "\'10/9/2017 14:38:30.957\'",
    "state.internal.triggerTimeString": "\'10/9/2017 16:57:07.967\'",
```
There are two time strings over there, how did you chose between those two?

bendichter: I would guess the data collection starts at the trigger time

bendichter: I don't really have a preference for how we add the data to the dict as long as we are robust to missing metadata

h-mayorquin: Good. Now I added that. I also added a new assertion in case the sampling frequency was not found.

CodyCBakerPhD:
retriggering tests

src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -10,7 +11,16 @@
     HAVE_PIL = False
 
 from ..tiff.tiffdatainterface import TiffImagingInterface
```
CodyCBakerPhD, line 13: @h-mayorquin Sorry, I'm just now realizing - this is using the `TiffImagingExtractor` instead of the `ScanImageTiffImagingExtractor`; the second one is preferred since it supports memmaping of these files whereas the first one loads the entire file as an array.

Also, as a bonus, the `ScanImageTiffImagingExtractor` has an attribute `_metdata` that has that description string.

Also also, this would swap the dependency from `PIL` to `scanimagetiff` (see ROIExtractors for detail).

h-mayorquin: Good catch, I did not notice this either.

https://github.com/catalystneuro/nwb-conversion-tools/pull/586

CodyCBakerPhD:
LGTM - @h-mayorquin is this good to go?

h-mayorquin:
> LGTM - @h-mayorquin is this good to go?

Yes.

CodyCBakerPhD (approved):

PR #582, opened 2022-06-30
Add spikeinterface support and  and gin test for `CEDRecordingInterface`

h-mayorquin:
Add spikeinterface support to `CEDRecordingInterface`. Part of #500

As discussed with @bendichter  today,  this did not have an associated gin test with `spikeextractors` so I am only adding `spikeinterface` support as it is.

I also had a problem running the old format (`.smr`) tests in my own system but spikeinterface testing suite indicates that it should be supported. I am adding it here to see if it is a problem of local dependencies exclusive to me. If not, then probably we will have to support only the new format ("smrx") for a while until we figure this out.

### DIFF
--- a/src/nwb_conversion_tools/datainterfaces/ecephys/ced/ceddatainterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ecephys/ced/ceddatainterface.py
@@ -1,9 +1,10 @@
-"""Authors: Luiz Tauffer."""
+"""Authors: Heberto Mayorquin, Luiz Tauffer."""
+from pathlib import Path
 from platform import python_version
 from sys import platform
 from packaging import version
 
-from spikeextractors import CEDRecordingExtractor
+from spikeinterface.extractors import CedRecordingExtractor
 
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
 from ....utils import get_schema_from_method_signature, FilePathType
@@ -20,9 +21,10 @@ if platform == "darwin" and version.parse(python_version()) < version.parse("3.8
 
 
 class CEDRecordingInterface(BaseRecordingExtractorInterface):
-    """Primary data interface class for converting a CEDRecordingExtractor."""
+    """Primary data interface class for converting data from CED (Cambridge Electronic Design)
+    ."""
 
-    RX = CEDRecordingExtractor
+    RX = CedRecordingExtractor
 
     @classmethod
     def get_source_schema(cls):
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD (dismissed):

h-mayorquin:
This has the same error than the one in my personal system. It seems that at least temporary we will only be able to support the new format.

CodyCBakerPhD:
> This has the same error than the one in my personal system. It seems that at least temporary we will only be able to support the new format.

Yup, that's fine. We've had trouble with `sonpy` before (hence the platform/PyVersion tests targeted at that) so it's not too much of a surprise

CodyCBakerPhD (approved):

src/nwb_conversion_tools/datainterfaces/ecephys/ced/ceddatainterface.py
```diff
@@ -37,6 +39,9 @@ def get_all_channels_info(cls, file_path: FilePathType):
         assert HAVE_SONPY, INSTALL_MESSAGE
         return cls.RX.get_all_channels_info(file_path=file_path)
 
-    def __init__(self, file_path: FilePathType, smrx_channel_ids: list, verbose: bool = True):
+    def __init__(self, file_path: FilePathType, verbose: bool = True):
```
bendichter, line 42: @h-mayorquin why did you remove this optional arg? iirc this was needed in a prior conversion

CodyCBakerPhD: The old data share no longer exists (wasn't copied to the LBNL drive but instead was a link to their share, which has since been deleted) but I have a hard copy that I'm sending now.

CodyCBakerPhD: @h-mayorquin Confirmed that the previous `.smrx` files can be loaded in the new SpikeInterface without needing the extra argument (or class method)

bendichter: Can we please look into recreating this feature in the new pipeline?

h-mayorquin: Selecting only the channels that you want to load?

bendichter: getting the names of the channels as a class method and selecting only the channels you want to load as the init

h-mayorquin: Ok, check:

https://github.com/catalystneuro/nwb-conversion-tools/issues/583

PR #586, opened 2022-07-04
Scan image interface to use proper roiextractor

h-mayorquin:
Updating the `ScanImageImagingInterface` extractor to use `ScanImageTiffImagingExtractor` from roiextractors instead of `TiffImagingExtractor`.

### DIFF
--- a/src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
@@ -24,18 +24,6 @@ def extract_extra_metadata(file_path):
     return extra_metadata
 
 
-from PIL import Image, ExifTags
-
-
-def extract_extra_metadata2(file_path):
-    image = Image.open(file_path)
-    image_exif = image.getexif()
-    exif = {ExifTags.TAGS[k]: v for k, v in image_exif.items() if k in ExifTags.TAGS and type(v) is not bytes}
-    extra_metadata = {x.split("=")[0]: x.split("=")[1] for x in exif["ImageDescription"].split("\r") if "=" in x}
-
-    return extra_metadata
-
-
 class ScanImageImagingInterface(BaseImagingExtractorInterface):
 
     IX = ScanImageTiffImagingExtractor
--- a/src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
@@ -2,19 +2,32 @@ import json
 from dateutil.parser import parse as dateparse
 from typing import Optional
 
+from roiextractors import ScanImageTiffImagingExtractor
 
 try:
-    from PIL import Image, ExifTags
+    from ScanImageTiffReader import ScanImageTiffReader
 
-    HAVE_PIL = True
+    HAVE_SCAN_IMAGE_TIFF = True
 except ImportError:
-    HAVE_PIL = False
+    HAVE_SCAN_IMAGE_TIFF = False
 
-from ..tiff.tiffdatainterface import TiffImagingInterface
+
+from ..baseimagingextractorinterface import BaseImagingExtractorInterface
 from ....utils import FilePathType
 
 
 def extract_extra_metadata(file_path):
+
+    description = ScanImageTiffReader(str(file_path)).description(iframe=0)
+    extra_metadata = {x.split("=")[0]: x.split("=")[1] for x in description.split("\r") if "=" in x}
+
+    return extra_metadata
+
+
+from PIL import Image, ExifTags
+
+
+def extract_extra_metadata2(file_path):
     image = Image.open(file_path)
     image_exif = image.getexif()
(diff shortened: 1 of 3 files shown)


src/nwb_conversion_tools/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -2,19 +2,32 @@
...
 
 def extract_extra_metadata(file_path):
+
+    description = ScanImageTiffReader(str(file_path)).description(iframe=0)
+    extra_metadata = {x.split("=")[0]: x.split("=")[1] for x in description.split("\r") if "=" in x}
+
+    return extra_metadata
+
+
+from PIL import Image, ExifTags
```
CodyCBakerPhD, line 27: Do you still need PIL at all then?

h-mayorquin: No, forgot to remove.

h-mayorquin:
Some problem with `scanimage-tiff-reader==1.4.1`. There are not many releases:
https://pypi.org/project/scanimage-tiff-reader/#history
And the activity is stalled:
https://gitlab.com/vidriotech/scanimagetiffreader-python

CodyCBakerPhD:
> Some problem with scanimage-tiff-reader==1.4.1. There are not many releases:
https://pypi.org/project/scanimage-tiff-reader/#history
And the activity is stalled:
https://gitlab.com/vidriotech/scanimagetiffreader-python

The problem is actually how it was built prior to release - all you have to do is install `wheel` in the environment prior to iteration over the requirements in our own setup.

See this line from the original PR: https://github.com/catalystneuro/roiextractors/pull/144/files#diff-7dc87d4394e1756c519dbfd0b80d3b31377f643f0bc25d3ed807ce8a3794023dR31

h-mayorquin:
> > Some problem with scanimage-tiff-reader==1.4.1. There are not many releases:
> > https://pypi.org/project/scanimage-tiff-reader/#history
> > And the activity is stalled:
> > https://gitlab.com/vidriotech/scanimagetiffreader-python
> 
> The problem is actually how it was built prior to release - all you have to do is install `wheel` in the environment prior to iteration over the requirements in our own setup.
> 
> See this line from the original PR: https://github.com/catalystneuro/roiextractors/pull/144/files#diff-7dc87d4394e1756c519dbfd0b80d3b31377f643f0bc25d3ed807ce8a3794023dR31

Interesting that the environments don't have wheel installed. I somehow thought this should come by default with pip.

I am concerned about the `scanimage-tiff-reader` package not being maintained. For example, it only tests against against Python 3.6 which does not receive security supports anymore:

https://endoflife.date/python

h-mayorquin:
I was also hoping we could use this but it does not have memmap support.

CodyCBakerPhD:
> I am concerned about the scanimage-tiff-reader package not being maintained. For example, it only tests against against Python 3.6 which does not receive security supports anymore:

We can always contact and request updates - however, _we_ are now doing their job of testing against higher python versions, so as long as they don't prohibit (via upper bounds) certain dependencies like `numpy`, _we_ can ensure that everything works as expected, which it appears to. Are we getting any new or obvious deprecation warnings from this?

CodyCBakerPhD:
> I am concerned about the scanimage-tiff-reader package not being maintained. For example, it only tests against against Python 3.6 which does not receive security supports anymore:

Also analogous to the other  `tiffile` library we use on ROIExtractors: https://pypi.org/project/tiffile/#history

h-mayorquin:
@CodyCBakerPhD 
> > I am concerned about the scanimage-tiff-reader package not being maintained. For example, it only tests against against Python 3.6 which does not receive security supports anymore:
> 
> We can always contact and request updates - however, _we_ are now doing their job of testing against higher python versions, so as long as they don't prohibit (via upper bounds) certain dependencies like `numpy`, _we_ can ensure that everything works as expected, which it appears to. Are we getting any new or obvious deprecation warnings from this?

> > I am concerned about the scanimage-tiff-reader package not being maintained. For example, it only tests against against Python 3.6 which does not receive security supports anymore:
> 
> We can always contact and request updates - however, _we_ are now doing their job of testing against higher python versions, so as long as they don't prohibit (via upper bounds) certain dependencies like `numpy`, _we_ can ensure that everything works as expected, which it appears to. Are we getting any new or obvious deprecation warnings from this?

I don't think that we are doing their job. We are testing that the library can read a specific file -that might as well be old- in some environments that we use ourselves. 

Moreover, I am not coming at it from that angle. I am concerned about building infrastructure around something that might break or starts failing silently just because it does not keep up to date with the changing environment of Python and software as a whole. Given the ecological context of nwbct I feel that being reliable is more important than begin memory efficient (which is what we gain by using this library with their memmap capabilities). 

> Also analogous to the other tiffile library we use on ROIExtractors: https://pypi.org/project/tiffile/#history
>
Yes, the same argument applies. In this case even more strong as we don't gain anything by using this that' can be done by PIL for example.

In fact, I think this suggests a path forward. We could build a reliable TIFF reader based on a more reliable library like PIL or any other and then keep this specific one -by this or any other name- that allows us to do memmaped reading.

CodyCBakerPhD (approved):

PR #592, opened 2022-07-07
Bump PyNWB version

CodyCBakerPhD:
PyNWB just released a new version; bumping it in `full` requirements and testing here

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,5 +1,5 @@
 dandi==0.39.6
-pynwb==2.0.1
+pynwb==2.1.0
 hdmf==3.2.1
 tqdm==4.60.0
 natsort==7.1.1


bendichter:
Now pynwb allows you to not write x,y,z,location, and imp. If we want to use that feature we will need to bump the minimum version too.

CodyCBakerPhD:
> Now pynwb allows you to not write x,y,z,location, and imp. If we want to use that feature we will need to bump the minimum version too.

Good to remember that - we can discuss at next team meeting if we want to require the Conversion Tools to use minimally only the most recent versions of everything.

For now, we've technically always treated those as optional internally; if the recording object has that info, we write it, otherwise set them to NaN ourselves. Raised reminder issue here: https://github.com/catalystneuro/nwb-conversion-tools/issues/594

bendichter (approved):

PR #595, opened 2022-07-07
Prepare for v0.11.37 release

CodyCBakerPhD:
Been a while since we cut a release - now that PyNWB 2.1.0 is out, and we seem debugged for it, just putting together a task list of what else we need.

- [x] Testing SI pinning here, but @h-mayorquin will also report if he recalls why we need current dev branch (or a recent history hash pin of it, anyway).

- [x] @h-mayorquin also volunteered to handle a new ROIExtractors release sometime in the next day.

Anyway, those two should be the only thing we need since we're delaying the Neuroscope LFP changes until `neo` cuts a release including those fixes.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -3,7 +3,6 @@ on:
   schedule:
     - cron: "0 0 * * *"  # daily
   pull_request:
-  workflow_dispatch:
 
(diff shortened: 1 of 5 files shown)


h-mayorquin:
Nice!

Issue #579, opened 2022-06-30
Add support for multiple segments recorders in `add_electrical_series`


h-mayorquin:
Right now within the `add_electrical_series` function in the spikeinterface module the `get_traces` methods only support the case when the recording is mon-segment. 

However, some of the data that we have in gin is multi-segment (specifically I know it is for Neuralynx).  What we are doing right now is to sub-segment the recorder and only propagate one segment:

https://github.com/catalystneuro/nwb-conversion-tools/blob/ca17dd2146b883d60d641090772f236da1a39869/src/nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py#L85-L87

We should modify the function `add_electrical_series` to account for this and write multiple segments.

I was discussing with @bendichter today and for this simple cases could just iterate over every segment and write each as its own electrical series.

h-mayorquin:
Related to #443

CodyCBakerPhD:
To clarify, are you saying the neuralynx files on GIN have more than one `seg_index` per file?

h-mayorquin:
Yes.

CodyCBakerPhD:
![image](https://user-images.githubusercontent.com/51133164/176811424-0b83d6e5-321a-49fa-9af5-8aba8a6652ea.png)

So they do! And interestingly, it seems more complicated to see *how* many segments one might have in total via spikeextractors (even using `dirname`), so maybe this is another case to just support SpikeInterface and not the legacy (they both pipe to `neo` anyhow).

h-mayorquin:
Thanks for checking, I realized that I misread your question above and only answered in positive to the first as I could not parse the second. I think I missed an "exist/are available" there.

CodyCBakerPhD:
Also, how does SpikeInterface handle the start times of the different segments? Is there an internal timestamp denoting the delay between segements? Are they assumed contiguous (in which case why consider them separate at all?)

h-mayorquin:
Every segment has its timestamps. For the data in gin they are not contiguous.

CodyCBakerPhD:
OK then, that is definitely something to add to the `add_electrical_series` method

PR #596, opened 2022-07-14
[neuralynx] Exploit metadata exposed from neo neuralynxrawio 

JuliaSprenger:
## Motivation

The complete neuralynx files (including the headers) are parsed by the neo neuralynxio and do not need to be loaded and parsed again using an alternative approach here. This PR benefits from the changes in https://github.com/NeuralEnsemble/python-neo/pull/1137, but also works with older versions of neo.

## How to test the behavior?
```
Check the NWB metadata when loading any neuralynx dataset. The essential session start time and identifier should be the same as before.
```

## Checklist

### DIFF
--- a/src/nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
+++ b/src/nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
@@ -1,8 +1,7 @@
-"""Authors: Heberto Mayorquin, Cody Baker and Ben Dichter."""
+"""Authors: Heberto Mayorquin, Cody Baker, Ben Dichter and Julia Sprenger"""
 import warnings
 from pathlib import Path
 from natsort import natsorted
-from dateutil import parser
 import json
 
 from spikeinterface.extractors import NeuralynxRecordingExtractor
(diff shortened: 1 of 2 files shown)


JuliaSprenger:
Currently the tests fail as additional metadata are not permitted to be added to the set of NWB metadata. Are there any fields these additional metadata should be written to? Or should I permit additional metadata keys for the validation?

bendichter:
@JuliaSprenger thanks for putting this together! We are usually interested in a few metadata fields. It is not our intention to read all of the header as it is for neo. I'll go through and make adjustments to this PR

src/nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -16,64 +15,91 @@
...
-        open_line = raw_header.split("\n")[2]
-        spliced_line = open_line[24:35] + open_line[-13:]
-        return dict(session_start_time=parser.parse(spliced_line, dayfirst=False))
-
-
-def get_filtering(channel_path: FolderPathType) -> str:
-    """Get the filtering metadata from an .nsc file.
+
+    key_mapping = {
+        "recording_closed": "session_stop_time",
```
bendichter, line 32:
```suggestion
```

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -236,7 +236,9 @@ class TestConverter(NWBConverter):
                     member=interface_kwarg, container=converter.data_interface_objects["TestRecording"].source_data
                 )
         metadata = converter.get_metadata()
-        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+        metadata["NWBFile"].update(session_start_time=metadata["NWBFile"]["session_start_time"].astimezone())
+        if "session_stop_time" in metadata["NWBFile"]:
+            metadata["NWBFile"].update(session_stop_time=metadata["NWBFile"]["session_stop_time"].astimezone())
```
bendichter, line 241:
```suggestion
```

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -236,7 +236,7 @@ class TestConverter(NWBConverter):
                     member=interface_kwarg, container=converter.data_interface_objects["TestRecording"].source_data
                 )
         metadata = converter.get_metadata()
-        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+        metadata["NWBFile"].update(session_start_time=metadata["NWBFile"]["session_start_time"].astimezone())
```
h-mayorquin, line 239: I think this is going to change the session_start_time for all the data in gin and not all interfaces have the session_start_time in their metadata. I don't think it needs to be changed right now.

h-mayorquin (commented):
This PR is implicitly dropping the usage of `spikeextractors` in favor of `spikeinterface` which might be fine but then we should remove the gin test as well, no point in having this. Me and @CodyCBakerPhD were discussing dropping the support for this interface to solve #579.

tests/test_on_data/test_gin_ecephys.py
JuliaSprenger: I implemented this change as in principle also a `session_stop_time` can be extracted now, which will be inconsistent if the `session_start_time` is later than the `session_stop_time`. But if this is not an issue for you I can revert the line. @h-mayorquin What do you prefer?

src/nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -16,64 +14,83 @@
...
+    else:
+        neo_annotations = extractors[0].neo_reader.raw_annotations
+        signal_info = neo_annotations["blocks"][0]["segments"][0]["signals"][0]
+        annotations = signal_info["__array_annotations__"]
+        common_header = {k: annotations.get(k, None) for k in key_mapping}
+
+        # # extraction of general metadata and remapping of of keys to nwb terms
+        # general_metadata['session_start_time'] = annotations['recording_opened']
+        # general_metadata['session_id'] = annotations.get('SessionUUID', '')
+
```
JuliaSprenger, line 51:
```suggestion

```

JuliaSprenger:
> It is not our intention to read all of the header as it is for neo. I'll go through and make adjustments to this PR

So you intentionally go for a lossy conversion and don't even want to put the additional metadata in some comments or notes to be tracked with the data?

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: I don't know if the  `session_stop_time` is something we would actually want to automatically extract, either - there could be separate data added to the file (like behavioral tracking) that has timestamps that go beyond when the recording was deemed to have 'stopped'.

CodyCBakerPhD: Actually it looks like the `session_stop_time` doesn't even exist in the schema (https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=session_start_time#id147) so sorry for that confusion. And yeah, what you say about `session_start_time` is true and is always something to be aware of for multiple interfaces

CodyCBakerPhD: >  In fact, I am a bit surprised it does not throw an error as it is now.

They are now failing due to this reason. The CI wasn't making it past the minimal tests before the fix to external forks not having access to the secrets/environment variable

JuliaSprenger: > Regarding an inconsistent `session_start_time` (that is, smaller than `session_stop_time`) that you mention, how would this line solve or address this?


The line doesn't adress this issue, but at makes sure to use consistent start and stop times. So if you ever are going to validate for consistency, this won't raise an error.

CodyCBakerPhD: The metadata fields covered by NWB and mapped in the conversion tools are seen in this section of our metadata JSON schema: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/schemas/metadata_schema.json#L106-L138

Though that's not quite complete (on the to-do list to make that schema more readable as well as improve the coverage of it), there's also electrode-level metadata too: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py#L49-L56

CodyCBakerPhD: Basic idea is to associate pieces of metadata as adjacently to their corresponding neurodata types as possible. If there's some metadata related to the acquisition device, it should go in one of the device (https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=device#device) fields, if it's session-level metadata it can go at the NWBFile or even Subject level (https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/schemas/metadata_schema.json#L11-L61), if it's electrode-level then associate it per-electrode, etc.

If a given piece of metadata doesn't make too much sense in any particular field, the description of any of those objects is usually the best place for adding free-text information.

h-mayorquin: For `spikeglx` we added a dictionary with a bunch of extra data in the description o the device:
https://github.com/catalystneuro/nwb-conversion-tools/blob/c297c3f48be249275f214a6367cd5e79cb2e038b/src/nwb_conversion_tools/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py#L132-L161

CodyCBakerPhD:
@JuliaSprenger Just a heads up - this project has shifted over to https://github.com/catalystneuro/neuroconv now

CodyCBakerPhD:
Has moved to https://github.com/catalystneuro/neuroconv/pull/170

PR #599, opened 2022-07-18
testing CI from fork

CodyCBakerPhD:
Trying to figure out why some of these tests that don't need to run on external forks are failing skip based on identical environment variable retrieval attempts


CodyCBakerPhD:
And yes, as proven from #597 setting `pull_request_target` fixes the issue but as Julia said if it triggers automatically then our secrets could get stolen by anyone.

Possible alternative could be a `workflow_dispatch` type logic for a manual trigger against the PR target?

CodyCBakerPhD:
OK that's all it took - I didn't realize the environment variable was being set, but being set to `""`. Fixed now.

PR #603, opened 2022-07-20
Refactor write segmentation III (add summary images)

h-mayorquin:
This is related to #601 and #602

This PR deals with refactoring the part of the `write_segmentation` function that adds the summary images (mean and correlation). I refactored this by adding an extra function whose sole purpose is to perform this addition to nwb. I also added some tests for this newly introduced function.

Also removed an unnecessary warning statement that we had.

This requires the branch in the following PR in roiextractors:
https://github.com/catalystneuro/roiextractors/pull/179

The reason being that the dummy segmentation extractor required for testing is a recent addition and the latest commit (specified in the has) also added an option for having a dummy segmentation extractor without summary images. As this runs in both our internal and data tests it is required in both requirement files

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -17,7 +17,7 @@ opencv-python==4.5.1.48
 spikeextractors==0.9.10
 spikeinterface>=0.94.0
 neo==0.10.2
-roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@add_dummy_segmentation_extractor_without_summary_images
+roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@155efa5d0597529b5b02a8ada633d8b0452ab1d7
 pyintan==0.3.0
 pyopenephys==1.1.2
 sonpy>=1.7.1
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -17,7 +17,7 @@ opencv-python==4.5.1.48
 spikeextractors==0.9.10
 spikeinterface>=0.94.0
 neo==0.10.2
-roiextractors==0.4.17
+roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@add_dummy_segmentation_extractor_without_summary_images
 pyintan==0.3.0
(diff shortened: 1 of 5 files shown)


requirements-full.txt
```diff
@@ -17,7 +17,7 @@ opencv-python==4.5.1.48
 spikeextractors==0.9.10
 spikeinterface>=0.94.0
 neo==0.10.2
-roiextractors==0.4.17
+roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@155efa5d0597529b5b02a8ada633d8b0452ab1d7
```
weiglszonja, line 20: Can you remind me why this is needed?

h-mayorquin: The dummy segmentation extractor required for testing is a recent addition and the latest commit also added an option for having a dummy segmentation extractor without summary images. As this runs in both our internal and data tests it is required in both requirement files.

weiglszonja: Ah I see, thanks!

src/nwb_conversion_tools/tools/roiextractors/roiextractors.py
```diff
@@ -373,6 +373,32 @@ def get_nwb_segmentation_metadata(sgmextractor):
     return metadata
 
 
+def add_summary_images(
+    nwbfile: NWBFile, segmentation_extractor: SegmentationExtractor, images_set_name: str
+) -> NWBFile:
```
weiglszonja, line 378: can you also add a docstring here?

h-mayorquin: Yes, thanks for enforcing this.

weiglszonja: Thank you! I think it looks good.

weiglszonja (approved):

h-mayorquin:
Let's merge and close this as this is done in the new repo.

PR #604, opened 2022-07-21
[Testing]: Bypass git-annex install if unneeded

CodyCBakerPhD:
@h-mayorquin This should prevent that error from happening in the future - my guess is chocolatey servers went down for a while, but there's no reason we should actually need to install it every time.

This PR only install git-annex if there is a need to update the cache.


h-mayorquin (approved):
Thanks

